PR #12, opened 2019-01-24
Travis fix

ahartikainen:
#### Submission Checklist

- [x ] Run unit tests
- [ ] Declare copyright holder and open-source license: see below (??)

#### Summary

I fixed the branch by maedoc.
I did some changes do travis works now.
This might not work on Windows 2.7.

#### Copyright and Licensing

### DIFF
--- a/cmdstanpy/model.py
+++ b/cmdstanpy/model.py
@@ -224,7 +224,7 @@ class Run:
     def cmd(self):
         """Generate the command line for this run.
         """
-        if hasattr(self, '_cmd'):
+        if self._cmd is not None:
             return self._cmd
         if not hasattr(self.model, 'exe'):
             self.model.compile()
@@ -243,7 +243,7 @@ class Run:
     def start(self, wait=True):
         """Start the run; invokes executable in subprocess.
         """
-        if self.proc:
+        if self.proc is not None:
             raise RuntimeError('run has already started')
         logger.info('starting %s', ' '.join(self.cmd))
         logger.debug('starting %r', self.cmd)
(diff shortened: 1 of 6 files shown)


cmdstanpy/model.py
```diff
@@ -1,8 +1,8 @@
 import os
-import subprocess
+import subprocess32 as subprocess
```
maedoc: I'd protect the `subprocess32` import in a try-catch so that it defaults to stdlib when user doesn't have the 32 variant installed, in part because author doesn't test on Windows etc

ahartikainen: Yes, this needs better fix. Python 2.7 doesn't have timeout, so I added this for tmp fix. I can put it in try-except

seantalts:
Awesome, thanks @ahartikainen and @maedoc! @maedoc you're reviewing this right?

maedoc:
@ahartikainen thanks for picking up the broken branch and the various fixes along the way. 

> @maedoc you're reviewing this right?

it looks great, so I would merge, but not sure what you etiquette is in stan-dev, do you (a) want multiple reviews before a merge and (b) do a merge commit, squash/merge, rebase/merge?

ahartikainen:
PyStan3 + httpStan takes a step forward on formalism, but this is still non-released version, so in my opinion squash merges are still ok.

In PyStan we try to combine commits to a few meaningful ones. After the first release, I think this repo could also follow some similar idea.

maedoc:
thanks for the feedback, 'tis noted.  since this is just a CI fix PR I'll squash.

Issue #11, opened 2019-01-23
Use black to format code


ahartikainen:
#### Summary:

Formatting code with black makes it more clean.

https://github.com/ambv/black

#### Description:

Black automatic formatting reduces developer hours and keeps codebase clean.

#### Additional Information:

Developer environment needs to be Python 3.6, but the formatted code is still supported by 2.7

maedoc:
Travis runs flake8 before pytest, which is good enough in my experience, to get in the ballpark of "clean readable code", but maybe I'm manually following some of the conventions Black implements.

This would be a precommit hook though, not something to run on CI build.

seantalts:
On Stan we have both - a lot of our developers dislike installing additional tools, so we have a Jenkins stage that will autoformat their code and commit it back to the branch. It's not pretty: https://github.com/stan-dev/math/blob/develop/Jenkinsfile#L64

infrahumano:
I was suggesting in #21 adding at least a few flake8 plugins to improve readability and uniformize divergent practices (for instence with the use of quotations). That could be also resolved with `black`, if I'm not mistaken. 

Another style piece that `black` would not resolve is the use of docstrings. In the new branch I noticed a practice exemplified by [this line](https://github.com/stan-dev/cmdstanpy/blob/refactor/api-per-func-spec/cmdstanpy/lib.py#L24). I would suggest collecting all those parameter descriptions in a proper docstring perhaps following [the Google docstring style](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html). I haven't googled it up, but I am sure there are plugins for flake8 that check this as well.

ahartikainen:
Black can do that. (increases readability).

I like the numpydoc style, because it is quite clear.

mitzimorris:
just found this comparison - https://medium.com/3yourmind/auto-formatters-for-python-8925065f9505
will check out black.

mitzimorris:
diagnostic messages use emojis?  :yuck:

ahartikainen:
This works now 

From cmdstanpy root folder 

    black cmdstanpy

mitzimorris:
many thanks!

PR #21, opened 2019-05-10
Refactor/api per func spec

mitzimorris:
#### Submission Checklist

- [x] Run unit tests

#### Summary

Wraps cmdstan binaries for a python environment; returns sample as ndarray or pandas dataframe.

Includes jupyter notebook illustrating workflow.

Developed per initial spec here: https://github.com/stan-dev/design-docs/blob/master/designs/0002-cmdstanpy_func_spec.md and further discussion with Stan devs.

#### Copyright and Licensing


mitzimorris:
hi @maedoc - ready for reviewing

cmdstanpy/__init__.py
```diff
@@ -0,0 +1,19 @@
+import atexit
+import os.path
+import shutil
+import tempfile
+
+
+CMDSTAN_PATH = os.path.abspath(os.path.join('.', 'releases', 'cmdstan'))
+TMPDIR = tempfile.mkdtemp()
+STANSUMMARY_STATS = ['Mean', 'MCSE', 'StdDev', '5%', '50%', '95%',
+                         'N_Eff', 'N_Eff/s', 'R_hat']
```
infrahumano: There is an indentation problem here.

mitzimorris:
@infrahumano - saw your comment re indentation - code formatting tools used here are:

- yapf to format
- flake8 to check
- emacs for editing/fixing problems reported by flake8

it would be good to adopt a set of conventions and tools - happy to get your input.
best place for this discussion are the Stan forums - https://discourse.mc-stan.org/c/stan-dev

cmdstanpy/cmds.py
```diff
@@ -0,0 +1,184 @@
...
+        ))
+    tp.close()
+    tp.join()
+    if not runset.check_retcodes():
+        msg = "Error during sampling"
+        for i in range(chains):
+            if runset.retcode(i) != 0:
+                msg = '{}, chain {} returned error code {}'.format(
+                    msg, i, runset.retcode(i))
+        raise Exception(msg)
```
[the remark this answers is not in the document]
maedoc: If return codes from CmdStan are semantic (1 -> unable to initialize, 2 -> max iters w/o convergence, …), they would be best translated to specific exceptions.  If they aren't, a `RuntimeException` would be appropriate, with a separate method for returning `stderr` lines.  Return an error msg full of retcodes here isn't going to be particularly valuable.

I agree with @infrahumano 's use of `enumerate` but disagree about dropping the `retcode != 0` which reminds us that retcode is a C-ism that doesn't go away by letting Python coerce to `bool` in the `if` statement.  This needs to match CmdStan's semantics of return codes.

cmdstanpy/__init__.py
```diff
@@ -0,0 +1,19 @@
+import atexit
+import os.path
+import shutil
+import tempfile
+
+
+CMDSTAN_PATH = os.path.abspath(os.path.join('.', 'releases', 'cmdstan'))
+TMPDIR = tempfile.mkdtemp()
```
maedoc: for the record, I still find this weird but it's orthogonal to the overall PR and can be improved in the future.

cmdstanpy/cmds.py
```diff
@@ -0,0 +1,184 @@
...
+        print('translating to {}'.format(hpp_file))
+        stanc_path = os.path.join(CMDSTAN_PATH, 'bin', 'stanc')
+        cmd = [stanc_path, '--o={}'.format(hpp_file), stan_file]
+        print('stan to c++: make args {}'.format(cmd))
+        do_command(cmd)
+        if not os.path.exists(hpp_file):
+            raise Exception('syntax error'.format(stan_file))
+
+    exe_file = os.path.join(path, model_name)
+    if platform.system().lower().startswith('win'):
```
maedoc: you can use the `.exe` extension on other platforms too ;)

cmdstanpy/cmds.py
```diff
@@ -0,0 +1,184 @@
...
+        exe_file += '.exe'
+    if not overwrite and os.path.exists(exe_file):
+        # print('model is up to date') # notify user or not?
+        return Model(stan_file, exe_file)
+    cmd = ['make', 'O={}'.format(opt_lvl), exe_file]
+    print('compiling c++: make args {}'.format(cmd))
+    try:
+        do_command(cmd, CMDSTAN_PATH)
+    except Exception:
```
maedoc: It's sometimes useful to use an API specific `Exception` subclass to ensure you don't catch things you shouldn't be catching.  Python mostly gets this right for you, e.g. you won't catch a `KeyboardInterrupt`, but if there are other exceptions that *should* bubble up to the user, it may be preferable to catch a more specific set of exceptiosn e.g. `OSError, subprocess.CalledProcessError`.

cmdstanpy/utils.py
```diff
@@ -0,0 +1,126 @@
...
+
+
+def scan_stan_csv(filename:str) -> Dict:
+    """Capture essential config, shape from stan_csv file."""
+    dict = {}
+    draws_found = 0
+    file_is_data = False
+    with open(filename) as fp:
+        line = fp.readline().strip()
+        while len(line) > 0 and line.startswith("#"):
```
maedoc: This function is too long; I would break up the nested stuff into functions.  Even if it makes the data munging a little more complex, this makes future modifications, to track the CmdStan csv format, less bug prone.

cmdstanpy/utils.py
```diff
@@ -0,0 +1,126 @@
...
+    dict['draws'] = draws_found
+    return dict
+
+
+def is_prefix(pre:str, names:List[str]):
+    for name in names:
+        if name.startswith(pre) and (len(name) == len(pre)
+                                          or name[len(pre)] == '.'):
+            return True
+    return False
```
maedoc:
```
return any([pre == name.split('.')[0] for name in names])
```
is clearer to my eye, but I let you judge.

maedoc:
@infrahumano I would suggest contributing to the issue on style [over here](https://github.com/stan-dev/cmdstanpy/issues/11).  Once an appropriate styler is chosen it should be incorporated into CI & pre commit hooks so we can keep SNR high.

cmdstanpy/cmds.py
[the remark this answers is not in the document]
infrahumano: Good point. And for the different exceptions depending on retcode maybe one could define a dictionary somewhere mapping codes to exceptions?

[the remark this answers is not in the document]
maedoc: @syclik @bob-carpenter if I may single you out for a quick question: does CmdStan make meaningful use of exit/return codes from the executable (apart from SIGSEGV and the like)?

syclik:
I don’t believe it does. It could be changed to do so.

cmdstanpy/cmds.py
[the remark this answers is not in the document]
bob-carpenter: I don't see anything in the doc about that:   https://github.com/stan-dev/cmdstan/releases/download/v2.19.1/cmdstan-guide-2.19.1.pdf

@mitzimorris may also know as she's been working a lot with this lately.

cmdstanpy/utils.py
```diff
@@ -0,0 +1,126 @@
...
+
+def _rdump_array(key:str, val:np.ndarray) -> str:
+    """Flatten numpy ndarray, format as Rdump variable declaration."""
+    c = 'c(' + ', '.join(map(str, val.T.flat)) + ')'
+    if (val.size, ) == val.shape:
+        return '{key} <- {c}'.format(key=key, c=c)
+    else:
+        dim = '.Dim = c{}'.format(val.shape)
+        struct = '{key} <- structure({c}, {dim})'.format(key=key, c=c, dim=dim)
+        return struct
```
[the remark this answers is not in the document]
bob-carpenter: I do not like this single return style.  It's too hard to reason about variables spread all over and whether they get defined appropriately before the return statement.  I would much rather see the if and else return directly.  

Having said that, I don't really program in Python much, so if everyone in Python hews to single-return style (like we used to use in C to get around the lack of exceptions and need for cleanup and returns codes), I'm not that strongly opposed.

cmdstanpy/cmds.py
bob-carpenter: I agree that it's important to get the exception levels right here.

cmdstanpy/utils.py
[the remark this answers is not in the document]
mitzimorris: using single quotes for strings and double quotes for docstrings - changed this everywhere.

leaving multiple returns as is.

cmdstanpy/utils.py
```diff
@@ -0,0 +1,126 @@
...
+            if isinstance(val, np.ndarray) and val.size > 1:
+                line = _rdump_array(key, val)
+            elif isinstance(val, list) and len(val) > 1:
+                line = _rdump_array(key, np.asarray(val))
+            else:
+                try:
+                    val = val.flat[0]
+                except AttributeError:
+                    pass
+                line = '%s <- %s' % (key, val)
```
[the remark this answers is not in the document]
mitzimorris: done!

cmdstanpy/__init__.py
mitzimorris: agreed.

cmdstanpy/cmds.py
mitzimorris: just trying to be consistent w/ cmdstan makefile.

cmdstanpy/utils.py
mitzimorris: completely agree - much clearer!  done!

mitzimorris: got rid of this function altogether - using logic directly in calling function.

cmdstanpy/utils.py
mitzimorris: broke out function to scan csv file and returns dict for config and drawset headers and shape.

cmdstanpy/cmds.py
mitzimorris: @maedoc raises good points - I've filed an issue so that these can be addressed by a subsequent PR.

PR #32, opened 2019-05-17
fix for windows and make sampling a bit more robust

ahartikainen:
#### Submission Checklist

- [ ] Run unit tests

#### Summary

Add `.replace("\\", "/")` to all paths going inside `cmd`.

Also added a "fix" to make sampling more robust (`try-finally`).

#### Copyright and Licensing

Ari Hartikainen

### DIFF
--- a/cmdstanpy/cmds.py
+++ b/cmdstanpy/cmds.py
@@ -26,14 +26,14 @@ def compile_model(
         raise Exception('no such stan_file {}'.format(stan_file))
     path = os.path.abspath(os.path.dirname(stan_file))
     program_name = os.path.basename(stan_file)
-    model_name = os.path.splitext(program_name)[0]
+    model_name, _ = os.path.splitext(program_name)
 
     hpp_name = model_name + '.hpp'
-    hpp_file = os.path.join(path, hpp_name)
+    hpp_file = os.path.join(path, hpp_name).replace("\\", "/")
     if overwrite or not os.path.exists(hpp_file):
         print('translating to {}'.format(hpp_file))
         stanc_path = os.path.join(cmdstan_path(), 'bin', 'stanc')
-        cmd = [stanc_path, '--o={}'.format(hpp_file), stan_file]
+        cmd = [stanc_path, '--o={}'.format(hpp_file), stan_file.replace("\\", "/")]
         print('stan to c++: make args {}'.format(cmd))
         do_command(cmd)
         if not os.path.exists(hpp_file):
@@ -45,7 +45,7 @@ def compile_model(
     if not overwrite and os.path.exists(exe_file):
         # print('model is up to date') # notify user or not?
         return Model(stan_file, exe_file)
-    cmd = ['make', 'O={}'.format(opt_lvl), exe_file]
+    cmd = ['make', 'O={}'.format(opt_lvl), exe_file.replace("\\", "/")]
     print('compiling c++: make args {}'.format(cmd))
     try:
         do_command(cmd, cmdstan_path())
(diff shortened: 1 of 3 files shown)


ahartikainen:
This is still not fixing paths with space. This could be fixed by using only relative paths or manually moving all the model files to tmp path.

ahartikainen:
Is there some problen with travis?

maedoc:
These sorts of replacements are not usually required when using the `os.path` APIs.  What problems did you run into specifically?

I can imagine problems with spaces in paths when generating commands line arguments like `[..., '--o=c:\my stuff\model.hpp', ...]` needs to be `[..., '--o="c:\my stuff\model.hpp"', ...]`, it seems like there aren't that many paths to handle, and it could just be a matter of fixing the format strings. wdyt?

maedoc:
> Is there some problen with travis?

The CI script to build CmdStan seems [to be failing](https://travis-ci.org/stan-dev/cmdstanpy/builds/533765215#L230):
```
$ ./make_cmdstan.sh
~/build/stan-dev/cmdstanpy/releases ~/build/stan-dev/cmdstanpy
release dir: /home/travis/build/stan-dev/cmdstanpy/releases
latest cmdstan: 
download rc code: 0
gzip: stdin: not in gzip format
tar: Child returned status 1
tar: Error is not recoverable: exiting now
tar rc 2
./make_cmdstan.sh: line 40: cd: cmdstan: No such file or directory
make: *** No rule to make target 'build'.  Stop.
installed cmdstan-
ls: cannot access 'releases/*': No such file or directory
```

ahartikainen:
They are not needed for python, but for cmdstan that only works with posix paths (this is known feature/bug).

Space issue is also due to `make`. So if these are handled by copying files or some other way, then users would not need to worry about where to keep files.

maedoc:
> So if these are handled by copying files or some other way

ah indeed. the original pycmdstan code did everything in a unique tmp dir, so effectively this, and never had an issue.

mitzimorris:
> Is there some problen with travis?

the builds keep failing because Travis fails to get CmdStan from GitHub - the download doesn't get kicked off properly.  re-running the build usually works.   seems like this happens pretty often, if not continually.  a related problem is if GitHub is slow, download gets timed out - depends on time of day, day of week.

I've updated cmdstanpy's `make_cmdstan.sh` script to check the return code on the github download.  that makes it easier to spot the problem.

cmdstanpy/cmds.py
```diff
@@ -147,11 +147,13 @@ def sample(
             cores, cpu_count()))
         cores = cpu_count()
     runset = RunSet(args=args, chains=chains)
-    tp = ThreadPool(cores)
-    for i in range(chains):
-        tp.apply_async(do_sample, (runset, i))
-    tp.close()
-    tp.join()
+    try:
```
mitzimorris: I don't understand why there's a try/finally here w/out exception handling.

cmdstanpy/lib.py
```diff
@@ -429,7 +429,7 @@ def summary(self) -> pd.DataFrame:
             suffix='.csv', prefix=tmp_csv_file, dir=TMPDIR, text=True
         )
         cmd = '{} --csv_file={} {}'.format(
-            cmd_path, tmp_csv_path, ' '.join(self.csv_files)
+            cmd_path, tmp_csv_path.replace("\\", "/"), ' '.join(self.csv_files).replace("\\", "/")
```
mitzimorris: CmdStan's tools should be able to handle Windows paths on a Windows system - why is this necessary?

mitzimorris:
> This is still not fixing paths with space. This could be fixed by using only relative paths

That is what the CmdStan manual says to do:

> Stan programs do not need to be in the <cmdstan-home> directory. The current limitation is that the target executable name can not have spaces – this includes the path to the executable. Spaces in the full path can be avoided by using relative paths. For Windows users, if using the full path, include the drive letter and use forward slashes, e.g. make c:/cmdstan/bernoulli/bernoulli.exe.

cmdstanpy/cmds.py
ahartikainen: If that functions fails, it will close the Threadpool and then throw an Exception.

ahartikainen:
I added some examples using pathlib. We should be thinking do we really want to go with pathlib or not. If true, without explicit str call it will only work with py36+.

Also if we are going with it, should it be first class citizen or not?

I don't think this PR should be merged, but it could be splitted to a few PRs. 

Maybe we could have a function to transform values to Windows compatible format.

cmdstanpy/cmds.py
mitzimorris: got it!  adding this logic as part of PR #40

mitzimorris:
Closing this PR -

PR #35, opened 2019-05-18
Feature/23 cmdstan path

mitzimorris:
#### Submission Checklist

- [x] Run unit tests

#### Summary
get location of CmdStan (`cmdstanpy.CMDSTAN_PATH`) from environment var `CMDSTAN` if set,
else use default location `./releases/cmdstan` - the latter is set up via script `make_cmdstan.sh`.

as part of this PR, made `make_cmdstan.sh` more verbose to help catch Travis CI fails due to initial GitHub download timing out.

#### Copyright and Licensing


ahartikainen:
Could you also create a function to set cmdstan path?

    cmdstanpy.set_cmdstan("path/...")

Also, I don't think it is common to run python on specific folders.

Is that releases where?

Should the default location be in `~/.cmdstan` etc?

mitzimorris:
@ahartikainen 

> Could you also create a function to set cmdstan path?

is it legal to write a function in a module which calls reload on that module?
as implemented, `cmdstanpy.CMDSTAN_PATH` is a constant and its value is determined a module load time.  is there a more Pythonic way to set things up?

mitzimorris:
@ahartikainen - just made you reviewer - also note that PR involves more verbose `make_cmdstan.sh` script.   this helps catch problems with Travis, and should be useful to users with similar connectivity problems.

mitzimorris:
> Is that releases where?

relative to wherever cmdstanpy is installed, because that location is user's control.

cmdstanpy/__init__.py
```diff
@@ -1,10 +1,14 @@
 import atexit
+import os
 import os.path
 import shutil
 import tempfile
 
+if 'CMDSTAN' in os.environ:
```
maedoc: To be a little pedantic, it's preferable to encapsulate this sort of logic in a function and call that function instead of referencing a module global everywhere.  For one, once the module is imported, one can't debug the logic without reloading.

maedoc: In such a function, one'd also have an
```python
if not os.path.exists(cmdstan_path):
    raise CmdStanNotFound(cmdstan_path)
if not os.path.exists(os.path.join(cmdstan_path, 'bin', 'stanc')):
    warn("CmdStan does not seem to be built")
```
etc.  It's a little pedantic, but this already presented a problem on Travis since the tests failed and CmdStanPy didn't notice that CmdStan wasn't built.

maedoc:
> > Could you also create a function to set cmdstan path?

> is it legal to write a function in a module which calls reload on that module?
as implemented, cmdstanpy.CMDSTAN_PATH is a constant and its value is determined at module load time. is there a more Pythonic way to set things up?

See my comment about writing and using a function to get the path. The `set_cmdstan_path` function could simply be a shortcut for `os.environ['CMDSTAN'] = new_path_value`.  A subsequent call to `cmdstan_path` would pick up the new value. (One would hope this function isn't invoked in different threads; can document as non-reentrant)

mitzimorris:
@ahartikainen - merged this, but now understand your comments re location of CmdStan releases dir.  will fix in subsequent PRs.

cmdstanpy/__init__.py
```diff
@@ -1,10 +1,14 @@
...
+    CMDSTAN_PATH = os.environ['CMDSTAN']
+else:
+    CMDSTAN_PATH = os.path.abspath(os.path.join('.', 'releases', 'cmdstan'))
```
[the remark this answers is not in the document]
mitzimorris: > I'm not sure should we install anything manually under site-packages

Doh!!!!  please forgive my being new to Python.  I was only thinking about case for install via GitHub.

> have multiple cmdstan in "default" location

yes, the intent here is that whenever there's a new release of CmdStan, the user can rerun this script and it will get the new release, leaving previous releases in place.

PR #40, opened 2019-05-27
Feature/30 extract adaptation info

mitzimorris:
#### Submission Checklist

- [x] Run unit tests

#### Summary
Added logic to parse adaptation information from stan_csv files:

 + check metric consistency
 + parse stepsize and metric from stan_csv files
 + make available as attributes on `PosteriorSample` object

#### Copyright and Licensing

### DIFF
--- a/cmdstanpy/lib.py
+++ b/cmdstanpy/lib.py
@@ -399,6 +399,10 @@ class PosteriorSample(object):
         """sampler output csv files."""
         self._sample = None
         """assembled draws across all chains, stored column major."""
+        self._metric = None
+        """per-chain metrics."""
+        self._stepsize = None
+        """per-chain stepsizes."""
         if run is None:
             raise ValueError('missing sampler run info')
         if csv_files is None:
@@ -409,6 +413,8 @@ class PosteriorSample(object):
         self._chains = run['chains']
         self._draws = run['draws']
         self._column_names = run['column_names']
+        self._param_names = run['param_names']
+        self._metric_type = run['metric']
 
     def __repr__(self) -> str:
         return 'PosteriorSample(chains={},  draws={}, columns={})'.format(
@@ -465,7 +471,7 @@ class PosteriorSample(object):
                 if not (p in self._column_names or p in pnames_base):
                     raise ValueError('unknown parameter: {}'.format(p))
         if self._sample is None:
-            self._sample = self.get_sample()
+            self.parse_sample()
         data = self._sample.reshape(
             (self._draws * self._chains), len(self._column_names), order='A'
         )
@@ -512,11 +518,34 @@ class PosteriorSample(object):
         return self._column_names
 
     @property
+    def param_names(self) -> (str, ...):
+        """
+        Names of parameter, transformed parameter, and generated quantity model variables.
+        """
+        return self._param_names
+
+    @property
     def csv_files(self) -> (str, ...):
         """
         Full path name to stan_csv files returned by sampler.
         """
         return self._csv_files
+
+    @property
+    def metric_type(self) -> str:
+        return self._metric_type
+
+    @property
+    def metric(self) -> np.ndarray:
+        if self._metric is None:
+            self.parse_sample()
+        return self._metric
+
+    @property
+    def stepsize(self) -> np.ndarray:
+        if self._stepsize is None:
+            self.parse_sample()
+        return self._stepsize
 
     @property
     def sample(self) -> np.ndarray:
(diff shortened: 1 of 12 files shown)


mitzimorris:
hi @ahartikainen and @maedoc - if either of you have cycles to review, your input would be much appreciated.

cmdstanpy/lib.py
```diff
@@ -399,6 +399,10 @@ def __init__(self, run: Dict = None, csv_files: Tuple[str] = None) -> None:
         """sampler output csv files."""
         self._sample = None
         """assembled draws across all chains, stored column major."""
+        self._metric = None
+        """per-chain metrics."""
```
ahartikainen: This is not really that important, but I have used to read comments above the line with `#` before the comment.

Also, if these are comments for users then they should be put in docstring. There is no limit for verbosity.

cmdstanpy/lib.py
```diff
@@ -527,27 +552,59 @@ def sample(self) -> np.ndarray:
...
-        The first time this function is called it assembles the sample
-        from the stan_csv files; subsequent calls to this function
-        return the assembled sample.
+        Allocates and populates the stepsize, metric, and drawset arrays
+        by parsing the validated stan_csv files if they don't already exist.
         """
-        sample = np.empty(
+        if not (self._stepsize is None and self._metric is None and
+                    self._sample is None):
+            return
```
ahartikainen: This could have explicit `None` but can be without. I guess it depends on the linter what is recommend.

cmdstanpy/lib.py
```diff
@@ -527,27 +552,59 @@ def sample(self) -> np.ndarray:
...
-            with open(self._csv_files[chain], 'r') as fd:
-                for line in fd:
-                    if line.startswith('#') or line.startswith('lp__,'):
-                        continue
-                    vs = [float(x) for x in line.split(',')]
-                    sample[draw, chain, :] = vs
-                    draw += 1
-        return sample
+            try:
+                fp = open(self._csv_files[chain], 'r')
```
ahartikainen: This could be handled with `with` that does close the file at the end

cmdstanpy/lib.py
```diff
@@ -527,27 +552,59 @@ def sample(self) -> np.ndarray:
...
+                        xs = line.split(',')
+                        self._metric[chain, :] = [float(x) for x in xs]
+                    else:
+                        for i in range(self._num_params):
+                            line = fp.readline().lstrip(' #\t')
+                            xs = line.split(',')
+                            self._metric[chain, i, :] = [float(x) for x in xs]
+                    # draws
+                    for i in range(self._draws):
+                        line = fp.readline().lstrip(' #\t')
```
ahartikainen: What comments can be found in sample? If they are not needed could this loop be replaced with `np.loadtxt` or `np.genfromtxt`?

cmdstanpy/lib.py
```diff
@@ -527,27 +552,59 @@ def sample(self) -> np.ndarray:
...
+                        xs = line.split(',')
+                        self._sample[i, chain, :] = [float(x) for x in xs]
+                finally:
+                    fp.close()
+            except IOError:
+                raise IOError()
```
ahartikainen: I don't think this is working as it should. This will loose all the traceback info.

If we are not doing any exception handling, we don't need try-except block.

To raise original exception, just call `raise` without error, it will raise the last exception.

cmdstanpy/utils.py
```diff
@@ -128,46 +128,127 @@ def check_csv(filename: str) -> Dict:
...
+        fp = open(filename, 'r')
+        try:
+            lineno = scan_config(fp, dict, lineno)
+            lineno = scan_column_names(fp, dict, lineno)
+            lineno = scan_metric(fp, dict, lineno)
+            lineno = scan_draws(fp, dict, lineno)
+        finally:
+            fp.close()
+    except IOError:
+        raise IOError()
```
ahartikainen: Here the same, `with open` is better than try-except

And if not exception handling is done, not try block is needed

cmdstanpy/utils.py
```diff
@@ -128,46 +128,127 @@ def check_csv(filename: str) -> Dict:
...
+                        'csv'):
+                    dict['data_file'] = key_val[1].strip()
+                elif key_val[0].strip() != 'file':
+                    dict[key_val[0].strip()] = key_val[1].strip()
+        cur_pos = fp.tell()
         line = fp.readline().strip()
-        lineno += 1
-        while len(line) > 0 and line.startswith('#'):
-            line = fp.readline().strip()
+    fp.seek(cur_pos)
```
ahartikainen: I don't remember if there are some problems on Windows with tell and seek.

cmdstanpy/utils.py
```diff
@@ -128,46 +128,127 @@ def check_csv(filename: str) -> Dict:
...
+    return lineno
+
+
+def scan_column_names(fp: TextIO, dict: Dict, lineno: int) -> int:
```
ahartikainen: dict is reserved name, maybe some other name should be used?

cmdstanpy/utils.py
```diff
@@ -128,46 +128,127 @@ def check_csv(filename: str) -> Dict:
...
-            lineno += 1
-    dict['draws'] = draws_found
-    return dict
+                    'line {}: invalid or missing mass matrix '
+                    'specification'.format(lineno))
+        return lineno
 
 
-def parse_header(line: str, dict: Dict) -> None:
+def scan_draws(fp: TextIO, dict: Dict, lineno: int) -> int:
```
ahartikainen: Do we really need to scan draws? (Maybe)

cmdstanpy/lib.py
mitzimorris: removing this set of comments - redundant with comments on the corresponding @property accessors.

cmdstanpy/lib.py
mitzimorris: didn't realize that 'with' statement builds in try/except/finally behavoir.  thanks for the correction!

cmdstanpy/lib.py
mitzimorris: the timing information is written out as a comment following the draws.
the `stan_csv` format is an abomination and we must replace it.  until then, brute-force parsing.

cmdstanpy/lib.py
mitzimorris: removed try-except block.

cmdstanpy/utils.py
mitzimorris: I think that under certain circumstances when error is encountered in the `generated quantities` block, the sampler doesn't write the draw to the csv_file but does continue processing.  This isn't necessarily an error, nonetheless, the user needs to know if the sampler isn't behaving as advertised, especially if downstream analysis expects all chains to have the same shape.

cmdstanpy/utils.py
mitzimorris: using `open` per comment.

cmdstanpy/utils.py
mitzimorris: web search "python" "seek" "windows" only turned up old bug in python 2 about seek and write.

Issue #41, opened 2019-05-27
windows install script or instructions for CmdStan


mitzimorris:
#### Summary:

CmdStanPy requires a local install of CmdStan.  The environment variable `CMDSTAN` is used to specify the path to the CmdStan install.  If this variable isn't set, CmdStanPy uses default directory location `~/.cmdstanpy`.   The script `make_cmdstan.sh` downloads and compiles the latest release of CmdStan from GitHub.  By default it installs the latest version of CmdStan in the default directory.    Flags `-d` and  `-v` are used to specify the directory and version, respectively. 

Modify script `make_cmdstan.sh` as needed for Windows or create a Windows-appropriate install script or installer for CmdStan.

ahartikainen:
This is a tricky one.

By default these are not installed

    sh
    tar

Symlinks are not a thing on Windows... at least they are not a thing we should be touching.

I personally use Cmder Full (https://cmder.net/) which has tar and sh installed by default. Also symlink works on W10, but it failed with too-long-path (some boost docs cause this on PyStan).

Then one option would be to do this with `batch`-file (`.bat`), but giving command line arguments is very tricky. The language is not great.

We could write this with powershell. That could maybe work.

I think our best option is to do this with python. With suitable commandline interface (https://click.palletsprojects.com/en/7.x/) we can create basically same experience as `make_cmdstan.sh` has on other systems.

With Python we could "stream" untar with https://docs.python.org/3/library/tarfile.html
It might be slower, but I have not yeat tested this.

So to tackle this with python we need

    tarfile
    request
    click
    subprocess
    os (`os.makedirs(....)`)
    shutil (maybe)

We can assume user has

    compiler (under `CC`/`CXX`?)
    make

mitzimorris:
excellent suggestion -  of course install script should be written in python!
performance is not an issue here, so slower untar is fine.

we're using shutil already to do the cleanup on the TMPDIR.

mitzimorris:
> Symlinks are not a thing on Windows... at least they are not a thing we should be touching.

OK, there are other ways to hack this.

mitzimorris:
> Symlinks are not a thing on Windows... at least they are not a thing we should be touching.

agreed - symlinks are bad.   alternatives?

mitzimorris:
[...]

Issue #17, opened 2019-04-17
PyPi package?


seantalts:
#### Summary:
Would like to install this from pip if possible!

maedoc:
this is what I had for pycmdstan
```yml
release_pypi:
  stage: deploy
  image: $registry/$tag
  only:
    - tags
  script:
    - python setup.py sdist bdist_wheel
    - twine upload -u $TWINE_USERNAME2 -p $TWINE_PASSWORD2 dist/*
```
something similar should be possible with Jenkins or Travis

ahartikainen:
Before this goes to pypi, I think space in path problem needs to be fixed.

wsuchy:
@ahartikainen could you please elaborate more and/or point to related issue?

ahartikainen:
One can not call CmdStan with an option (for example file location) which contains space. This bug is due to `make` and I don't know if it can be fixed.

What we could do, is to copy all the needed files to temporary location (this is compilation step) and compile our model there, and after compilation do some copy-paste to move created files back to correct location.

I think sample / optimize options can contain space, so it's only the compilation step.

wsuchy:
I was able to invoke `./stanc "../examples/test space/prophet.stan"` as well as call to compiled model with spaces - note the double quotes. It works also if I replace space with `\[space]`. I am not sure though if that works on Windows as well.
I we implement that fix, would it unlock us then the ability to publish the package?

ahartikainen:
Calling individual parts works, it is the `make` that fails

https://www.cmcrossroads.com/article/gnu-make-meets-file-names-spaces-them

I think this is the last part to be fixed before PyPi.

There is probably still a problem with https://github.com/stan-dev/cmdstanpy/blob/master/bin/install_cmdstan (I can test it right now), but that does not stop us making this public.

wsuchy:
I've got your point now. For Linux and Mac it would be sufficient to move the file to temporary dir. Unfortunately the Windows temp location contains user name that still can consist of spaces. 
How about we just raise exception at the moment when we detect spaces in the path?

ahartikainen:
That sounds a good solution.

One other thing could be to use relative paths, that could work if cmdstan is installed under the user folder? But I think for now, it is easier to just raise an error.

I just wonder if this going to kick back at us.

ahartikainen:
Hey, this could work for Windows

https://stackoverflow.com/questions/23598289/how-to-get-windows-short-file-name-in-python

[...]

PR #45, opened 2019-06-09
Feature/packaging

mitzimorris:
#### Submission Checklist

- [x] Run unit tests

#### Summary

- Replaced CmdStan install script `make_cmdstan.sh` with Python script `bin/install_cmdstah.py`- (issue https://github.com/stan-dev/cmdstanpy/issues/41)

- Updated `setup.py` to make `install_cmdstan` available as command line script with package install

- Tested package upload and download via TestPyPI - works on my machine (issue https://github.com/stan-dev/cmdstanpy/issues/17)

- Updated docs to use ScikitLearn templates and styling.

- Edited docs and ipython notebook to reflect current API

#### Copyright and Licensing


bin/install_cmdstan
```diff
@@ -0,0 +1,164 @@
+#!/usr/bin/env python
+"""
+Download and install a CmdStan release from GitHub.
+Optional command line arguments:
+   -v, --version : version, defaults to latest
+   -d, --dir : install directory, defaults to '~/.cmdstanpy
+"""
+import argparse
```
ahartikainen: argparse works. Click makes it a bit easier to create "manual"

bin/install_cmdstan
```diff
@@ -0,0 +1,164 @@
...
+import contextlib
+import os
+import os.path
```
ahartikainen: This is not needed, `os` imports `os.path` by default

bin/install_cmdstan
```diff
@@ -0,0 +1,164 @@
...
+import re
+import shutil
+import subprocess
+import sys
+import tarfile
+import urllib.request
```
ahartikainen: Is there a difference with `request`? Is urllib the default? (vs urllib2, urllib3)

bin/install_cmdstan
```diff
@@ -0,0 +1,164 @@
...
+        """)
+
+
+def install_version(cmdstan_version):
+    with pushd(cmdstan_version):
+        print('Building {} binaries'.format(cmdstan_version))
+        cmd = ['make', 'build']
+        proc = subprocess.Popen(
+            cmd, cwd=None, stdout=subprocess.PIPE, stderr=subprocess.PIPE
+        )
```
ahartikainen: Would `subprocess.call` work better?

bin/install_cmdstan
```diff
@@ -0,0 +1,164 @@
...
+    try:
+        file_tmp, _ = urllib.request.urlretrieve(url, filename=None)
+    except urllib.error.URLError as err:
+        print('Failed to download CmdStan version {} from github.com'.format(
+            version))
+        print(err)
+        sys.exit(3)
+    print('Download successful, file: {}'.format(file_tmp))
+    try:
+        tar = tarfile.open(file_tmp)
```
ahartikainen: Does `with tarfile.open ... as tar` work? Or do this need `finally` to close the file?

bin/install_cmdstan
```diff
@@ -0,0 +1,164 @@
...
+        tar.extractall(os.getcwd())
```
ahartikainen: Is there some progressbar option?

bin/install_cmdstan
```diff
@@ -0,0 +1,164 @@
...
+        cmd = ['make', 'build']
+        proc = subprocess.Popen(
+            cmd, cwd=None, stdout=subprocess.PIPE, stderr=subprocess.PIPE
+        )
+        proc.wait()
+        stdout, stderr = proc.communicate()
+        if proc.returncode:
+            print('Command "make build" failed')
+            if stderr:
+                print(stderr.decode('ascii').strip())
```
ahartikainen: `utf-8`?

bin/install_cmdstan
```diff
@@ -0,0 +1,164 @@
+#!/usr/bin/env python
```
ahartikainen: Could this file have `.py` extension? It still can be called without `.py` on linux

bin/install_cmdstan
```diff
@@ -0,0 +1,164 @@
...
+            with open('tmp_test_w', 'w') as fd:
+                pass
+            os.remove('tmp_test_w')  # cleanup
+        except OSError as e:
+            raise ValueError('Cannot write files to directory {}'.format(
+                install_dir)) from e
+
+
+def main():
+    parser = argparse.ArgumentParser()
```
ahartikainen: Should this output help/manual if no arguments are given?

bin/install_cmdstan
```diff
@@ -0,0 +1,164 @@
...
+        )
+        proc.wait()
+        stdout, stderr = proc.communicate()
+        if proc.returncode:
+            print('Command "make build" failed')
+            if stderr:
+                print(stderr.decode('ascii').strip())
+            sys.exit(3)
+        print('Test model compilation')
+        cmd = ['make', os.path.join('examples', 'bernoulli', 'bernoulli')]
```
ahartikainen: On some systems make is not called make. Maybe it could try to use environmental var with default `make`

bin/install_cmdstan
mitzimorris: went with argparse because it's part of the standard library this script has a simple enough set of command line options.

bin/install_cmdstan
mitzimorris: thanks - will clean up!

bin/install_cmdstan
mitzimorris: for me (Mac OS, Python 3.7.0) there is a difference w/ request - import of `urllib` gives the following error:

> AttributeError: module 'urllib' has no attribute 'request'

bin/install_cmdstan
mitzimorris: need to capture stderr if make fails.

bin/install_cmdstan
mitzimorris: using `finally` to close.

bin/install_cmdstan
ahartikainen: Yes, the import format is correct

I was thinking https://2.python-requests.org/en/master/

But I think urllib is the correct choice here.

bin/install_cmdstan
mitzimorris: progressbar option as part of `extractall` function?  no.

agreed that we need progress bars for download, untar, and compile - will file an issue.

bin/install_cmdstan
mitzimorris: doesn't seem to work on Mac OS.

bin/install_cmdstan
mitzimorris: good catch!  changed to utf-8 everywhere.

bin/install_cmdstan
mitzimorris: if no arguments are given the latest version is installed in `~/.cmdstanpy`.
it only prints the usage if there's an unrecognized argument.

it's in the documentation, FWIW.

bin/install_cmdstan
mitzimorris: this will be a problem for the `compile_model` function as well.
needs more investigation - will file another issue.

mitzimorris:
@ahartikainen - thanks for your careful review!

if you think I've addressed all issues, will merge.
I've tested the dist on TestPyPI - do you think this is ready for upload to PyPI?

ahartikainen:
LGTM

PR #52, opened 2019-06-21
Feature/28 dense metric

mitzimorris:
#### Submission Checklist

- [x] Run unit tests

#### Summary

Refactor of function `sample` and object `SamplerArgs`.

 + Better argument names - as similar as possible to RStan and/or CmdStan
 + Full set of options for all sampler arguments
 + Proper documentation for all sampler arguments

Previously, the `sample` function had too many arguments, e.g., because input data and inits can be either a Python dictionary or a filepath string, there were pairs of arguments `data` and `data_file`, etc.   Changed to single argument which allows specified union of types.

The `sample` function hadn't implemented the full set of CmdStan options - in particular:

 + no option of using a dense mass matrix for the metric
 + no option for per-chain seed, step-size, metric
 + no option for user-specified chain ids
 + no option for specifying warmup schedule (init_buffer, window, term_buffer)

Now the `metric` argument takes either the string `diag` or `dense`, or
a filename or list of filenames.  Added utility functions to scan JSON or Rdump
file to get dimensions of the `inv_metric`

Instead of exposing argument `init_buffer`, `window`, and `term_buffer`
the `sample` command takes a tuple `warmup_schedule` which specifies
the % of iterations allocated to each phase of warmup.  This is in line with
the warning message that the sampler returns when there are fewer than
the sum of the default values for `init_buffer`, `window`, and `term_buffer`.
This avoids the confusing issue of what the `window` argument means
(initial size of window which keeps doubling during adaptation of the mass matrix).

The `SamplerArgs` `validate` function now validates all arguments and
checks inter-arg dependencies as much as possible.  There are unit tests
for good/bad configurations.

#### Copyright and Licensing


cmdstanpy/cmds.py
```diff
@@ -12,32 +12,34 @@
...
 from cmdstanpy import TMPDIR
 from cmdstanpy.lib import Model, StanData, RunSet, SamplerArgs
 from cmdstanpy.utils import cmdstan_path
 
 
 def compile_model(
-    stan_file: str = None, opt_lvl: int = 1, overwrite: bool = False,
-    include_paths: List[str] = None
+    stan_file: str = None,
+    opt_lvl: int = 1,
```
ahartikainen: should the default be 3?

cmdstanpy/cmds.py
```diff
@@ -12,32 +12,34 @@
...
+    include_paths: List[str] = None,
 ) -> Model:
     """
     Compile the given Stan model file to an executable.
 
     :param stan_file: Path to Stan program
 
-    :param opt_lvl: Optimization level for c++ compiler, one of {0, 1, 2, 3}
+    :param opt_lvl: Optimization level for C++ compiler, one of {0, 1, 2, 3}
       where level 0 optimization results in the shortest compilation time
```
ahartikainen: On Windows at least with PyStan my python crashes with `-O0`

cmdstanpy/cmds.py
```diff
@@ -79,140 +81,226 @@ def compile_model(
...
+        as a dictionary with entries matching the data variables,
+        or as the path of a data file in JSON or Rdump format.
+
+    :param chains: Number of sampler chains, should be > 1.
+
+    :param cores: Number of processes to run in parallel. If this value
+        exceeds the number of available processors, only max processors chains
+        will be run in parallel.
+
+    :param seed: The seed for random number generator or a list of per-chain
```
ahartikainen: What is the maximum value accepted?

cmdstanpy/cmds.py
```diff
@@ -79,140 +81,226 @@ def compile_model(
...
+        only the diagonal elements.
+
+        The metric is initialized to a unit matrix.  If the value of the
+        metric argument is a string other than `diag` or `dense` or a
+        list of strings, these are treated as pathnames to a data file
+        in JSON or Rdump format which contains entry `inv_metric` which
+        is either a vector or a full matrix, depending on whether the
+        diagonal or dense covariance matrix is to be estimated.
+
+        The length of the list of pathnames must match the number of chains.
```
ahartikainen: pathnames -> paths (?)

ahartikainen: Also, what happens if the path is duplicated? On Windows that will probably crash, unless each file is copied to unique location (tempfile)

cmdstanpy/lib.py
```diff
@@ -81,194 +82,360 @@ def data_file(self) -> str:
...
+        if self.metric is not None:
+            dims = None
+            if isinstance(self.metric, str):
+                if self.metric == 'diag':
+                    self.metric = 'diag_e'
+                elif self.metric == 'dense':
+                    self.metric = 'dense_e'
+                else:
+                    if not os.path.exists(self.metric):
+                        raise ValueError('no such file {}'.format(self.metric))
```
ahartikainen: I think self.metric could also accept `diag_e` and `dense_e` kwargs

cmdstanpy/cmds.py
mitzimorris: 2^32-1  - will doc

cmdstanpy/cmds.py
mitzimorris: no checks on pathname uniqueness - can add.

cmdstanpy/lib.py
mitzimorris: good point - will do.

cmdstanpy/cmds.py
mitzimorris: that slows down compilation time which is already quite slow.  once we get compilation sped up (Bob's working on that now), we can change default value.

cmdstanpy/lib.py
mitzimorris: the same goes for the inits arg as well, no?  checking there as well.

cmdstanpy/cmds.py
mitzimorris: list of `inits` arg paths must be unique as well, no?  adding checks on both args.

cmdstanpy/cmds.py
ahartikainen: Is there a difference in sampling time?

mitzimorris: a year ago there was some discussion on discourse:  https://discourse.mc-stan.org/t/o2-vs-o3-compiler-optimization-level/3714

we could use boolean arg `optimize` where False is optimization level 1 and True is optimization level either 2 or 3 (sounds like 2 might be preferable)

Issue #60, opened 2019-07-01
add method `optimize`


mitzimorris:
#### Summary:
Add methods to run the CmdStan `optimize` method and to retrieve estimates from the resulting Stan csv files.

#### Description:
To add this functionality we need to:
  + add method `optimize` to `Model` class
  + add methods to parse this information out of the Stan csv file

the current set of csv parsing utilities are specific to the csv produced by the NUTS-HMC sampler.  parsing the `optimize` csv is far more straightforward.

#### Additional Information:
Provide any additional information here.

mitzimorris:
implemented via https://github.com/stan-dev/cmdstanpy/pull/69  - many thanks Chris!

PR #69, opened 2019-07-10
Implemented Optimize support

wsuchy:
#### Submission Checklist

- [x] Run unit tests

#### Summary
Related issue: https://github.com/stan-dev/cmdstanpy/issues/60
Implemented call to `optimize` method. 
When optimizing all unrelated methods from `StanFit` will throw an exception. 
I am open for a discussion regarding methods' naming: `optimized_params_*`
#### Copyright and Licensing

- Salesforce

### DIFF
--- a/cmdstanpy/cmdstan_args.py
+++ b/cmdstanpy/cmdstan_args.py
@@ -257,13 +257,13 @@ class OptimizeArgs(object):
             else:
                 raise ValueError("iter must be type of int")
 
-    def compose_command(self, idx: int, cmd: str) -> str:
+    def compose(self, idx: int, cmd: str) -> str:
         """compose command string for CmdStan for non-default arg values.
         """
 
         cmd = cmd + ' method=optimize'
         if self.algorithm:
-            cmd = cmd + 'algorithm={}'.format(self.algorithm.lower())
+            cmd = cmd + ' algorithm={}'.format(self.algorithm.lower())
         if self.init_alpha is not None:
             cmd = cmd + ' init_alpha={}'.format(self.init_alpha)
         if self.iter is not None:
@@ -283,7 +283,7 @@ class CmdStanArgs(object):
         model_name: str,
         model_exe: str,
         chain_ids: Union[List[int], None],
-        method_args: Union[SamplerArgs, FixedParamArgs],
+        method_args: Union[SamplerArgs, FixedParamArgs, OptimizeArgs],
         data: str = None,
         seed: Union[int, List[int]] = None,
         inits: Union[float, str, List[str]] = None,
--- a/cmdstanpy/cmdstan_args.py
+++ b/cmdstanpy/cmdstan_args.py
@@ -224,6 +224,53 @@ class FixedParamArgs(object):
         pass
 
 
+class OptimizeArgs(object):
+    """Container for arguments for the optimizer."""
+
+    OPTIMIZE_ALGOS = {"BFGS", "LBFGS", "Newton"}
+
+    def __init__(self, algorithm: str = None, init_alpha: float = None, iter: int = None) -> None:
+
+        self.algorithm = algorithm
+        self.init_alpha = init_alpha
+        self.iter = iter
+
+    def validate(self, chains=None) -> None:
+        """
+        Check arguments correctness and consistency.
+        """
+        if self.algorithm is not None and self.algorithm not in self.OPTIMIZE_ALGOS:
+            raise ValueError(
+                "Please specify optimizer algorithms as one of [{}]".format(", ".join(self.OPTIMIZE_ALGOS)))
+
+        if self.init_alpha is not None:
+            if isinstance(self.init_alpha, float):
+                if self.init_alpha < 0.0:
+                    raise ValueError("init_alpha must be greater than 0")
+            else:
+                raise ValueError("init_alpha must be type of float")
+
+        if self.iter is not None:
+            if isinstance(self.iter, int):
+                if self.iter < 0:
+                    raise ValueError("iter must be greater than 0")
+            else:
+                raise ValueError("iter must be type of int")
+
+    def compose_command(self, idx: int, cmd: str) -> str:
+        """compose command string for CmdStan for non-default arg values.
+        """
+
+        cmd = cmd + ' method=optimize'
+        if self.algorithm:
+            cmd = cmd + 'algorithm={}'.format(self.algorithm.lower())
+        if self.init_alpha is not None:
+            cmd = cmd + ' init_alpha={}'.format(self.init_alpha)
+        if self.iter is not None:
+            cmd = cmd + ' iter={}'.format(self.iter)
+        return cmd
+
+
 class CmdStanArgs(object):
     """
     Container for CmdStan command line arguments.
--- a/cmdstanpy/cmdstan_args.py
+++ b/cmdstanpy/cmdstan_args.py
@@ -227,7 +227,12 @@ class OptimizeArgs(object):
 
     OPTIMIZE_ALGOS = {"BFGS", "LBFGS", "Newton"}
 
-    def __init__(self, algorithm: str = None, init_alpha: float = None, iter: int = None) -> None:
+    def __init__(
+            self,
+            algorithm: str = None,
+            init_alpha: float = None,
+            iter: int = None
+            ) -> None:
 
         self.algorithm = algorithm
         self.init_alpha = init_alpha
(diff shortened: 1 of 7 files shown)


cmdstanpy/cmdstan_args.py
```diff
@@ -222,6 +222,61 @@ def validate(self, chains):
...
+        """
+        if self.algorithm is not None and \
+                self.algorithm not in self.OPTIMIZE_ALGOS:
+            raise ValueError(
+                "Please specify optimizer algorithms as one of [{}]"
+                .format(", ".join(self.OPTIMIZE_ALGOS))
+            )
+
+        if self.init_alpha is not None:
+            if isinstance(self.init_alpha, float):
```
ahartikainen: from numbers import Real

cmdstanpy/cmdstan_args.py
```diff
@@ -222,6 +222,61 @@ def validate(self, chains):
...
+                if self.init_alpha < 0.0:
```
ahartikainen: `.0` is not needed

cmdstanpy/cmdstan_args.py
```diff
@@ -222,6 +222,61 @@ def validate(self, chains):
...
+                    raise ValueError("init_alpha must be greater than 0")
+            else:
+                raise ValueError("init_alpha must be type of float")
+
+        if self.iter is not None:
+            if isinstance(self.iter, int):
```
ahartikainen: from numbers import Integral

cmdstanpy/cmdstan_args.py
```diff
@@ -222,6 +222,61 @@ def validate(self, chains):
...
+        """
+
+        cmd = cmd + ' method=optimize'
+        if self.algorithm:
+            cmd = cmd + ' algorithm={}'.format(self.algorithm.lower())
+        if self.init_alpha is not None:
+            cmd = cmd + ' init_alpha={}'.format(self.init_alpha)
+        if self.iter is not None:
+            cmd = cmd + ' iter={}'.format(self.iter)
+        return cmd
```
ahartikainen: these could use `cmd += ' iter...`, but I don't the current format is bad.

cmdstanpy/model.py
```diff
@@ -138,6 +138,86 @@ def compile(
...
+        dummy_chain_id = 0
+        self._do_sample(stanfit, dummy_chain_id)
+
+        if not stanfit._check_retcodes():
+            msg = 'Error during optimizing'
+            if stanfit._retcode(dummy_chain_id) != 0:
+                msg = '{} Got returned error code {}'.format(
+                    msg, stanfit._retcode(dummy_chain_id)
+                )
+            raise Exception(msg)
```
ahartikainen: RStan returns the result even if failure happens. What should CmdStanPy do?

cmdstanpy/utils.py
```diff
@@ -125,16 +125,19 @@ def rdump(path: str, data: Dict) -> None:
...
+    if is_optimizing:
+        draws_spec = 1
     else:
-        draws_spec = 1000
-    if 'thin' in dict:
-        draws_spec = int(math.ceil(draws_spec / dict['thin']))
+        if 'num_samples' in dict:
+            draws_spec = int(dict['num_samples'])
+        else:
+            draws_spec = 1000
```
ahartikainen: draws_spec = int(dict.get('num_samples', 1000))

cmdstanpy/utils.py
```diff
@@ -125,16 +125,19 @@ def rdump(path: str, data: Dict) -> None:
             fd.write('\n')
 
 
-def check_csv(path: str) -> Dict:
+def check_csv(path: str, is_optimizing: bool = False) -> Dict:
     """Capture essential config, shape from stan_csv file."""
-    dict = scan_stan_csv(path)
+    dict = scan_stan_csv(path, is_optimizing=is_optimizing)
```
ahartikainen: I would use some other name than `dict`, but again it is not something to be worried about.

cmdstanpy/utils.py
```diff
@@ -206,6 +210,7 @@ def scan_column_names(fp: TextIO, config_dict: Dict, lineno: int) -> int:
     lineno += 1
     names = line.split(',')
     config_dict['column_names'] = tuple(names)
+    config_dict['num_params'] = len(names) - 1
```
ahartikainen: Is that -1 for `lp__`?

cmdstanpy/utils.py
```diff
@@ -270,23 +275,32 @@ def scan_metric(fp: TextIO, config_dict: Dict, lineno: int) -> int:
...
-        if len(line.split(',')) != num_cols:
+        data = line.split(',')
+        if len(data) != num_cols:
             raise ValueError(
                 'line {}: bad draw, expecting {} items, found {}'.format(
                     lineno, num_cols, len(line.split(','))
                 )
             )
+        if not first_draw and store_first:
+            config_dict['first_draw'] = np.array(data, dtype=np.float64)
```
ahartikainen: does this if-block need to set the first_draw or the store_first to False?

cmdstanpy/stanfit.py
```diff
@@ -124,6 +126,37 @@ def stepsize(self) -> np.ndarray:
             self._assemble_sample()
         return self._stepsize
 
+    @property
+    def is_optimizing(self) -> bool:
+        """Returns true if we are optimizing rather than sampling"""
+        return self._is_optimizing
+
+    @property
```
mitzimorris: is there a reason why first draw would be invalid for any of the other methods?
(I've seen R users hack together exactly this function - e.g. as part of simulation-based model development workflow).

cmdstanpy/utils.py
mitzimorris: this should be put inside a test (cf line 132) 
`if is_optimizing:`
 
the columns in the csv output for the HMC sampler include more things than just `__lp` and params.  which is why when parsing the csv file for the HMC sampler output, the num_params is determined by the number of columns in the  inverse mass matrix (I think - will recheck the code)

wsuchy: @mitzimorris / @ahartikainen this is only the default value for `'num_params'`. In sampling it gets replaced with original implementation, don't you think it is enough to subtract `lp__` from total number of columns to get param length? 
Another question is how should we count vector params, as a single length or total (see my implementation of `optimized_params_dict` - it mimics pystan behavior)?

wsuchy: (Please see my comment above)

cmdstanpy/stanfit.py
wsuchy: I've kept all `optimized_params_*` methods working in sampling context, yet I wonder if the naming is fine

cmdstanpy/model.py
mitzimorris: CmdStanPy should do what CmdStan does.

cmdstanpy/utils.py
mitzimorris: OK w/ implementation for 'num_params'

container type parameters - vector, matrix, and arrays - are counted by the number of elements - this will match the dimension of the mass matrix.

I'm not sure about `optimized_params_dict` -

the csv output reflects the names in the c++ program - e.g. if you have parameters `matrix[2][2] foo` then the generated c++ model names are `foo.1.1` ... `foo.2.2`

I think it would be good to be able to access `foo.1.2` as well as `foo`.  looks like the way that `optimized_params_dict` is set up you lose the column names?  and therefore all container type parameters get flattened into a vector?

cf implementation `get_drawset` - https://github.com/stan-dev/cmdstanpy/blob/dd9e7aba44cc2f64861aaa715ced00ba3b05f7a6/cmdstanpy/stanfit.py#L393-L397 

it's a bit of a grey area here.  my direct experience with vector parameters was working with the model whose output is used as test data - https://github.com/stan-dev/cmdstanpy/tree/60-implement-optimize/test/data/runset-big  - when working with that model in RStan, I usually looked at just the first and last few elements of the parameter.

mitzimorris: OK, great - this makes sense.

cmdstanpy/stanfit.py
```diff
@@ -134,29 +134,56 @@ def is_optimizing(self) -> bool:
...
 
     @property
     def optimized_params_pd(self) -> pd.DataFrame:
         """returns optimized params as pandas DataFrame"""
-        self._optimizing_only()
         return pd.DataFrame([self._first_draw], columns=self.column_names)
 
-    @property
-    def optimized_params_dict(self) -> dict:
+    def optimized_params_dict(self) -> OrderedDict:
```
mitzimorris: prefer the previous implementation.

mitzimorris: thought about it more - definitely prefer that optimized_params_dict directly reflect parameters in the model, i.e., all elements of a container are distinct params.   please revert to earlier implementation.

cmdstanpy/utils.py
wsuchy: Reverted to previous implementation.

mitzimorris:
looks awesome  - many thanks!!!

PR #73, opened 2019-07-13
Added dict support for optimizing

wsuchy:
#### Submission Checklist

- [x] Run unit tests

#### Summary
I've just realized I forgot to add support dict for `data` and `inits` parameters which is already supported by sampling method.
#### Copyright and Licensing

- Salesforce

### DIFF
--- a/cmdstanpy/cmdstan_args.py
+++ b/cmdstanpy/cmdstan_args.py
@@ -290,7 +290,7 @@ class CmdStanArgs(object):
         model_exe: str,
         chain_ids: Union[List[int], None],
         method_args: Union[SamplerArgs, FixedParamArgs, OptimizeArgs],
-        data: str = None,
+        data: Union[str, dict] = None,
         seed: Union[int, List[int]] = None,
         inits: Union[float, str, List[str]] = None,
         output_basename: str = None,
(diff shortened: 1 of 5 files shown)


cmdstanpy/cmdstan_args.py
```diff
@@ -381,9 +381,12 @@ def validate(self) -> None:
                             ' found {}'.format(self.seed[i])
                         )
 
-        if self.data is not None:
+        if isinstance(self.data, str):
             if not os.path.exists(self.data):
                 raise ValueError('no such file {}'.format(self.data))
+        elif not isinstance(self.data, dict):
```
mitzimorris: don't you need to update line 293  - type hint on data - else how will self.data be a dict?

cmdstanpy/cmdstan_args.py
```diff
@@ -381,9 +381,12 @@ def validate(self) -> None:
...
+            if isinstance(self.method_args, OptimizeArgs):
```
mitzimorris: include SamplerArgs as well here please

wsuchy: @mitzimorris I see in many `SamplerArgsTest` tests that data is null - is that expected to run sampling without any data file?

mitzimorris: yes, models without a data block are valid.

one use case is modelswhere the transformed data block is used to generate data for each chain/run via calls to the RNG functions.

wsuchy: I've fixed the logic, now we throw exception when data is None *and* we are optimizing, or when data is different than str or dict.

Issue #22, opened 2019-05-14
investigate granularity of exceptions required for subprocesses as well as CmdStan return codes - are they useful?


mitzimorris:
#### Summary:
Reviewer comments on PR https://github.com/stan-dev/cmdstanpy/pull/21 raise several good points - 

> If return codes from CmdStan are semantic (1 -> unable to initialize, 2 -> max iters w/o convergence, …), they would be best translated to specific exceptions. If they aren't, a RuntimeException would be appropriate, with a separate method for returning stderr lines. Return an error msg full of retcodes here isn't going to be particularly valuable.

and:

> It's sometimes useful to use an API specific Exception subclass to ensure you don't catch things you shouldn't be catching. Python mostly gets this right for you, e.g. you won't catch a KeyboardInterrupt, but if there are other exceptions that should bubble up to the user, it may be preferable to catch a more specific set of exceptiosn e.g. OSError, subprocess.CalledProcessError.

Breaking this out into an issue to be addressed in its own PR.

#### Additional Information:
Provide any additional information here.

seantalts:
Hey, I'm not sure if this issue covers this, but the error messages from CmdStan are swallowed:
```
start chain 1.  
finish chain 1.  
Traceback (most recent call last):
  File "silly.py", line 34, in <module>
    chains=1)
  File "silly.py", line 10, in wrapper
    res = f(*args, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/cmdstanpy/cmds.py", line 165, in sample
    raise Exception(msg)
Exception: Error during sampling, chain 0 returned error code 70
```
This happens any time something goes wrong from CmdStan, in this case I just passed in data that was the wrong shape. Should I make a new issue?

maedoc:
@seantalts this is a problem for me as well.  There doesn't seem to be a way to retrieve the `StanFit` object, where the stderr is being stored, which is a local variable in the `sample` method, when the exception is raised.  A workaround could be to catch the exception and go back a few frames to dig out the local vars, but ideally the stderr should be easy to see, especially in interactive usage.

One way to keep the API as it is, would be to avoid blocking in the `sample` method, and just return the fit object with the chains running.  When any call on the `StanFit` API requires results, block until CSV files are validated.  This would allow the user to check progress on the fit object as well.  If there are errors, the user can query the fit object for what you've called the console files to see what's up.

seantalts:
That seems nice!

mitzimorris:
[...]

PR #75, opened 2019-07-15
Don't raise exceptions on retcodes != 0

wsuchy:
#### Submission Checklist

- [x] Run unit tests

#### Summary
Unfortunately Stan's exit code different than 0 doesn't mean no result. I am checking for empty file instead to throw an exception.

#### Copyright and Licensing

- Salesforce

### DIFF
--- a/cmdstanpy/model.py
+++ b/cmdstanpy/model.py
@@ -261,7 +261,7 @@ class Model(object):
                 msg = '{} Got returned error code {}'.format(
                     msg, stanfit._retcode(dummy_chain_id)
                 )
-            raise Exception(msg)
+            print(msg)
         stanfit._validate_csv_files()
         return stanfit
 
@@ -486,7 +486,7 @@ class Model(object):
                     msg = '{}, chain {} returned error code {}'.format(
                         msg, i, stanfit._retcode(i)
                     )
-            raise Exception(msg)
+            print(msg)
         stanfit._validate_csv_files()
         return stanfit
 
--- a/cmdstanpy/utils.py
+++ b/cmdstanpy/utils.py
@@ -179,6 +179,8 @@ def scan_stan_csv(path: str, is_optimizing: bool = False) -> Dict:
     """Process stan_csv file line by line."""
     dict = {}
     lineno = 0
+    if not os.path.getsize(path):
+        raise RuntimeError("Error! Stan result file is empty")
     with open(path, 'r') as fp:
         lineno = scan_config(fp, dict, lineno)
         lineno = scan_column_names(fp, dict, lineno)
--- a/test/test_model.py
+++ b/test/test_model.py
@@ -239,8 +239,10 @@ class SampleTest(unittest.TestCase):
         bern_model = Model(stan_file=stan, exe_file=exe)
         bern_model.compile()
 
-        with self.assertRaisesRegex(Exception, 'Error during sampling'):
-            bern_fit = bern_model.sample(chains=4,
+        with self.assertRaisesRegex(
+                Exception, 'Error! Stan result file is empty'
+        ):
+            _ = bern_model.sample(chains=4,
                                          cores=2,
                                          seed=12345,
                                          sampling_iters=100)


maedoc:
I understand the why, but this is a slippery slope of trying to interpret failure modes of CmdStan, not to mention coordinating across different interfaces.  It'd be waaay cleaner to push CmdStan to have clearer exit codes.

wsuchy:
@maedoc I would love to see CmdStan returning clearer codes, especially some of them being related to *real* errors (e.g no output available) and some of them just reporting math problems, yet this might take ages for them to fix, and we need to get this working in order to replace pystan in fbprophet (https://github.com/facebook/prophet/issues/1045)

maedoc:
Yeah, again, I get why.   Still, I wouldn't trust a result that *is* found in a file when the exit code isn't 0. I think this needs some input from a core Stan devs.

bob-carpenter:
> I would love to see CmdStan returning clearer codes, especially some of them being related to real errors (e.g no output available) and some of them just reporting math problems, yet this might take ages for them to fix

It's pretty simple C++ code in CmdStan.  I don't think anyone will be opposed to clarifying return codes, so all we need is a clear design in an issue and it should be straightforward to implement.

bob-carpenter:
>  Stan's exit code different than 0 doesn't mean no result

There isn't even doc in our manual about what to expect in the way of return codes.  I filed an [issue with CmdStan to fix the return codes](https://github.com/stan-dev/cmdstan/issues/715).  I'll do a quick survey of the existing C++ to see what's being returned now.  I'm open to suggestions for behavior.  I haven't thought at all about embedding CmdStan in a service, so your input would be greatly appreciated.

yamamotoseiji:
Is there an intermediate compromise that we can arrive at to get cmdstanpy in a better place and work in the changes to cmdstan incrementally?

mitzimorris:
did a quick read  through CmdStan file `command.hpp` and the stan::services methods.
the file `stan/src/stan/services/error_codes.hpp` defines the set of return codes,

for optimization,  return code `70`  indicates that optimization terminated with an error -
as @maedoc said, in that case, the answer cannot be trusted - the optimizer would have
run not found the approximate solution to the function.  
so you're correct - "exit code different than 0 doesn't mean no result" - exit code 70 means invalid result.

the sampler might return non-zero return code is `78` , which means that CmdStan's crazy-ass
argument parser routine can't figure out what to do or has been given conflicting config.
CmdStanPy does its own checking of the config, precisely to avoid CmdStan's command line argument config error messages - but we've probably missed stuff - in this case, the sampler could output csv headers, but there would be no sample.

@wsuchy does this help any?

cmdstanpy/model.py
```diff
@@ -261,7 +261,7 @@ def optimize(
                 msg = '{} Got returned error code {}'.format(
                     msg, stanfit._retcode(dummy_chain_id)
                 )
-            raise Exception(msg)
+            print(msg)
```
mitzimorris: I don't think just printing the message is sufficient - if the retcode is not 0, the csv files are junk and should be ignored - don't parse them.

cmdstanpy/model.py
```diff
@@ -486,7 +486,7 @@ def sample(
                     msg = '{}, chain {} returned error code {}'.format(
                         msg, i, stanfit._retcode(i)
                     )
-            raise Exception(msg)
+            print(msg)
```
mitzimorris: I don't think the csv files will parse if there's a non-zero ret code, so this is probably OK.

cmdstanpy/model.py
wsuchy: I've tested it out and in case of fbprophet when the exit code is 70 the csv file is still readable (just the content is not to be trusted). After all we check the csv content separately.
I will push the change where we raise exception if the exit code is != 0 and != 70

mitzimorris:
I'm not very familiar with the optimizer - found the following in the unit tests - do these strings show up in the csv file or are they sent to the console?

```
  EXPECT_TRUE(bfgs.get_code_string(0) == "Successful step completed");
  EXPECT_TRUE(bfgs.get_code_string(10) == "Convergence detected: absolute parameter change was below tolerance");
  EXPECT_TRUE(bfgs.get_code_string(20) == "Convergence detected: absolute change in objective function was below tolerance");
  EXPECT_TRUE(bfgs.get_code_string(21) == "Convergence detected: relative change in objective function was below tolerance");
  EXPECT_TRUE(bfgs.get_code_string(30) == "Convergence detected: gradient norm is below tolerance");
  EXPECT_TRUE(bfgs.get_code_string(31) == "Convergence detected: relative gradient magnitude is below tolerance");
  EXPECT_TRUE(bfgs.get_code_string(40) == "Maximum number of iterations hit, may not be at an optima");
  EXPECT_TRUE(bfgs.get_code_string(-1) == "Line search failed to achieve a sufficient decrease, no more progress can be made");
  EXPECT_TRUE(bfgs.get_code_string(42) == "Unknown termination code");
  EXPECT_TRUE(bfgs.get_code_string(32) == "Unknown termination code");
  EXPECT_TRUE(bfgs.get_code_string(23) == "Unknown termination code");
  EXPECT_TRUE(bfgs.get_code_string(94) == "Unknown termination code");
```

cmdstanpy/model.py
mitzimorris: sounds good!

maedoc:
> do these strings show up in the csv file or are they sent to the console?

They show up on stout and it's useful to check for some of them to see if a restart would help.

maedoc:
> I haven't thought at all about embedding CmdStan in a service, so your input would be greatly appreciated.

It'd be really nice to have return codes distinguish between the input errors (argument or data invalid) and algorithm failures (bfgs and advi especially). In the latter case, a restart can frequently help, and it's useful for the wrapper/service to use that to reschedule the chain.  I have HPC scripts grepping for specific messages  on stderr but an exit code would be easier.

mitzimorris:
> It'd be really nice to have return codes distinguish between the input errors (argument or data invalid) and algorithm failures (bfgs and advi especially). In the latter case, a restart can frequently help, and it's useful for the wrapper/service to use that to reschedule the chain. I have HPC scripts grepping for specific messages on stderr but an exit code would be easier.

Bob's writeup on the CmdStan issue: https://github.com/stan-dev/cmdstan/issues/715  is a clearer version of my comment above.

this is related to #22 - what kinds of exceptions should CmdStanPy be throwing?

wsuchy:
> I'm not very familiar with the optimizer - found the following in the unit tests - do these strings show up in the csv file or are they sent to the console?
> 
> ```
>   EXPECT_TRUE(bfgs.get_code_string(0) == "...");
...
> ```

Yes, I can see (at least some of) them in the console. The cvs output file is similar to the one with sampling minus metric section

wsuchy:
Currently the problem is in this line: https://github.com/stan-dev/stan/blob/develop/src/stan/services/optimize/lbfgs.hpp#L186 which is explained here: https://github.com/stan-dev/stan/blob/develop/src/stan/optimization/bfgs.hpp#L121
As you can see even a problem like `Maximum number of iterations hit , may not be at an optima` will cause the return code to be 70 and therefore cmdstanpy will throw an exception

mitzimorris:
anything with return code '70' is a run that finished, but the answer is invalid.

if we don't want to throw and exception, is there some way to indicate that the returned StanFit object is invalid so that Prophet can handle it gracefully?

this is also an issue with the sampler - when method `diagnose` reports divergences the sample is most likely invalid - even if only one parameter is problematic, it's a joint model and therefore the fit is invalid.

maedoc:
If the result of a subprocess is invalid, then raising an exception and providing a descriptive message is the least-surprising behavior here, and the client should be ready to catch that exception if they want to proceed with normal execution.

In the specific case of optimization though, eg scipy's optimizers don't raise exceptions when they terminate due to max iterations. It may make sense to not raise on this specific condition even if others (non finite lp) should still raise.

mitzimorris:
totally agree with all of @maedoc's points.   to properly address them will require changes not just to CmdStan but also to the Stan services - I will file more issues and see how many of them I can address on the CmdStan side.

in the meantime,  how about the following:

- allow non-zero return codes (everywhere)

- add StanFit method `validate_optimize` which checks the console logs.  it can return either a boolean or an enum if we need finer distinctions for the ways in which things are invalid.

wsuchy:
I am not a big fan of exceptions to begin with. To me the program should throw it when something _exceptional_ happens that can't be predicted like lack of disk space, out of memory, etc. 
Form the other hand throwing exceptions just because model + given data can't be optimized isn't anything special, it is just fully deterministic behavior and the user will have to wrap the call in try..catch block just because the optimizer might not find a perfect solution.

Imagine something simpler, you've got a program to finding roots for a polynomial and you give it the input that doesn't have roots in real numbers (say 3x^2+2x+5 =0) - I wouldn't expect my program to be interrupted just because of that, rather, I would expect to return None and give me some explanation. The same should happen when the roots aren't exactly matching (but rounded). 
In many modern languages people don't use exceptions anymore in the favor of union type of Error and Success (Haskell, Scala, Kotlin, etc),  even numpy gives you result when dividing by 0!

Now, my proposition is the optimize function would return a tuple of `(bool, str)` where first argument is true iff exit code is 0, and second will be populated from the last console output. Later on, when cmdstan normalizes exit codes we could replace it with enum.

maedoc:
Sounds like you're using the wrong language then! Python, NumPy, Scipy etc use exceptions all over the place.

Optimization isn't deterministic in CmdStan unless you provide the initial values for the parameters, which is why restarting can be useful.

wsuchy:
> Sounds like you're using the wrong language then!

Not sure if that is an invective or not so will leave it without a comment.
> Python, NumPy, Scipy etc use exceptions all over the place.

Did I write they didn't use it? No, I wrote numpy doesn't throw even when you do division by 0 (unlike bare Python). Python mostly throws exceptions when param type is wrong - this is how they deal with dynamic types. 
Language is just a tool - you can write both good and bad code in any language.

> Optimization isn't deterministic in CmdStan unless you provide the initial values for the parameters, which is why restarting can be useful.

This is exactly what is fbprophet doing, they provide init params.

How throwing exceptions benefits over returning a value? What is the benefit of having stack trace in optimization problem?

maedoc:
Yep I was +1 for not raising for optimization to follow Scipy behavior.  I think @mitzimorris suggestion is a good compromise.

mitzimorris:
> Now, my proposition is the optimize function would return a tuple of (bool, str) where first argument is true iff exit code is 0, and second will be populated from the last console output. Later on, when cmdstan normalizes exit codes we could replace it with enum.

this sounds like an excellent plan!

wsuchy:
Good news, it turns out that the return code 70 will occur rarer that I initially assumed. After detailed inspecting of the lbfgs optimizer code it looks like when there is at least one iteration the return code will in fact be 0. 
This means that we could keep the code like it is right now, I am closing the PR.

Issue #48, opened 2019-06-14
Feature / Run CmdStan in background


ahartikainen:
#### Summary:

Start sampling, stop python; ... start python, get progress or read csv files

#### Description:

It could be helpful to have possibility to start cmdstan sampling to background and then later get the csv (or use some other interface).

#### Additional Information:

Similar interface as

    nohup model sample output.csv  > sample.log &

maedoc:
would it be ok to
```
nohup python -m cmdstanpy.run model sample output.csv --log sample.log &
```
? quitting Python means serialized state to file e.g. pid file, and it seems a large overhead in terms of code, vs small cost of a Python process in background waiting for CmdStan executable to finish.

WardBrian:
I realize this issue is old, but is this need resolved by using `from_csv` now?

ahartikainen:
I wonder if there is a possibility we add `nohup` (or whatever is the correct way to do this for each platform) interface?

Maybe some special function that kind of calls `model.sample` (or whatever) and then when the return is empty --> crash (?) -> return None for user? Of course this would mean that the special function needs to have explicit location the output.

Then it is up to the user collect csv files when ready.

WardBrian:
I'm not sure I follow what the use case is for this, really. Are you suggesting we make a version of `sample()` that is async?

ahartikainen:
The nohup stuff is something I have done while starting sampling in servers over ssh which take a long time.

ahartikainen:
But on the other hand, one could always create a python script and manually call that with nohup.

WardBrian:
Is there any problem with using `nohup python my_script.py`?

ahartikainen:
Probably not. maybe we should close this issue

Issue #46, opened 2019-06-10
add progress bar or similar to `install_cmdstan` and `compile_model` and `sample` functions


mitzimorris:
#### Summary:
Please provide a short couple sentence summary.

#### Description:
Describe the issue as clearly as possible.

#### Additional Information:
Provide any additional information here.

maedoc:
`tqdm` would be the go to package in Python.

I think this would need to run a background thread to periodically parse the csv or stdout for progress.

mitzimorris:
progress bar for sampling added by      https://github.com/stan-dev/cmdstanpy/pull/107

mitzimorris:
for installation and compilation we could display a spinning wheel - cf.
https://pypi.org/project/progress/

but can't really calculate % of task completed as there's no reporting from the compiler.

ahartikainen:
But maybe we could if we parse the output from compiler?

mitzimorris:
how do we know what the compiler plan is?   incremental progress can only be visualized if we know the overall number of steps involved.  the sampler reports % of total because there's a total and it's reporting 100/1000, 200/2000 etc.

ahartikainen:
True, but do we get somehow information what file is currently compiled? Or is that just illusion I usually get reading warnings.

mitzimorris:
make generates a ".d" file which lists the dependencies, but make will only re-compile the dependencies if the source is newer than any existing target.  maybe we could track compiling (which is fast) vs. linking which is not, but it would be a challenge and add a lot of technical debt to the code.

a spinning wheel, however annoying, does indicate that the process is alive and is easy to implement.

PR #76, opened 2019-07-16
Automatically set number of used cores

wsuchy:
#### Submission Checklist

- [x] Run unit tests

#### Summary
Instead of having number of cores set to 1 by default, let's set it to the number of chains with the limitation of `cpu_count - 2` (to leave some cores for the OS).

#### Copyright and Licensing

- Salesforce

### DIFF
--- a/cmdstanpy/model.py
+++ b/cmdstanpy/model.py
@@ -419,21 +419,22 @@ class Model(object):
                             ' found {}'.format(chain_ids[i])
                         )
 
+        cores_avail = cpu_count()
+
         if cores is None:
-            # leave at least 2 cores for the OS
-            cores = min(cpu_count() - 2, chains)
+            cores = max(min(cores_avail - 2, chains), 1)
 
         if cores < 1:
             raise ValueError(
                 'cores must be a positive integer value, found {}'.format(cores)
             )
-        if cores > cpu_count():
+        if cores > cores_avail:
             print(
                 'requested {} cores, only {} available'.format(
                     cores, cpu_count()
                 )
             )
-            cores = cpu_count()
+            cores = cores_avail
 
             # TODO:  issue 49: inits can be initialization function
 
--- a/cmdstanpy/model.py
+++ b/cmdstanpy/model.py
@@ -271,7 +271,7 @@ class Model(object):
         self,
         data: Union[Dict, str] = None,
         chains: int = 4,
-        cores: int = 1,
+        cores: Union[int, None] = None,
         seed: Union[int, List[int]] = None,
         chain_ids: Union[int, List[int]] = None,
         inits: Union[Dict, float, str, List[str]] = None,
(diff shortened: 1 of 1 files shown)


maedoc:
How is cpu count being computed? If this returns the number of hyperthreads, then performance suffers à lot when Stan chains overlap on the same physical core.

mitzimorris:
I don't see any problem with doing things this way.  I don't think we need to enforce uniform behavoir across the interfaces down to this level.  any thoughts from @jgabry or @ahartikainen ?

wsuchy:
@maedoc  yes it does give you number of HT, yet my assumption is that on modern machines number of chains will usually be smaller (x2) than number of total HT cores. Example: for 4 chains and 12 HT cores (6 real) you will get still full performance. I am not sure if there is an easy way of getting number of real cpus that would work on Mac, Linux and Windows.

mitzimorris:
> my assumption is that on modern machines number of chains will usually be smaller (x2) than number of total HT cores

we need to plan for a wide range of potential users. 
is this an issue for a 'non-modern' machine?  old-school cluster?

maedoc:
Anyone running Stan likely has 2 cores or more, but I would strive to avoid more chains than physical cores as resource contention makes it too slow (except on Xeon Phi...)

wsuchy:
@mitzimorris It will not impact old machines, we are actually interested in achieving to performance for modern MacBook etc where you have e.g 6 physical cores by 12 logical due to the hyper threading.
@maedoc I tested it with prophet model and got the result (MacBookPro, 6 cores, 12 HT):
12 chains, 12 cores: TIME: 242.229299359
12 chains 6 cores: TIME: 286.67894008400003
It turns out to be the opposite, the HT improved the performance by ~ 15.5%

maedoc:
That's not the general trend I've seen across the hardware I've benchmarked, but it could be differences in model as well, especially if yours sit nicely in L3 cache.

wsuchy:
@maedoc as we are talking about the default value here (that users can change) do you think we could close the loop now and merge it, or perhaps more investigation is needed?

mitzimorris:
> 12 chains, 12 cores: TIME: 242.229299359

with this PR, you would get 10 chains for 12 cores, correct?
what do the times look like for that?

mitzimorris:
> psutil.cpu_count(logical=False)
> that would require us to use another library @mitzimorris if that ok to have it as another dependency?

are there other features that could benefit from psutil?  e.g. issues #46 or #48 ?

wsuchy:
With 12 chains and 10 cores the result is  308.07261098000004 (note that I intentionally left some resources for the OS). I am not sure if the library would solve any of the issues out-of-the-box. Actually both issues could be easily solved if CmdStan implemented a way of talking to it via TCP but that's a different story.

mitzimorris:
so no to adding dependency on psutil.
agreed that since this only affects default behavior it's OK.

PR #79, opened 2019-07-18
Minor fixes

wsuchy:
#### Submission Checklist

- [x] Run unit tests

#### Summary
Minor fixes like exception type, more logging ad additional args check.

#### Copyright and Licensing

- Salesforce

### DIFF
--- a/cmdstanpy/model.py
+++ b/cmdstanpy/model.py
@@ -435,10 +435,10 @@ class Model(object):
                 'cores must be a positive integer value, found {}'.format(cores)
             )
         if cores > cores_avail:
-            self._logger.warn(
-                'requested {} cores, only {} available'.format(
-                    cores, cpu_count()
-                )
+            self._logger.warning(
+                'requested %u cores, only %u available',
+                cores,
+                cpu_count()
             )
             cores = cores_avail
 
--- a/cmdstanpy/model.py
+++ b/cmdstanpy/model.py
@@ -130,7 +130,7 @@ class Model(object):
             hpp_file = os.path.splitext(stan_file)[0] + '.hpp'
             hpp_file = Path(hpp_file).as_posix()
             if overwrite or not os.path.exists(hpp_file):
-                self._logger.info('stan to c++ ({})'.format(hpp_file))
+                self._logger.info('stan to c++ (%s)', hpp_file)
                 stanc_path = os.path.join(
                     cmdstan_path(), 'bin', 'stanc' + EXTENSION
                 )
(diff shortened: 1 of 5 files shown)


cmdstanpy/model.py
```diff
@@ -493,6 +493,7 @@ def _do_sample(self, stanfit: StanFit, idx: int) -> None:
         """
         cmd = stanfit.cmds[idx]
         self._logger.info('start chain {}.  '.format(idx + 1))
+        self._logger.debug("sampling: %s", cmd)
```
mitzimorris: please use new style formatting everywhere.

maedoc:
Logging statements in principle should always pass arguments to the logging method, and not format at the call site, to avoid formatting work which isn't required.

cmdstanpy/model.py
ahartikainen: Did this need some setup to work correctly (logging format style?), but I'm not sure if this affects our use of the logging module.

wsuchy:
Indeed, see here: https://docs.python.org/3/library/logging.html#logging.Logger.debug the logger does formatting for you

cmdstanpy/model.py
mitzimorris: I see, according to this:  https://docs.python.org/3/howto/logging-cookbook.html#use-of-alternative-formatting-styles  new style formatting requires config - default is old-style formatting, as above.

bob-carpenter:
> Logging statements in principle should always pass arguments to the logging method, and not format at the call site, to avoid formatting work which isn't required.

If the logger methods are limited, you can use this convenient Java idiom:

```
if error
  if logger level > some threshold
    do some work to construct an error message
    log message at logger level
```

This is what we do before throwing exceptions in the C++ code.

mitzimorris:
I would like to see new-style formating, e.g.  `'blah blah {}'.format(x)` used everywhere for the (foolish) sake of consistency, but it would also be OK to use old style formatting if this is what older Python 3 versions require - in which case, let's use old-style formatting with single quotes everywhere.

here's a grep for calls to `self._logger`:

`cmdstanpy/model.py` :
```
133:                self._logger.info('stan to c++ ({})'.format(hpp_file))

173:                self._logger.error('make cmd failed %s', e)

495:        self._logger.info('start chain {}.  '.format(idx + 1))
502:        self._logger.info('finish chain {}.  '.format(idx + 1))
```

 `cmdstanpy/utils.py` :
```
62:                    self._logger.debug('input tempfile: {}'.format(fd.name))
```

apologies for being a small-minded hobgoblin - do we need to explicitly adopt a style guide?
I hate to add too much process to a lightweight wrapper module.

mitzimorris:
just read most recent comments - agreed - old style formatting, single quotes please.
see previous comments for things that need changed.

PR #82, opened 2019-07-21
Windows short path

ahartikainen:
#### Submission Checklist

- [x] Run unit tests

#### Summary

On Windows, if abspath contains spaces, use shortpath notation, which does not contain spaces.

#### Copyright and Licensing

Ari Hartikainen

### DIFF
--- a/cmdstanpy/utils.py
+++ b/cmdstanpy/utils.py
@@ -499,6 +499,8 @@ def windows_short_path(path: str) -> str:
             output_buf_size = needed
 
     short_path = (
-        os.path.join(short_base_path, file_name) if file_name else short_base_path
+        os.path.join(short_base_path, file_name)
+        if file_name
+        else short_base_path
     )
     return short_path
--- a/cmdstanpy/utils.py
+++ b/cmdstanpy/utils.py
@@ -472,9 +472,8 @@ def windows_short_path(path: str) -> str:
 
     if not os.path.exists(base_path):
         raise RuntimeError(
-            'Windows short path function needs a valid directory. Base directory does not exist: "{}"'.format(
-                base_path
-            )
+            'Windows short path function needs a valid directory. '
+            'Base directory does not exist: "{}"'.format(base_path)
         )
 
     import ctypes
--- a/cmdstanpy/utils.py
+++ b/cmdstanpy/utils.py
@@ -426,12 +426,14 @@ def read_rdump_metric(path: str) -> List[int]:
         data = fp.read().replace('\n', '')
         m1 = re.search(r'inv_metric\s*<-\s*structure\(\s*c\(', data)
         if not m1:
-            return None
-        m2 = re.search(r'\.Dim\s*=\s*c\(([^)]+)\)', data, m1.end())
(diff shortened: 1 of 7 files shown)


ahartikainen:
This works correctly for the model building. Sampling also works without any problems.

stansummary takes long time. I will still see if there is some problem or not. (also testing diagnostics) <- these are from path with spaces.

ahartikainen:
Ok, there is apparently still more stuff breaking on Windows (tempfiles are a mess, but we don't actually use them). I think I have fixed now most of the problems.

Can we somehow create TMPDIR so we could have convenience function to remove old CmdStanPy folders? This is something that will fail with Windows (delete used files, at least probably, need to test).

(code still needs cleanup)

ahartikainen:
Ok, there is still PermissionError trying to save the results.

ahartikainen:
Ok, I think now it should work as intended. There were some files left open, so I rearranged open/close so it works.

(There is still some cleanup to do, I will continue this in the evening)

ahartikainen:
Ok, now everything should work on Windows. I tested install_cmdstan it worked fine.

I have a conda environment with gcc installed (mingw-w64) (notice, cmdstan/cmdstanpy don't need libpython)

    conda install m2w64-toolchain -c msys2

Then running in cmdstan folder

    set MAKE=mingw32-make
    python bin/install_cmdstan

Those `.wait()` call somehow tripped Windows, and they are not needed, because `.communicate()` calls it inside itself. There is possibility that `.communicate` will overflow the output (60 000 chr?), but all the important bits are moved/used with files so I think this is ok for now. Later we could implement output going straight to external file?

This is ready for review after the tests are passed.

cmdstanpy/utils.py
```diff
@@ -410,12 +422,14 @@ def read_rdump_metric(path: str) -> List[int]:
...
         m1 = re.search(r'inv_metric\s*<-\s*structure\(\s*c\(', data)
         if not m1:
-            return None
-        m2 = re.search(r'\.Dim\s*=\s*c\(([^)]+)\)', data, m1.end())
-        if not m2:
-            return None
-        dims = m2.group(1).split(',')
-        return [int(d) for d in dims]
+            return_value = None
+        else:
```
mitzimorris: else is for readability?  the semanticist in my objects, the pragmatist applauds.

mitzimorris:
is there a way to add unit tests for these on a windows machine?
does this require and update to Travis config as well?

ahartikainen:
Yes, we can add `.appveyor.yml`, I can do it in another PR.

cmdstanpy/utils.py
ahartikainen: No, the code behaves differently. It could probably look nicer.

If we return a value within a `with` block, the file is not closed, which causes problems on Windows.

mitzimorris:
do we need more unit tests for this behavior?

ahartikainen:
I think the only testable new thing, is the short path code, which needs appveyor.

Other things are just minor/bugfixes.

Issue #89, opened 2019-07-26
Add rload function


maedoc:
#### Summary:
Add a `rload` that reads a file generated by `rdump`.

#### Description:
Some of my workflows find use for reading an Rdump file back into memory, e.g. instead of rerunning an expensive script that generates the Rdump file in the first place.  In `pycmdstan` I had a `rload` but it doesn't seem to have survived.

#### Additional Information:
From [here](https://gitlab.thevirtualbrain.org/tvb/pycmdstan/blob/master/pycmdstan/io.py#L39) it reads
```python
def rload(fname):
    """Load a dict of data from an R dump format file.
    """
    with open(fname, 'r') as fd:
        lines = fd.readlines()
    data = {}
    for line in lines:
        lhs, rhs = [_.strip() for _ in line.split('<-')]
        if rhs.startswith('structure'):
            *_, vals, dim = rhs.replace('(', ' ').replace(')', ' ').split('c')
            vals = [float(v) for v in vals.split(',')[:-1]]
            dim = [int(v) for v in dim.split(',')]
            val = np.array(vals).reshape(dim[::-1]).T
        elif rhs.startswith('c'):
            val = np.array([float(_) for _ in rhs[2:-1].split(',')])
        else:
            try:
                val = int(rhs)
            except:
                try:
                    val = float(rhs)
                except:
                    raise ValueError(rhs)
        data[lhs] = val
    return data
```
and may find use within other library functions such as the read_metric function.

#### Current Version:
I just pip installed it, 0.4.1 I think.

ahartikainen:
Does this handle dataframe input?

maedoc:
> Does this handle dataframe input?

Unless `dataframe` means the `A <- structure(1, 2, 3, 4, c=(2,2))` thing, no.  I put this together by inspecting Rdump files meant for Stan models.

ahartikainen:
Ok, just making sure. At least that is what PyStan etc implement.

There are however some dataframe in rdump format in example models, and I have yet tried if that works for CmdStan / RStan. They are not working for PyStan.

mitzimorris:
will add this back in!

mitzimorris:
@ahartikainen - re:
> some dataframe in rdump format in example models

which ones are these?  there's stuff in repo example models that doesn't work - we need to flag them and do some cleanup.

ahartikainen:
For example ARD data

[...]

PR #90, opened 2019-07-29
Feature/89 rload fn

mitzimorris:
#### Submission Checklist

- [x] Run unit tests

#### Summary

Add utility function `rload` - see issue #89 for discussion.
Test data for unit tests taken from Stan `src/test/test-models/good`.

#### Copyright and Licensing


cmdstanpy/utils.py
```diff
@@ -216,6 +216,56 @@ def rdump(path: str, data: Dict) -> None:
...
+        next_var = idx
+        var_data = ''.join(lines[start_idx:next_var]).replace('\n', '')
+        lhs, rhs = [_.strip() for _ in var_data.split('<-')]
+        lhs = lhs.replace('"','')
+        rhs = rhs.replace('L','')
+        if rhs.startswith('structure'):
+            *_, vals, dim = rhs.replace('(', ' ').replace(')', ' ').split('c')
+            vals = [float(v) for v in vals.split(',')[:-1]]
+            dim = [int(v) for v in dim.split(',')]
+            val = np.array(vals).reshape(dim[::-1]).T
```
ahartikainen: what is going on here?

Would this equal

    np.array(vals).reshape(dim[::-1]).T 
    --> np.array(vals, order='F').reshape(dim, order='F')
    or
    --> np.array(vals).reshape(dim, order='F')
    or
    --> np.reshape(vals, dim, order='F')

cmdstanpy/utils.py
```diff
@@ -216,6 +216,56 @@ def rdump(path: str, data: Dict) -> None:
...
+    if idx == len(lines):
+        return None
+    start_idx = idx
+    idx += 1
+    while True:
+        while idx < len(lines) and '<-' not in lines[idx]:
+            idx += 1
+        next_var = idx
+        var_data = ''.join(lines[start_idx:next_var]).replace('\n', '')
+        lhs, rhs = [_.strip() for _ in var_data.split('<-')]
```
ahartikainen: This is fine, just a small comment: using `_` instead of `item` seems weird to me.

cmdstanpy/utils.py
```diff
@@ -216,6 +216,56 @@ def rdump(path: str, data: Dict) -> None:
...
 
+def rload(fname: str) -> dict:
+    """Parse data and parameter variable values from an R dump format file.
+       This parser only supports the subset of R dump data as described
+       in the "Dump Data Format" section of the CmdStan manual, i.e.,
+       scalar, vector, matrix, and array data types.
+    """
+    data_dict = {}
+    with open(fname, 'r') as fp:
+        lines = fp.readlines()
```
ahartikainen: This is fine, larger than RAM files are not really that common.

cmdstanpy/utils.py
```diff
@@ -216,6 +216,56 @@ def rdump(path: str, data: Dict) -> None:
...
+        else:
+            try:
+                val = int(rhs)
+            except TypeError:
+                try:
+                    val = float(rhs)
+                except TypeError:
+                    raise ValueError(rhs)
+        data_dict[lhs] = val
+        if idx == len(lines):
```
ahartikainen: should this check be after `idx += 1` ?

cmdstanpy/utils.py
mitzimorris: agreed, that's not very clear.  I added a subroutine that processes the Rdump multi-dim structure - used an ugly regex with named groups for the essential bits - array values, array dimensions.   I ran the  Rdump file in R to see what R does, then added unit tests that verify that the array values end up in the right place.   hope this helps.

cmdstanpy/utils.py
mitzimorris: agreed, fixed.

cmdstanpy/utils.py
mitzimorris: no, this logic is correct.
line 238  will have done the incrementing already.
the loop is set up  this way to make sure that the final var decl gets processed.

mitzimorris:
I did a rewrite of the drump parsing logic and plugged it into the `read_rdump_metric` helper fn.  @ahartikainen, if you have any free cycles, please recheck.

(this feature is as much of a PITA as everything else to do with R, but I can see its utility...)

PR #96, opened 2019-08-02
C++ toolchain installation for Windows

ahartikainen:
#### Submission Checklist

- [x] Run unit tests

#### Summary

Enable user to install C++ with script / from python session.

Needs still some streamlining and editing, so any system can be used. Currently many functions
work only with Windows.

Also needs functionality to cmdstanpy to easily set-up make/local to use the installed toolchain.

#### Copyright and Licensing

Ari Hartikainen


ahartikainen:
install part is ready. Still need some convenience function to add RTools bin folders to path.

ahartikainen:
This is ready, needs some docs.

Also, this code needs some cleaning if we implement the same thing for the macOS and linux.

ahartikainen:
Yes, I think we could install toolchain for appveyor with this and then add tests for win vs others.

After we get this in, it would be interesting to do the same for clang on linux/macos.

ahartikainen:
I think this is ready now.

I need to figure out how to correctly test installation (and also so it doesn't reinstall). We could use install location in appveyor.yml, but not sure is that a good idea.

PR #106, opened 2019-08-16
minor changes to getting started guide

annapaux:
main change is fixing the model declaration: 
```
bernoulli_model = Model(
            stan_file=os.path.join(cmdstan_path(), 'examples', 'bernoulli', 'bernoulli.stan'),
            exe_file=os.path.join(cmdstan_path(), 'examples', 'bernoulli', 'bernoulli')
            )
```

the rest is insignificant.

### DIFF
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -152,7 +152,10 @@ the index '0' corresponds to the first chain in the ``StanFit``:
 .. code-block:: python
 
     chain_1 = bern_fit.sample[:,0,:]
-    chain_1
+    chain_1.shape       # (1000, 8)
+    chain_1[0]          # sample first draw:
+                        # array([-7.99462  ,  0.578072 ,  0.955103 ,  2.       ,  7.       ,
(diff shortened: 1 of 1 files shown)


docs/getting_started.rst
```diff
@@ -152,6 +152,7 @@ the index '0' corresponds to the first chain in the ``StanFit``:
 .. code-block:: python
 
     chain_1 = bern_fit.sample[:,0,:]
+    chain_1
```
ahartikainen: maybe you could add example what is seen with this or maybe use `.shape` (I don't remember what this should show)

    chain_1.shape  # (1000,)

mitzimorris: while this illustrates the structure of the numpy.ndarray, it doesn't fit will with the downstream analysis workflow.  a better example would be showing how to extract a single draw - either at random (random draw, random chain, all columns) - or the 1st draw from the 1st chain.

annapaux: I just put in both as an example - I thought a line more shouldn't hurt and in my experience the more detailed the tutorial the better.

PR #109, opened 2019-08-21
Bugfix - Update default pbar values

ahartikainen:
#### Submission Checklist

- [ ] Run unit tests

#### Summary

There was a problem if `sampling_iters` or `warmup_iters` was not defined expicitly, which caused exception in the thread. Now the numbers are inferred on the fly, if some of the values are None. This should work, even if CmdStan would change the default values.

ThreadPoolExecutor then ate the exception message, so now I added try - except block to progress bar parsing --> if it fails, it will throw a warning, and keep on going normally. So if in future something "fails", it will still work if the CmdStan part works correctly.

#### Copyright and Licensing

Ari Hartikainen


ahartikainen:
@mitzimorris what kind of tests should we try?

If we want to run sampling with tqdm, we would need to install it.

Maybe we can fix travis / appveyor so it will `pip install tqdm` after the normal tests are run (excluding tqdm test), and then run some specific test with tqdm installed?

ahartikainen:
We still need to make sure that `thin>1` will work.

Do you want to try on mac how the output looks? On Windows it looks ok. I can still try with Ubuntu.

mitzimorris:
I'll look at thin.  the other thing to consider is when `show_params=True`, setting the CmdStan `refresh` argument to a small number will result in a greater resolution on the progress bar.
this could be kludged in brute force by appending `refresh=X` somewhere in `_run_cmdstan`

ahartikainen:
Yeah, we could do that (unless user has set that value)

Maybe step every % or 0.5%

mitzimorris:
users can't set value of 'refresh' - I didn't expose that command line argument since the subprocess call swallows up the console output.

ahartikainen:
Ok, then we can add it when show_progress is used.

ahartikainen:
I think the code is now ready.

Let's still make sure it samples looks good locally.

mitzimorris:
tested it on my machine last night - it works!
I love this feature, but am going to do a careful review and try to figure out more unit tests.

ahartikainen:
Sounds good. Yeah adding more unit tests is a good idea.

ahartikainen:
What unit tests would you want to be implemented?

mitzimorris:
> What unit tests would you want to be implemented?
not sure - need to go over your code and tqdm more.
right now I'm traveling and won't be able to do anything for the next 7 or 8 days
just to let you know

ahartikainen:
No worries, we can come back to this when you have time.

cmdstanpy/model.py
```diff
@@ -652,91 +677,130 @@ def _read_progress(self, proc: subprocess.Popen, pbar: List[Any]) -> bytes:
...
-                        # refresh warmup and close the progress bar if
-                        # all the warmup samples are seen
-                        if refresh_warmup:
-                            pbar_warmup.refresh()
-                            if count_warmup == num_warmup:
+        try:
+            # iterate while process is sampling
+            # process stdout line by line
+            while proc.poll() is None:
+                # read line
```
mitzimorris: most of the line-by-line commentary is unnecessary - comment on line #681 is sufficient.
should be removed.

cmdstanpy/model.py
```diff
@@ -652,91 +677,130 @@ def _read_progress(self, proc: subprocess.Popen, pbar: List[Any]) -> bytes:
...
+                            sampling_cumulative_count += count
+            # update warmup pbar if needed
+            if warmup_cumulative_count:
+                pbar_warmup.update(warmup_cumulative_count)
+                pbar_warmup.refresh()
+            # update sampling pbar if needed
+            if sampling_cumulative_count:
+                pbar_sampling.update(sampling_cumulative_count)
+                pbar_sampling.refresh()
+
```
mitzimorris: also true for comments here, excepting comments about warmup vs. sampling lines #725-#726

test/example_script.py
```diff
@@ -0,0 +1,33 @@
+"""Example script to be runned after tests, tqdm installed."""
```
mitzimorris: script is good start towards automated test of tqdm.  OK for now.

cmdstanpy/model.py
mitzimorris: completely understand - I do this too.
merged this yesterday - we can do the cleanup via another PR.

PR #115, opened 2019-09-14
use Fortran order for utils.jsondumps

ahartikainen:
#### Submission Checklist

- [x] Run unit tests

#### Summary

Transform nD arrays with Fortran order

    np.array(...).T.tolist()

#### Copyright and Licensing

Ari Hartikainen

### DIFF
--- a/cmdstanpy/utils.py
+++ b/cmdstanpy/utils.py
@@ -193,9 +193,12 @@ def _rdump_array(key: str, val: np.ndarray) -> str:
 def jsondump(path: str, data: Dict) -> None:
     """Dump a dict of data to a JSON file."""
     for key, val in data.items():
-        if isinstance(val, np.ndarray) and val.size > 1:
-            # .T for Fortran order
-            data[key] = val.T.tolist()
+        if isinstance(val, np.ndarray):
+            if val.size > 1:
+                # .T for Fortran order
+                data[key] = val.T.tolist()
+            else:
+                data[key] = []
     with open(path, 'w') as fd:
         json.dump(data, fd)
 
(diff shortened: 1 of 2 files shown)


ahartikainen:
We still need to test that this works correctly. (Fortran order) --> test for nD data

ahartikainen:
Ready for review.

Do we need to checkout also json input or are we only reading symmetric data?

ahartikainen:
This group

        if isinstance(val, np.ndarray):
            if val.size > 1:
                # .T for Fortran order
                data[key] = val.T.tolist()
            else:
                data[key] = []

should we remove empty keys or keep them as empty?

mitzimorris:
what is the issue here?
does this change anything about the resulting .json file on disk?
my understanding is that in JSON, multi-dim arrays (lists of list) are written in row-major order (i.e., C-style) and the CmdStan is going to process the json file accordingly.

ahartikainen:
Ok (which is great news), if that is true, then there is only the issue with empty array. Do we remove those parameters?

mitzimorris:
good call - we should allow this.

ahartikainen:
So we should update jsondump (other PR)

```
def jsondump(path: str, data: Dict) -> None:
    """Dump a dict of data to a JSON file."""
    empty_keys = []
    for key, val in data.items():
        if isinstance(val, np.ndarray):
            val = val.tolist()
            data[key] = val
        if (isinstance(val, (list, tuple)) and not val) or val is None:
            empty_keys.append(key)
    for key in empty_keys:
        del data[key]
    with open(path, 'w') as fd:
        json.dump(data, fd)
```

I will close this PR.

PR #117, opened 2019-09-15
Refactor/112 optimize return obj

mitzimorris:
#### Submission Checklist

- [x] Run unit tests

#### Summary

Refactor StanFit object in order to handle different CmdStan methods.

+ enum class `Method` makes explicit the set of supported methods.

+ method `sample` returns `StanFit` object (as before)
+ method `optimize` returns `StanMLE` object
+ method `run_generated_quantities` returns `StanQuantities` object.
+ class `RunSet` handles calls CmdStan, manages csv and console output files, and  checks return codes.  This object wraps a CmdStanArgs object and is passed into the `Model` class method `_run_cmdstan`.   It is passed into the constructor for the `StanFit`, `StanMLE`, and `StanQuantities` objects.

Adding both `optimize` and `run_generated_quantities` object resulted in the StanFit object becoming an omnibus kind of object and required `is_optimizing` and `is_sampling` logic in order to prevent the user from calling the wrong accessor function.   Creating seperate objects for each method make it clear what information is returned by each and how it should be parsed out of the stan-csv file(s).

This change will make it easier to implement `fixed_params` and `advi` methods.  In each case, we can define a new class (as needed) with attributes appropriate to the information returned by the method.

#### Copyright and Licensing

Columbia University


cmdstanpy/model.py
```diff
@@ -459,19 +461,29 @@ def sample(
...
-                # These need update if CmdStan changes the defaults
-                # refresh every 0.5% --> 200 updates
+                # need 200 progress bar updates - 100 per warmup, sampling
+                # refresh = total iters * (1/200)
+                default_warmup = 1000
+                default_sampling = 1000
+                default_total = default_warmup + default_sampling
                 if warmup_iters is None and sampling_iters is None:
-                    refresh = 10  # 2000 // 200 => 0.5 %
+                    refresh = int(math.floor(default_total * 0.005))
```
ahartikainen: I think `int(default_total * 0.005)` is enough, it will automatically floor the value.

cmdstanpy/model.py
```diff
@@ -459,19 +461,29 @@ def sample(
...
                 elif warmup_iters is None:
-                    refresh = (1000 + sampling_iters) // 200
+                    refresh = int(
+                        math.floor((default_warmup + sampling_iters) * 0.005)
+                    )
```
ahartikainen: just int

cmdstanpy/model.py
```diff
@@ -459,19 +461,29 @@ def sample(
...
                 elif sampling_iters is None:
-                    refresh = (warmup_iters + 1000) // 200
+                    refresh = int(
+                        math.floor((warmup_iters + default_sampling) * 0.005)
+                    )
```
ahartikainen: just int

cmdstanpy/model.py
```diff
@@ -459,19 +461,29 @@ def sample(
...
                 else:
-                    refresh = max((warmup_iters + sampling_iters) // 200, 1)
-
-                # disable logger for console (temporary)
-                # use tqdm
+                    refresh = max(
+                        int(
+                            math.floor((warmup_iters + sampling_iters) * 0.005),
```
ahartikainen: just int

cmdstanpy/model.py
```diff
@@ -524,7 +536,7 @@ def sample(
                                 tqdm_pbar = tqdm.tqdm_notebook
                             except ImportError:
                                 msg = (
-                                    "'tqdm_notebook' could not be imported. "
+                                    'Cannot import package tqdm_notebook.\n'
```
ahartikainen: It tries to import `tqdm.tqdm_notebook` module, but I think this message is ok.

mitzimorris:
> Or should it be MAP -> Maximum A Posterior estimate?

it's a penalized MLE - optimization turns off computation of Jacobians, so it's not a MAP estimate because it doesn't account for change of variables.

cmdstanpy/model.py
mitzimorris: thanks - will change

cmdstanpy/model.py
mitzimorris: will change

cmdstanpy/model.py
mitzimorris: will make it explicit 'tqdm.tqdm_notebook'

PR #118, opened 2019-09-16
[WIP] update jsondump and rdump

ahartikainen:
#### Submission Checklist

- [ ] Run unit tests

#### Summary

Enable size 0 and 1 arrays.

#### Copyright and Licensing

Ari Hartikainen

### DIFF
--- a/cmdstanpy/utils.py
+++ b/cmdstanpy/utils.py
@@ -63,7 +63,7 @@ class MaybeDictToFilePath(object):
                     dir=TMPDIR, prefix='', suffix='.json'
                 )
                 self._logger.debug('input tempfile: %s', data_file)
-                jsondump(data_file, o)
+                rdump(data_file, o)
                 self._paths[i] = data_file
                 self._unlink[i] = True
             elif isinstance(o, str):
@@ -331,7 +331,9 @@ def jsondump(path: str, data: Dict) -> None:
             data[key] = val
         if isinstance(val, Sequence) and not val:
             raise ValueError(
-                'variable: {}, error: empty array not allowed'.format(val)
+                'variable: {}, error: empty array not allowed with JSON interface'.format(
+                    val
+                )
             )
     with open(path, 'w') as fd:
         json.dump(data, fd)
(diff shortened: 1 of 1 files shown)


ahartikainen:
I think there are two different bugs

size 1 real/int array/vector vs real/int
Empty arrays as data. Can CmdStan have them?

ahartikainen:
There is something weird going on with PR (deleted functions etc), so I need to figure that one out.

@mitzimorris do you know what should be the correct way to handle these

data array/vector with size 0
data array/vector with size 1

Can cmdstanpy currently handle these? From both interfaces JSON/RDump?

ps. Also if you want you can update this functionality to the current bigger PR, and I can then close this one

mitzimorris:
> data array/vector with size 0

CmdStan allows empty arrays (although couldn't make this happen using RStan).

however, the current JSON parser as implemented doesn't allow them - mea culpa -
I implemented an ad-hoc JSON parser back in 2014 because we couldn't find a C++ JSON parsing library that had implemented a SAX parser.  now there is one and there's a CmdStan issue on using this parser that I need to take care of ASAP.

> data array/vector with size 1

CmdStan has no problem with these.

here's the model - "test_inputs.stan":

```
data {
  int<lower=0> N;
  int<lower=0> y[N];
  vector<lower=0,upper=1>[N] x;
}
transformed data {
  print(N);
  print(y);
  print(x);
}
```

data file "test_zero_vecs.data.R":

```
N <- 0
y <- c()
x <- c()
```

data file "test_zero_vecs.data.json":
```
{
    "N" : 0,
    "y" : [],
    "x" : [],
}
```

ahartikainen:
Yeah, currently we can't have 0-sized array for CmdStan --> Let's not allow them now --> Let's raise an error if user tries to use one. (Or we can just let CmdStan fail, and after fix, it will work)

I updated the behavior for the 1-sized arrays. There was an off-by-one error with length handling.

I test against Sequence --> list, tuple and some other special formats -> json transform them to `[...]` and for RDump I call `np.asarray` for ndarray and sequence --> all are transformed to numpy array (unless they are already).

mitzimorris:
> Yeah, currently we can't have 0-sized array for CmdStan --> Let's not allow them now 

if we do this, we need to
a) test that Rdump files created by CmdStanPy's utility function 'rdump' work for 0-length arrays etc
b) change the behavoir of 'MaybeDictToFilePath' to generate rdump files instead of json files.

this should avoid the problem unless the user calls  `jsondump` directly.

in addition, we need to file an issue so that once the fix is in CmdStan and the next CmdStan release is out (scheduled for October), we update CmdStanPy so that everything works as it should.

PR #132, opened 2019-10-10
Feature/88 go faster

mitzimorris:
#### Submission Checklist

- [x] Run unit tests

#### Summary

Tightened up code which reads in stan csv files.  Added unit test that can be used to compare runtimes.

Conclusion is that underlying problem is lack of binary file format for output files.   Timing tests show that CmdStan and CmdStanPy have about the same runtime to run one chain of the sampler on a non-trivial model (2000 parameters, 1000 sampling iterations).

#### Copyright and Licensing


ahartikainen:
How much faster is this?

mitzimorris:
trivially faster.

mitzimorris:
in CmdStanPy, for the `sample` method, the set of stan-csv files are read twice - first by `_validate_csv_files`, then `_assemble_sample`

`_validate_csv_files` does the following:

-  parses the initial comments to get the sampler config, 
- parses out the column header,
- skips the warmup draws, if any
- parses the metric out of comment lines, (as needed)
- checks the output draws to make sure that the sample has the correct shape - i.e. number of draws and columns.

`_assemble_sample` gets the output draws and does the conversion from ascii to float.

for a sample of 1000 draws over 2000 parameters, changes here only sped things up by milliseconds -  the processing time in `_assemble_sample`  - multiple runs showed a consistent speedup, but not by much - on my machine things went from 1.4 sec to 1.3 seconds -

```
for x in {1..10}; do python -m unittest test.test_sample.StanMCMCTest.test_assemble_sample_big; done
```

OTOH, this does clean up one corner of the code.

rok-cesnovar:
What is the largest hurdle with adding a binary format for output in cmdstan? I would be interested in adding that to cmdstan. I guess this would help both cmdstanpy and cmdstanr. Is it the lack of consensus on the exact format? Would httpstan's [writer proto](https://github.com/stan-dev/httpstan/blob/master/protos/callbacks_writer.proto) be a good start?

mitzimorris:
I would love to see binary format for output in cmdstan.

mitzimorris:
closing for now, we can revisit later.

PR #164, opened 2019-11-03
Feature/prevent path duplicate

guihigashi:
#### Submission Checklist

- [ ] Run unit tests

#### Summary

The `utilscxx_toolchain_path` appends toolchain dirs to PATH even if they're already there.

This keeps the prepended non-duplicated entries only, which is relevant in the case of `find.exe` from `tool_path` having precedence over `C:\Windows\System32\find.exe`.

#### Copyright and Licensing


ahartikainen:
How is this an issue? Is it a problem if one compiles multiple models in one session?

guihigashi:
yes, in the case multiple models in one session. It's not really an issue, but let's say i'm writing several scripts (one per model) and each script calls `utils.cxx_toolchain_path()`. It'd be prepending the same entries over and over... since this behavior is avoidable, maybe it's nice to avoid...

Issue #193, opened 2020-01-05
Stan model error output is unreadable


evgenyneu:
#### Summary:

If there are syntax errors in *.stan file, the error output is messy.

#### Description:

Use the code below and run it from the Terminal

```
python eight_schools.py
```

The eight_schools.stan model has a syntax error (missing semicolon in declaration of `tau` parameter), so Stan complains. The problem is that the error messages are not printed very nicely and are hard to read:

```
INFO:cmdstanpy:stan to c++ (/Users/evgenii/Documents/2019_summer_project/logbook/a2020/a01/05_python_model/code/for_github/eight_schools.hpp)
ERROR:cmdstanpy:file /Users/evgenii/Documents/2019_summer_project/logbook/a2020/a01/05_python_model/code/for_github/eight_schools.stan, exception RuntimeError('ERROR\n SYNTAX ERROR, MESSAGE(S) FROM PARSER:\n error in \'/Users/evgenii/Documents/2019_summer_project/logbook/a2020/a01/05_python_model/code/for_github/eight_schools.stan\' at line 10, column 2\n  -------------------------------------------------\n     8:   real mu;                // population treatment effect\n     9:   real<lower=0> tau       // MISSING SEMICOLON, standard deviation in treatment effects\n    10:   vector[J] eta;          // unscaled deviation from mu by school\n         ^\n    11: }\n  -------------------------------------------------\n\nPARSER EXPECTED: ";"\n\nmake: *** [/Users/evgenii/Documents/2019_summer_project/logbook/a2020/a01/05_python_model/code/for_github/eight_schools.hpp] Error 253 ')
ERROR:cmdstanpy:model compilation failed
Traceback (most recent call last):
  File "eight_schools.py", line 3, in <module>
    model = CmdStanModel(stan_file="eight_schools.stan")
  File "/opt/miniconda3/lib/python3.7/site-packages/cmdstanpy/model.py", line 138, in __init__
    self._stan_file
ValueError: unable to compile Stan model file: /Users/evgenii/Documents/2019_summer_project/logbook/a2020/a01/05_python_model/code/for_github/eight_schools.stan
deleting tmpfiles dir: /var/folders/4s/0b1x7wr50dbdvx5skpv8j0sm0000gn/T/tmpfsltbmq5
done
```

[...]

PR #194, opened 2020-01-06
Make Stan model syntax errors readable

evgenyneu:
#### Submission Checklist

- [x] Run unit tests

#### Summary

It was not easy to read the error messages for .stan file with syntax errors. The error messages were all printed out in one line because the new line characters were escaped. I've changed it so the new lines are shown as new lines, so the errors are now more readable. :) See #193.

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting (this will be you or your assignee, such as a university or company): Me

Yes, agree

### DIFF
--- a/cmdstanpy/model.py
+++ b/cmdstanpy/model.py
@@ -257,7 +257,7 @@ class CmdStanModel:
                         do_command(cmd, cmdstan_path(), logger=self._logger)
                     except RuntimeError as e:
                         self._logger.error(
-                            'file %s, exception %s', stan_file, e
+                            'file %s, exception %s', stan_file, str(e)
                         )
                         compilation_failed = True
 
--- a/cmdstanpy/model.py
+++ b/cmdstanpy/model.py
@@ -257,7 +257,7 @@ class CmdStanModel:
                         do_command(cmd, cmdstan_path(), logger=self._logger)
                     except RuntimeError as e:
                         self._logger.error(
-                            'file %s, exception %s', stan_file, repr(e)
+                            'file %s, exception %s', stan_file, e
                         )
                         compilation_failed = True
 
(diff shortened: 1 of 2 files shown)


cmdstanpy/model.py
```diff
@@ -257,7 +257,7 @@ def compile(self, opt_lvl: int = 3, force: bool = False) -> None:
                         do_command(cmd, cmdstan_path(), logger=self._logger)
                     except RuntimeError as e:
                         self._logger.error(
-                            'file %s, exception %s', stan_file, repr(e)
+                            'file %s, exception %s', stan_file, e
```
mitzimorris: https://stackoverflow.com/a/1438297 - suggests that fix is changing `repr` to `str` - does this do the right thing w/r/t newlines?

test/test_model.py
```diff
@@ -116,17 +118,27 @@ def test_model_none(self):
         with self.assertRaises(ValueError):
             _ = CmdStanModel(exe_file=None, stan_file=None)
 
-    def test_model_bad(self):
+    def test_model_file_does_not_exist(self):
```
mitzimorris: yes - good call!

cmdstanpy/model.py
evgenyneu: I think leaving it as `e` is the same as `str(e)`, but I can change it to `str(e)` if you want. 

> does this do the right thing w/r/t newlines?

Yes, before, `repr` just escaped all the special characters, so newline became "\n" for example. The `str` does escape newlines, and that's what we want here.

mitzimorris: great, please use str(e)

evgenyneu: done

PR #197, opened 2020-01-12
Use single progress bar

evgenyneu:
#### Submission Checklist

- [x] Run unit tests

#### Summary

* Use a single progress bar instead of separate bars for warmup and sampling. This simplifies code since cmdstan output also combines warmup and sampling.

* Change the description of the bar from "warmup", to "sample" to "done" to indicate the stage of sampling.

* For Jupyter only, make each bar green as soon as it's done.

* I've deleted a lot of progress bar code following comment "read and process rest of the stdout if needed", I could not figure out what it was for, and everything worked for me without it.

#### In terminal

cmdstanpy_single_bar_terminal

#### In Jupyter

cmdstanpy_single_bar_jupyter

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting (this will be you or your assignee, such as a university or company): me

Yes, agree

## Code

To test the progress bar, run

```
python eight_schools.py
```

### eight_schools.py

```python
from cmdstanpy import CmdStanModel

model = CmdStanModel(stan_file="eight_schools.stan")

data = {
    "J": 8,
    "y": [28,  8, -3,  7, -1,  1, 18, 12],
    "sigma": [15, 10, 16, 11,  9, 11, 10, 18]
}

fit = model.sample(data=data, show_progress=True,
                   chains=4, cores=2,
                   sampling_iters=15000, warmup_iters=15000)

print(fit.summary())
```

### eight_schools.stan

```stan
data {
  int<lower=0> J;         // number of schools
  real y[J];              // estimated treatment effects
  real<lower=0> sigma[J]; // standard error of effect estimates
}
parameters {
  real mu;                // population treatment effect
  real<lower=0> tau;      // standard deviation in treatment effects
  vector[J] eta;          // unscaled deviation from mu by school
}
transformed parameters {
  vector[J] theta = mu + tau * eta;        // school treatment effects
}
model {
  target += normal_lpdf(eta | 0, 1);       // prior log-density
  target += normal_lpdf(y | theta, sigma); // log-likelihood
}
```

### DIFF
--- a/cmdstanpy/model.py
+++ b/cmdstanpy/model.py
@@ -601,21 +601,6 @@ class CmdStanModel:
             try:
                 import tqdm
 
-                # progress bar updates - 100 per warmup, sampling
-                if fixed_param or not adapt_engaged or warmup_iters == 0:
-                    num_updates = 100
-                    w_iters = 0
-                else:
-                    num_updates = 200
-                    if warmup_iters is None:
-                        w_iters = 1000
-                    else:
-                        w_iters = warmup_iters
-                s_iters = sampling_iters
-                if s_iters is None:
-                    s_iters = 1000
-                refresh = max(int((s_iters + w_iters) // num_updates), 1)
-                # disable logger for console (temporary) - use tqdm
                 self._logger.propagate = False
             except ImportError:
                 self._logger.warning(
@@ -681,6 +666,7 @@ class CmdStanModel:
                                 tqdm_pbar = tqdm.tqdm
                         else:
                             tqdm_pbar = tqdm.tqdm
+
                         # enable dynamic_ncols for advanced users
                         # currently hidden feature
                         dynamic_ncols = os.environ.get(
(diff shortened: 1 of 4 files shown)


ahartikainen:
A lot of the code was probably 'over-optimized'. The problem is that we only update everystep that Stan outputs update and if user sets that value to some weird cornercase then our bar might fail.

So I think it is safe to remove those bits and let's fix these issues when(if) they are seen.

cmdstanpy/model.py
```diff
@@ -690,24 +676,25 @@ def sample(
                             dynamic_ncols = False
                         else:
                             dynamic_ncols = True
-                        pbar = [
-                            # warmup
-                            tqdm_pbar(
+
+                        # Total number of iterations (warmup + sample)
+                        # 2000 is default in cmdstan
+                        total_iters = 2000
```
ahartikainen: Will this work with non default values?

evgenyneu: Oops, it did work when both the number of warmup and sampling iterations were omitted by the user. But it did not work if only one of them was omitted.

Fixed here:
https://github.com/stan-dev/cmdstanpy/pull/197/commits/79169bc0e7c5230852749f9bf27bb9977c62ce01

evgenyneu: Nah, this is better

https://github.com/stan-dev/cmdstanpy/pull/197/commits/749ed90275081b99a98728ef8d82a5d726c63e44

Stan outputs 12 / 12345, so we know the total number of iterations is 12345. No need to guess that. This is how it worked before, I think.

evgenyneu:
The build failed because `tqdm` Python library is not installed on build machine. Can we install it?

test/test_model.py
```diff
@@ -3,9 +3,11 @@
 import os
 import shutil
 import unittest
+from unittest.mock import Mock
```
mitzimorris: does unittest.mock work with PyTest?

cmdstanpy/model.py
```diff
@@ -1030,103 +1007,39 @@ def _read_progress(
...
-                                pbar_warmup.refresh()
-                                refresh_warmup = False
-                            # update values to full
-                            count, count_sampling = (
-                                raw_count - num_warmup - count_sampling,
-                                raw_count - num_warmup,
-                            )
-                            pbar_sampling.update(count)
+                        current_count = int(match.group(1))
+                        total_count = int(match.group(2))
```
mitzimorris: awesome cleanup!

test/test_model.py
ahartikainen: Yes. Here is also one example from ArviZ https://github.com/arviz-devs/arviz/blob/master/arviz/tests/test_utils.py

test/test_model.py
```diff
@@ -196,6 +198,36 @@ def test_model_compile_includes(self):
         model3 = CmdStanModel(stan_file=stan)
         self.assertTrue(model3.exe_file.endswith(exe.replace('\\', '/')))
 
+    def test_read_progress(self):
```
ahartikainen: Thanks for the test

mitzimorris:
many thanks!

PR #216, opened 2020-02-10
Bugfix/188 cmdstan env var

### LINKED ISSUE #188: Setting CMDSTAN environment variable causes test to fail
#### Summary:
cmdstanpy properly finds cmdstan when `CMDSTAN` environment is set, but one test does fail. 

#### Description:
There is a test to see that cmdstan is located at `~/.cmdstanpy/cmdstan-<version>`, but a user can change the location of cmdstan by setting `CMDSTAN` environment variable.

https://github.com/stan-dev/cmdstanpy/blob/f20c9bb9002387dfc662fe84d98d24f9df1850e8/test/test_utils.py#L35-L40

#### Current Version:
master

mitzimorris:
#### Submission Checklist

- [x] Run unit tests

#### Summary

added test logic to `test_utils.py` tests for location of cmdstan.
- `test_default_path` checks `CMDSTAN` env var is set
- `test_set_path` also checks that `CMDSTAN` env var is set after calling this function.

also changed logic to `utils.py` function `cmdstan_path` so that `CMDSTAN` env var will be set to default path. this is consistent with behavior in `set_cmdstan_path` which also sets the env var after validating the path.

Fixes #188.

#### Copyright and Licensing

### DIFF
--- a/test/test_utils.py
+++ b/test/test_utils.py
@@ -33,11 +33,15 @@ DATAFILES_PATH = os.path.join(HERE, 'data')
 
 
 class CmdStanPathTest(unittest.TestCase):
+
     def test_default_path(self):
-        abs_rel_path = os.path.expanduser(
-            os.path.join('~', '.cmdstanpy', 'cmdstan')
-        )
-        self.assertTrue(cmdstan_path().startswith(abs_rel_path))
+        if 'CMDSTAN' in os.environ:
+            self.assertEqual(cmdstan_path(), os.environ['CMDSTAN'])
+        else:
+            abs_rel_path = os.path.expanduser(
+                os.path.join('~', '.cmdstanpy', 'cmdstan')
+                )
+            self.assertTrue(cmdstan_path().startswith(abs_rel_path))
 
     def test_non_spaces_location(self):
         good_path = os.path.join(TMPDIR, 'good_dir')
--- a/test/test_utils.py
+++ b/test/test_utils.py
@@ -60,7 +60,6 @@ class CmdStanPathTest(unittest.TestCase):
                 if 'CMDSTAN' in os.environ:
                     del os.environ['CMDSTAN']
 
-
     def test_non_spaces_location(self):
         good_path = os.path.join(TMPDIR, 'good_dir')
         with TemporaryCopiedFile(good_path) as (pth, is_changed):
--- a/test/test_utils.py
+++ b/test/test_utils.py
@@ -81,12 +81,16 @@ class CmdStanPathTest(unittest.TestCase):
         shutil.rmtree(bad_path, ignore_errors=True)
 
     def test_set_path(self):
-        install_dir = os.path.expanduser(os.path.join('~', '.cmdstanpy'))
-        install_version = os.path.expanduser(
-            os.path.join(install_dir, get_latest_cmdstan(install_dir))
-        )
-        set_cmdstan_path(install_version)
-        self.assertEqual(install_version, cmdstan_path())
+        if 'CMDSTAN' in os.environ:
+            self.assertEqual(cmdstan_path(), os.environ['CMDSTAN'])
+        else:
+            install_dir = os.path.expanduser(os.path.join('~', '.cmdstanpy'))
+            install_version = os.path.expanduser(
+                os.path.join(install_dir, get_latest_cmdstan(install_dir))
+                )
+            set_cmdstan_path(install_version)
+            self.assertEqual(install_version, cmdstan_path())
+            self.assertEqual(install_version, os.environ['CMDSTAN'])
 
     def test_validate_path(self):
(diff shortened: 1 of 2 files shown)


test/test_utils.py
```diff
@@ -33,11 +33,22 @@
...
 
 class CmdStanPathTest(unittest.TestCase):
+
     def test_default_path(self):
-        abs_rel_path = os.path.expanduser(
-            os.path.join('~', '.cmdstanpy', 'cmdstan')
-        )
-        self.assertTrue(cmdstan_path().startswith(abs_rel_path))
+        if 'CMDSTAN' in os.environ:
+            self.assertEqual(cmdstan_path(), os.environ['CMDSTAN'])
```
ahartikainen: will function set the env variable here and then it is taken for comparison. Should you change the order or is this wanted behaviour.

Also, should we reset the path, if it is not the default at start?

try: finally block?

test/test_utils.py
```diff
@@ -33,11 +33,22 @@
...
+            del os.environ['CMDSTAN']
+            self.assertFalse('CMDSTAN' in os.environ)
+            set_cmdstan_path(path)
+            self.assertEqual(cmdstan_path(), path)
+            self.assertTrue('CMDSTAN' in os.environ)
+        else:
+            abs_rel_path = os.path.expanduser(
+                os.path.join('~', '.cmdstanpy', 'cmdstan')
+                )
+            self.assertTrue(cmdstan_path().startswith(abs_rel_path))
```
ahartikainen: Maybe use (os.abspath +) os.relpath comparison here?

relative path from abs_rel_path to env path should be empty (string) which should eval to False (bool), but you could also use explicit `==""`.

test/test_utils.py
wesbarnett: Shouldn't `CMDSTAN` be guaranteed to be set at this point (line 195 in utils.py above)?

test/test_utils.py
```diff
@@ -70,12 +81,16 @@ def test_non_spaces_location(self):
         shutil.rmtree(bad_path, ignore_errors=True)
 
     def test_set_path(self):
-        install_dir = os.path.expanduser(os.path.join('~', '.cmdstanpy'))
-        install_version = os.path.expanduser(
-            os.path.join(install_dir, get_latest_cmdstan(install_dir))
-        )
-        set_cmdstan_path(install_version)
-        self.assertEqual(install_version, cmdstan_path())
+        if 'CMDSTAN' in os.environ:
```
wesbarnett: If `CMDSTAN` is not set, then it should be a failure since if it is not set, utils.py should be setting it to the default location of cmdstan.

test/test_utils.py
wesbarnett: In other words, if it is not set, then the test should fail (or we have another test for that).

ahartikainen: yes, but this is a test. If user has set CMDSTAN to some specific (different than default) folder, runs tests and after tests notice that the CMDSTAN was changed.

Tests should not change anything global.

wesbarnett: When `cmdstan_path()` is called it sets `CMDSTAN`. So the line 39 will always be true.

wesbarnett: Ok I'm following it now.

mitzimorris: changed test so that it records value of CMDSTAN environment var before the test and uses finally block to reset accordingly.

test/test_utils.py
mitzimorris: I would prefer to leave this as is - is the objection to the "startswith" test?
relpath test didn't work for me, but samefile test did -
```
>>> foo = cmdstanpy.cmdstan_path()
>>> foo
'/Users/mitzi/.cmdstanpy/cmdstan-2.22.1'
>>> os.environ['CMDSTAN']
'/Users/mitzi/.cmdstanpy/cmdstan-2.22.1'
>>> os.path.relpath(os.environ['CMDSTAN'], foo)
'.'
>>> os.path.samefile(os.environ['CMDSTAN'], foo)
True
```

ahartikainen: yeah, startswith could fail when there are different mapping for the specific address.

samefile sounds totally fine

mitzimorris: samefile is too strict - this test as written is version agnostic - that's why it has to be "startswith"

ahartikainen: What do you mean by "this test as written is version agnostic"?

I think `samefile` just test if `relpath == "."` (not sure)

mitzimorris: the default path will include some version number - I need to change this test to use `utils.get_latest_cmdstan` - will do that, in which case this will work.

mitzimorris: nevermind - foggy brain - that's what the test is already doing - will change.

mitzimorris: OK, made changes per your suggestion - using `get_latest_version` to get full path to default cmdstan install.

PR #252, opened 2020-06-08
Feature/245 get var

mitzimorris:
#### Submission Checklist

- [x] Run unit tests

#### Summary
Add functions `stan_variable`, `stan_variables`, and `sampler_diagnostics` to CmdStanMCMC class.   Also added property `stan_var_dims` which gives a mapping from
program variable names to their dimensions.

Functions `stan_variable` and `stan_variables` ~~flatten out~~ stack the chains, i.e.,
given a sample 100 iters, 4 chains, scalar var, in the the underlying sample,
this has shape  (100,4)  but the returned variable has shape (400).

The `sampler_diagnostics` function leaves the chains dimension, because these
(some of) these diagnostics columns are used to determine convergence.

These functions return copies of the underlying draws ndarray, and the
function documentation mentions this explicitly. This would be a concern
only when the sample contains hundreds of millions of entries and the user
makes lots and lots of copies.

Profiled memory on a model with approx 150M entries to stress test using `psutil` lib - behavoir as expected.

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting (this will be you or your assignee, such as a university or company): Columbia University


cmdstanpy/stanfit.py
```diff
@@ -312,6 +315,15 @@ def column_names(self) -> Tuple[str, ...]:
...
         return self._column_names
 
+    @property
+    def stan_var_dims(self) -> Dict:
+        """
+        Dict mapping Stan program variable names to variable dimensions.
+        Scalar types have int value '1'.  Structured types have list of dims,
+        e.g.,  program variable ``vector[10] foo`` has entry ``('foo', [10])``.
+        """
+        return self._stan_var_dims
```
ahartikainen, line 325: this should return copy, otherwise

    print(a.stan_var_dims["a"]) --> 0
    res = a.stan_var_dims
    res["a"] = 3
    print(a.stan_var_dims["a"]) --> 3

ahartikainen: Also dims should be tuples

cmdstanpy/stanfit.py
```diff
@@ -547,21 +560,93 @@ def get_drawset(self, params: List[str] = None) -> pd.DataFrame:
...
-            len(self.column_names),
-            order='A',
-        )
-        drawset = pd.DataFrame(data=data, columns=self.column_names)
+        if self._drawset is None:
+            # pylint: disable=redundant-keyword-arg
+            data = self.sample.reshape(
+                (self.num_draws * self.runset.chains),
+                len(self.column_names),
+                order='A',
```
ahartikainen, line 568: Will this stack chains on top of each other or mix them? (in our normal usecase)

cmdstanpy/stanfit.py
```diff
@@ -547,21 +560,93 @@ def get_drawset(self, params: List[str] = None) -> pd.DataFrame:
...
+            )
+            self._drawset = pd.DataFrame(data=data, columns=self.column_names)
         if params is None:
-            return drawset
+            return self._drawset
         mask = []
         for param in params:
             for name in self.column_names:
                 if param == name or param == name.split('.')[0]:
                     mask.append(name)
```
ahartikainen, line 577: We can remove on for loop

```suggestion
        mask = []
        params = set(params)
        for name in self.column_names:
            if any(item in params for item in (name, name.split('.')[0]):
                mask.append(name)
```

cmdstanpy/stanfit.py
```diff
@@ -547,21 +560,93 @@ def get_drawset(self, params: List[str] = None) -> pd.DataFrame:
...
+                var_dims.extend(dims)
+                var_cols = [
+                    col for col in self._drawset if col.startswith(key + '.')
+                ]
+                result[key] = (
+                    self._drawset[var_cols]
+                    .to_numpy()
+                    .reshape(tuple(var_dims), order='A')
+                )
+        return result
```
ahartikainen, line 638: This can call the other function. 

Not sure if the `self.get_drawset()`is needed or can it also be in the `stan_variable` or is `self._stan_var_dims` filled in `self.get_drawset()`

```suggestion
        result = {}
        if self._drawset is None:
            self.get_drawset()
        for name in self.stan_var_dims:
            result[name] = self.stan_variable(name)
        return result
```

cmdstanpy/utils.py
```diff
@@ -629,6 +629,31 @@ def scan_column_names(fd: TextIO, config_dict: Dict, lineno: int) -> int:
...
+        if names[idx].endswith('__'):
+            pass
+        elif '.' not in names[idx]:
+            vars_dict[names[idx]] = 1
+        else:
+            vs = names[idx].split('.')
+            if idx < len(names) - 1 and names[idx + 1].split('.')[0] == vs[0]:
+                idx += 1
+                continue
+            dims = [int(vs[x]) for x in range(1, len(vs))]
```
ahartikainen, line 651: Does this work for ragged structures?

cmdstanpy/stanfit.py
mitzimorris: looks like there's no such thing as read-only properties in Python. :-(
in which case, this is a problem all over the place, no?

mitzimorris: requires recoding everything with __get__ and __set__?  per https://howto.lintel.in/how-to-create-read-only-attributes-and-restrict-setting-attribute-values-on-object-in-python/

cmdstanpy/stanfit.py
mitzimorris: the arrangement of items is draw1, chain1, draw2, chain1, .... drawN, chain1, draw1, chain2, draw2, chain2, ...., draw1, chainN, ...  drawN, chainN

is this stacked?

ahartikainen: Yes

cmdstanpy/stanfit.py
ahartikainen: Probably not, I think basic copy / deepcopy is enough

cmdstanpy/utils.py
mitzimorris: no, this won't work for ragged structures - they'd need to be coded differently.
does numpy allow ragged ndarrays?

ahartikainen: It kind of does, but really it doesn't

numpy mask might help.

But having non-ragged structure with nan might work

mitzimorris: there is a proposal for adding ragged arrays to the language - also tuples will be coming soon - so I think the next step will be to re-implement using xarray.

mitzimorris:
merging this version.
need to further understand how using xarray objects would affect downstream use.

PR #266, opened 2020-08-05
Bugfix/235 chain ids

mitzimorris:
#### Submission Checklist

- [x] Run unit tests

#### Summary

If sampler is run with a set of specified chain ids, use this id in the output file name, as advertised in the docs.
This requires passing making the list of chain_ids an argument to the RunSet object.

Because the `generate_quantities` method can take an  existing CmdStanMCMC sample object, exposed `chain_ids`
so that gq filenames will line up with sample filenames.

#### Copyright and Licensing

### DIFF
--- a/cmdstanpy/model.py
+++ b/cmdstanpy/model.py
@@ -398,7 +398,7 @@ class CmdStanModel:
             )
 
             dummy_chain_id = 0
-            runset = RunSet(args=args, chains=1)
+            runset = RunSet(args=args, chains=1, chain_ids=[1])
             self._run_cmdstan(runset, dummy_chain_id)
 
         if not runset._check_retcodes():
@@ -597,12 +597,12 @@ class CmdStanModel:
             chain_ids = [x + 1 for x in range(chains)]
         else:
             if isinstance(chain_ids, int):
-                if chain_ids < 0:
+                if chain_ids < 1:
                     raise ValueError(
-                        'Chain_id must be a non-negative integer value,'
+                        'Chain_id must be a positive integer value,'
                         ' found {}.'.format(chain_ids)
                     )
-                chain_ids = [chain_ids + i + 1 for i in range(chains)]
+                chain_ids = [chain_ids + i for i in range(chains)]
             else:
                 if not len(chain_ids) == chains:
                     raise ValueError(
@@ -691,7 +691,7 @@ class CmdStanModel:
                 refresh=refresh,
                 logger=self._logger,
             )
-            runset = RunSet(args=args, chains=chains)
+            runset = RunSet(args=args, chains=chains, chain_ids=chain_ids)
             pbar = None
             all_pbars = []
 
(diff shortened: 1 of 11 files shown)


test/test_runset.py
```diff
@@ -16,14 +16,15 @@ def test_check_retcodes(self):
         exe = os.path.join(DATAFILES_PATH, 'bernoulli' + EXTENSION)
         jdata = os.path.join(DATAFILES_PATH, 'bernoulli.data.json')
         sampler_args = SamplerArgs()
+        chain_ids = [1, 2, 3, 4]  # default
```
ahartikainen, line 19: are these always 1, 2, ... ?

cmdstanpy/model.py
```diff
@@ -818,16 +818,19 @@ def generate_quantities(
             sample_csv_files = mcmc_sample.runset.csv_files
             sample_drawset = mcmc_sample.get_drawset()
             chains = mcmc_sample.chains
+            chain_ids = mcmc_sample.chain_ids
         elif isinstance(mcmc_sample, list):
+            if len(mcmc_sample) < 1:
+                raise ValueError('MCMC sample cannot be empty list')
             sample_csv_files = mcmc_sample
+            chains = len(sample_csv_files)
+            chain_ids = [x + 1 for x in range(chains)]
```
ahartikainen, line 827: Does chain_ids affect the results anyway?

test/test_runset.py
mitzimorris: yes - in this test, we're constructing the default runset as opposed to trying to run the sampler.

cmdstanpy/model.py
mitzimorris: yes - the id is captured by CmdStan and passed in to all the stan::services methods - e.g. - https://github.com/stan-dev/cmdstan/blob/c709a0927eb191655b29764688ed1b0a75e36a0c/src/cmdstan/command.hpp#L284

this is used to advance the RNG  -
https://github.com/stan-dev/stan/blob/ae35c12b0f93fd8253b72b84fcb26ec91ae8d52a/src/stan/services/diagnose/diagnose.hpp#L29

the chain_id makes a convenient loop index, but really, it's there so that we can extend the notion of "split R-hat" to multiple chains all with the same seed.

that's why this needs to be correctly implemented.  the next thing to do is to write a little python script that can be called via SLURM to show how to run 100 or 1000 chains on a cluster.

mitzimorris:
good questions!  I'm going to take a look at what we do if the user doesn't specify the seed - perhaps in that case we should choose a random seed on the CmdStanPy and use it for all chains?

cmdstanpy/model.py
mitzimorris: just check with Bob on what the right thing to do, which is this:

if the user doesn't specify the seed, then the constructor should choose a random seed based on the timestamp (as in CmdStan).

mitzimorris:
just checked with Bob on what the right thing to do, which is this:

if the user doesn't specify the seed, then the constructor should choose a random seed based on the timestamp (as in CmdStan) which should be passed in to the CmdStan args so that it is the same for all chains.

mitzimorris:
checked the code - indeed, we do make sure to use the same seed across all chains.
improved (?) the wording on the error message for specifying per-chain seed/chain-ids.

I think the existing docs are good enough as it.  the example Python script for distributing jobs across a cluster and discussion should go on Discourse - for now, it doesn't *need* to be in the docs.

mitzimorris:
hi @ahartikainen - re-requesting review.

PR #271, opened 2020-08-09
Feature/238 validate csv opt

mitzimorris:
#### Submission Checklist

- [x] Run unit tests

#### Summary

Add argument `validate_csv` to CmdStanModel `sample` method.

Tests locally and on cluster indicates that skipping validation of CSV file
results in a 10% speedup.

#### Copyright and Licensing

### DIFF
--- a/cmdstanpy/stanfit.py
+++ b/cmdstanpy/stanfit.py
@@ -5,6 +5,7 @@ import re
 import shutil
 import copy
 import logging
+import math
 from typing import List, Tuple, Dict
 from collections import Counter, OrderedDict
 from datetime import datetime
@@ -272,7 +273,13 @@ class CmdStanMCMC:
     Container for outputs from CmdStan sampler run.
     """
 
-    def __init__(self, runset: RunSet, validate_csv: bool = True) -> None:
+    # pylint: disable=too-many-instance-attributes
+    def __init__(
+        self,
+        runset: RunSet,
+        validate_csv: bool = True,
+        logger: logging.Logger = None,
+    ) -> None:
         """Initialize object."""
         if not runset.method == Method.SAMPLE:
             raise ValueError(
@@ -280,6 +287,7 @@ class CmdStanMCMC:
                 'found method {}'.format(runset.method)
             )
         self.runset = runset
+        self._logger = logger or get_logger()
         # copy info from runset
         self._is_fixed_param = runset._args.method_args.fixed_param
         self._iter_sampling = runset._args.method_args.iter_sampling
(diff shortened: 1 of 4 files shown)


cmdstanpy/stanfit.py
```diff
@@ -364,7 +398,15 @@ def metric(self) -> np.ndarray:
         Metric used by sampler for each chain.
         When sampler algorithm 'fixed_param' is specified, metric is None.
         """
-        if not self._is_fixed_param and self._metric is None:
+        if self._is_fixed_param:
+            return None
+        if not self._validate_csv and self._metric is None:
+            self._logger.warning(
+                'csv files not yet validated, run method validate_csv_files()'
+                ' in order to retrieve sample metatdata.'
```
ahartikainen, line 406: metadata

cmdstanpy/stanfit.py
```diff
@@ -374,7 +416,15 @@ def stepsize(self) -> np.ndarray:
         Stepsize used by sampler for each chain.
         When sampler algorithm 'fixed_param' is specified, stepsize is None.
         """
-        if not self._is_fixed_param and self._stepsize is None:
+        if self._is_fixed_param:
+            return None
+        if not self._validate_csv and self._stepsize is None:
+            self._logger.warning(
+                'csv files not yet validated, run method validate_csv_files()'
+                ' in order to retrieve sample metatdata.'
```
ahartikainen, line 424: metadata

ahartikainen:
Wonder what is speed difference between arviz csv read vs our?

`from_cmdstan` does more than just reading, but there is a function to read csv --> dataframe + metadata

cmdstanpy/stanfit.py
mitzimorris: excellent catch - many thanks.

mitzimorris:
> Wonder what is speed difference between arviz csv read vs our?

very good question.  I need to do more timing testing.
the Stan CSV file format is a nightmare, but it's something that's going to need to be supported for a long time, so figuring out the most efficient way to handle these files would be great.  I just did a pass at documenting the format(s) in the new CmdStan User's Guide - https://mc-stan.org/docs/2_24/cmdstan-guide/stan-csv.html

mitzimorris:
fixing it old skool - *blush*

> perl -p -i -e 's/metatdata/metadata/g' stanfit.py

mitzimorris:
my original timing test was a hack of the bernoulli model so that it did more I/O but very little computation.
if I run this on the simple bernoulli model - then the difference in scanning/not scanning the CSV files is much greater.

if you want to use CmdStanPy just to do the fitting and plan on using Arviz or similar to read in the CSV files, then this will make the CmdStanPy run faster - and how much faster depends on how much work the sampler needs to do per iteration compared to the I/O for writing that iteration out to disk.

PR #280, opened 2020-08-14
Feature/22 better error handling

mitzimorris:
#### Submission Checklist

- [x] Run unit tests

#### Summary

When CmdStan method returns a non-zero error code, return as much useful information as can be scraped out of the stdout and stderr messages, plus information about the run and locations of resulting output files, if any.

Exposed and expanded `RunSet` method `get_err_msgs`; this function was written a long time ago, but wasn't being called by the CmdStanModel methods.  It attempts to get just the relevant messages from the information spew that's written to stdout, in addtion to all msgs written to stderr.  It seems to do an OK job on the most common and annoying problem - return code 70 which happens when there's a mismatch between the model's data block variable declarations and the input data.

#### Copyright and Licensing

### DIFF
--- a/cmdstanpy/model.py
+++ b/cmdstanpy/model.py
@@ -750,17 +750,27 @@ class CmdStanModel:
                 # re-enable logger for console
                 self._logger.propagate = True
 
-            err_msg = 'Error during sampling.\n'
             if not runset._check_retcodes():
-                for i in range(chains):
-                    if runset._retcode(i) != 0:
-                        err_msg = '{}chain {} returned error code {}\n'.format(
-                            err_msg, i + 1, runset._retcode(i)
+                bad_runs = [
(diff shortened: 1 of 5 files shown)


cmdstanpy/model.py
```diff
@@ -393,7 +387,7 @@ def optimize(
                 seed=seed,
                 inits=_inits,
                 output_dir=output_dir,
-                save_diagnostics=save_diagnostics,
+                save_diagnostics=False,
```
ahartikainen, line 390: This is not needed anymore?

mitzimorris: right - at the stan::services layer, https://github.com/stan-dev/stan/tree/develop/src/stan/services/optimize - if you look at these methods, they don't take a diagnostics file as an argument.

PR #283, opened 2020-08-26
Some fixes/additions to the variational arguments

b-r-oleary:
#### Submission Checklist

- [x] Run unit tests

#### Summary

Hi! I was exploring the possibility of migrating a project from pystan to cmdstanpy in order to access a more up-to-date version of stanc via a python interface. During that exploration, I found that there was some unexpected behavior when working with the variational interface. I poked around in the codebase and found a few problems that I attempted to fix in this PR. In particular:
- Some of the `Real` parameters in `VariationalArgs` were being constrained to be `>=1` rather than `>0`, which resulted in unexpected validation errors when I supplied values less than 1, e.g. `eta = 0.1`, or `tol_rel_obj = 0.05`.
- The interface did not support disabling eta adaptation, which is supported by cmdstan, and is supported by other stan interfaces.
- When disabling eta adaptation, the output file parser was expecting output data provided by eta adaptation that isn't present, and was also expecting a particular value of eta=1.
- cmdstanpy throws an error if stan reports that "The algorithm may not have converged.". Since pystan doesn't throw an error under these conditions, and since cmdstan returns a 0 exit status under this condition, this behavior was unexpected to me, especially given that I am sometimes interested in the output of variational inference even if the algorithm has not converged. So, I added an argument `require_converged` that would allow us to disable that check, without changing the default behavior of the function.

I ran the unit-tests locally in a container built from the Dockerfile in this repo and some of the tests failed, but I don't think that those test failures are related to any of these changes.

#### Copyright and Licensing

Brendon O'Leary

### DIFF
--- a/cmdstanpy/cmdstan_args.py
+++ b/cmdstanpy/cmdstan_args.py
@@ -403,6 +403,7 @@ class VariationalArgs:
         elbo_samples: int = None,
         eta: Real = None,
         adapt_iter: int = None,
+        adapt_engaged: bool = True,
         tol_rel_obj: Real = None,
         eval_elbo: int = None,
         output_samples: int = None,
@@ -413,6 +414,7 @@ class VariationalArgs:
         self.elbo_samples = elbo_samples
         self.eta = eta
         self.adapt_iter = adapt_iter
+        self.adapt_engaged = adapt_engaged
         self.tol_rel_obj = tol_rel_obj
         self.eval_elbo = eval_elbo
         self.output_samples = output_samples
@@ -503,9 +505,13 @@ class VariationalArgs:
             cmd.append('elbo_samples={}'.format(self.elbo_samples))
         if self.eta is not None:
             cmd.append('eta={}'.format(self.eta))
-        if self.adapt_iter is not None:
-            cmd.append('adapt')
-            cmd.append('iter={}'.format(self.adapt_iter))
+        cmd.append('adapt')
+        if self.adapt_engaged:
+            cmd.append('engaged=1')
+            if self.adapt_iter is not None:
+                cmd.append('iter={}'.format(self.adapt_iter))
+        else:
+            cmd.append('engaged=0')
         if self.tol_rel_obj is not None:
             cmd.append('tol_rel_obj={}'.format(self.tol_rel_obj))
         if self.eval_elbo is not None:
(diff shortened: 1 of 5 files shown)


cmdstanpy/cmdstan_args.py
```diff
@@ -336,7 +336,7 @@ def validate(self, chains=None) -> None:  # pylint: disable=unused-argument
                     'init_alpha must not be set when algorithm is Newton'
                 )
             if isinstance(self.init_alpha, Real):
-                if self.init_alpha < 0:
+                if self.init_alpha <= 0:
```
mitzimorris, line 339: must be greater than 0 - https://github.com/stan-dev/cmdstan/blob/3e93d44ca33200bb3f09218451ffa36ab3c8a01d/src/cmdstan/arguments/arg_init_alpha.hpp#L13

cmdstanpy/cmdstan_args.py
```diff
@@ -132,7 +132,7 @@ def validate(self, chains: int) -> None:
                 )
         if self.step_size is not None:
             if isinstance(self.step_size, Real):
-                if self.step_size < 0:
+                if self.step_size <= 0:
```
mitzimorris, line 135: must be strictly greater than zero - https://github.com/stan-dev/cmdstan/blob/3e93d44ca33200bb3f09218451ffa36ab3c8a01d/src/cmdstan/arguments/arg_stepsize.hpp#L13

ahartikainen: I think it was wrong

If `init_alpha` was 0, it would have passed, but now it is catched.

See

    0 < init_alpha

vs 

    init_alpha <= 0

mitzimorris: doh!  you are both correct - got the comparison backward.

cmdstanpy/cmdstan_args.py
mitzimorris: my mistake - this change it correct - got things backwards.
would `if not self.init_alpha > 0` be easier to understand?

mitzimorris:
thanks @ahartikainen for pointing out my mistake - the code is correct, it just needs to pass the linter and unit tests.

mitzimorris:
many thanks!

PR #287, opened 2020-09-04
Stan variable

LukasNeugebauer:
#### Submission Checklist

- [x] Run unit tests

#### Summary

CmdStanMCMC.stan_variable now returns pandas.DataFrames with named columns instead of numpy.ndarrays. I also adapted the test that checks the output shape. This is because we're now always returning 2D DataFrames, while the dimensionality was variable before.

#### Copyright and Licensing

Lukas Neugebauer


cmdstanpy/stanfit.py
```diff
@@ -754,20 +754,23 @@ def stan_variable(self, name: str) -> np.ndarray:
...
-            var_dims.extend(dims)
-            return self._draws[
-                self._draws_warmup :, :, idxs[0] : idxs[-1] + 1
-            ].reshape(tuple(var_dims), order='A')
+            return pd.DataFrame({
+                n: self._draws[
+                    self._draws_warmup:, :, x
+                ].reshape(dim0, order='A')
+                for x, n in idxs
+            })
```
ahartikainen, line 773: Hi, I think there is now a possibility to streamline this a bit, at the same time we can use regex to find columns.

Please check if this code works.

Import re at start
```
import re
```

This works for all the variable types. Also it takes one chunk on transforms that to dataframe, so no need for dictionary handling.


```
dims = self._stan_variable_dims[name]
idxs = []
names = []
pattern = r'^{}(\.\d)*$'.format(name)
for i, column_name in enumerate(self.column_names):
    if re.search(pattern, column_name):
        names.append(column_name)
        idxs.append(i)

var_dims = [dim0]
var_dims.extend(dims)
return pd.DataFrame(self._draws[
    self._draws_warmup :, :, idxs[0] : idxs[-1] + 1
].reshape(tuple(var_dims), order='A'), columns=names)
```

LukasNeugebauer: Hey! 
Thanks a lot! Of course you're right. I made a few tweaks to your suggestion to make it work. Hope that's fine.

cmdstanpy/stanfit.py
```diff
@@ -751,26 +751,19 @@ def stan_variable(self, name: str) -> pd.DataFrame:
...
-                for x in enumerate(self.column_names)
-                if x[1].startswith(name + '.')
-            ]
-            return pd.DataFrame({
-                n: self._draws[
-                    self._draws_warmup:, :, x
-                ].reshape(dim0, order='A')
-                for x, n in idxs
-            })
+        dims = np.prod(self._stan_variable_dims[name])
```
ahartikainen, line 754: Good use of prod here

cmdstanpy/stanfit.py
```diff
@@ -727,47 +727,43 @@ def draws_as_dataframe(
...
+                idxs.append(i)
+        return pd.DataFrame(
+            self._draws[
+                self._draws_warmup:, :, idxs
+            ].reshape((dim0, dims), order='A'),
+            columns=names
+        )
 
     def stan_variables(self) -> Dict:
         """
```
ahartikainen, line 769: What should we do for this function? Or maybe we could basically copy the `stan_variable`, but just change it so it works with multiple names. I don't mind of duplicated code, but I would like to create dataframe only once.

Would it make sense to use concat? And suggest users to learn filter

https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.filter.html

cc @mitzimorris

LukasNeugebauer: It could return the whole DataFrame as default and a DataFrame containing only the requested variables if names are given. But I guess that would make the stan_variable function kind of redundant.

mitzimorris:
> a pandas dataframe would allow us to keep the column names from the Stan csv file.

could you add unit tests showing how this works?
the Lotka-Volterra model is solving the ODE to get populations for species _u_ and _v_ at time _t_ - this is coded as transformed data variable `real z[N, 2]`  - each column is a species - either preditor or prey - each row is a year.  column 1 names are `z.1.1` through `z.20.1`  and column 2  names are `z.1.2` through `z.20.2`.
it would be nice to show how using column names helps - the R code in the case study 
does this make sense?

LukasNeugebauer:
> > a pandas dataframe would allow us to keep the column names from the Stan csv file.
> 
> could you add unit tests showing how this works?

You mean tests that check that the expected column names show up in the DataFrame?

ahartikainen:
Do we want to keep `x.4.7` or go with `x[4,7]`

LukasNeugebauer:
I think [4,7] is nicer than .4.7.

cmdstanpy/stanfit.py
ahartikainen: True, we could make it so that stab_variable would call stan_variables.

What is the output in CmdStanR?

LukasNeugebauer: I don't know because I don't use R but I'll have a look at it later. Does stan_variables have to be a function? How about a cached property and stan_variable just uses filter on the columns?

LukasNeugebauer:
Is the renaming of columns from .4.7 to [4,7] something that comes up frequently? Would that be worth a function on its own?

cmdstanpy/stanfit.py
LukasNeugebauer: Just had a look at the CmdStanR and the equivalent seems to be CmdStanMCMC::draws() which returns a iterations X chains X variables array or a iterations X chains array when called with a variable name as argument.

mitzimorris: maybe this obviates the need for `stan_variables`?

mitzimorris: what's useful about `stan_variables` and corresponding `stan_variable_dims` is that the returned dict gives you the names of all Stan variables present in the output - we need something that combines this and `stan_variable_dims`, information-wise.

ahartikainen: So maybe we could still return a dict of dataframes in `stan_variables`, that sounds fine by me.

ahartikainen:
@mitzimorris what your opinion on renaming parameters? If CmdStanR doesn't do it, then we probably shouldn't so it either.

mitzimorris:
> Is the renaming of columns from .4.7 to [4,7] something that comes up frequently? 

renaming to `[4,7]` should be done, as this is what has been done, not just in the interfaces, but also in CmdStan's `stansummary` function.  the name `foo.4.7` should be thought of as the internal name - that's just the header that's produced by the Stan model code.

LukasNeugebauer:
> > Is the renaming of columns from .4.7 to [4,7] something that comes up frequently?
> 
> renaming to `[4,7]` should be done, as this is what has been done, not just in the interfaces, but also in CmdStan's `stansummary` function. the name `foo.4.7` should be thought of as the internal name - that's just the header that's produced by the Stan model code.

> Is the renaming of columns from .4.7 to [4,7] something that comes up frequently?

This should work:

```
def rename_columns(column_names):
    return [re.sub(r",([\d,]+)$", r"[\1]", column.replace(".",",")) for column in column_names]
```

Any suggestions? If not, should I put it in `utils.py`? Or just rename the columns in `stan_variable`?

ahartikainen:
Hmm, Should we rename things already at the csv read step?

mitzimorris:
> Hmm, Should we rename things already at the csv read step?

yes!  reasons:
 - it lines up with how variables are used in Stan program
 - that's what CmdStanR does
 - that's how parameters are reported by `stansummary` cmds

LukasNeugebauer:
Sorry for the late reply, I was quite busy. I'll have a look at where we would have to rename the columns and try to implement it.

ahartikainen (approved):
LGTM

mitzimorris (approved):
LGTM!

ahartikainen:
@bletham FYI I think this is a breaking change for prophet.

mitzimorris:
you do need to get rid of pylint and flake8 complaints

LukasNeugebauer:
> you do need to get rid of pylint and flake8 complaints

That's the stuff in the Travis CI build, right? I had no idea what these tests are doing and was wondering why one failed anyway.
Ok, so to do for tomorrow: Find out what pylint and flake8 are and fix whatever's wrong!

mitzimorris:
many thanks!

LukasNeugebauer:
My pleasure. Hopefully not the last one. Thanks for guiding me through this! :)

PR #303, opened 2020-10-05
Feature/227 cmdstan install

mitzimorris:
#### Submission Checklist

- [x] Run unit tests

#### Summary

Added logic to `cmdstanpy/utils.py` function `install_cmdstan` to update environment var `CMDSTAN` so that user doesn't have to do additional call to `set_cmdstan_path`; instead newly installed version is used automatically.

Updated docs accordingly, added section calling out use of `install_cmdstan` to keep up with latest CmdStan release.

#### Copyright and Licensing


cmdstanpy/utils.py
```diff
@@ -902,6 +912,29 @@ def install_cmdstan(
...
+    else:
+        version = '-'.join(['cmdstan', version])
+    new_cmdstan_path = os.path.join(cmdstan_dir, version)
+
+    # validate, set path directly - avoid circular error messages
+    stanc_bin = os.path.join(new_cmdstan_path, 'bin', 'stanc' + EXTENSION)
+    if not os.path.exists(stanc_bin):
+        logger.warning('CmdStan installation failed - missing stanc compiler')
+        return False
+    os.environ['CMDSTAN'] = new_cmdstan_path
```
ahartikainen, line 937: So, this only works if called from python session.

Next time you open a python, the env var does not exist.

mitzimorris: right - if you install something somewhere, you'll have to write down where you put it.

mitzimorris: this fix is to address the reasonable user expectation that if you call a function to install a cmdstan version, then going forward - _in the session_  - that becomes the version of cmdstan that is used.   if you call function `install_cmdstan` without specifying the base dir ($HOME/.cmdstan) or version, you'll get this for free.
if, however, you specify some other base dir, or some version other than the latest, you don't.

so this fix takes into account non-default dir and version, and does the right thing.  trying to keep track of what is done in one session for future sessions is a workflow issue, and that's somebody else's problem.  does this make sense?

ahartikainen (approved):

mitzimorris:
@ahartikainen - test of install_cmdstan is hanging on windows - this might be a real problem, if so, it's in latest version on PyPi - 9.67.  do you have a windows machine you could test 9.67 on?

test/test_install_cmdstan.py
```diff
@@ -31,6 +38,30 @@ def test_retrieve_version(self):
...
+    def test_install_cmdstan_specify_dir(self):
+        cur_cmdstan_path = ''
+        if 'CMDSTAN' in os.environ:
+            cur_cmdstan_path = os.environ['CMDSTAN']
+            del os.environ['CMDSTAN']
+        self.assertFalse('CMDSTAN' in os.environ)
+
+        version = '2.24.1'
+        tmpdir = os.path.join(DATAFILES_PATH, 'tmp-' + str(time()))
+        os.mkdir(tmpdir)
```
ahartikainen, line 50: `os.makedirs(tmpdir, exist_ok=True)`

ahartikainen (commented):
I will test

test/test_install_cmdstan.py
mitzimorris: many thanks!!!

mitzimorris: nope, test still hangs on Windows.  without this, tests run in 12 minutes.  this test tries to do a download from GitHub - maybe not allowed in the CI setting?  in any case, I think it'll be OK to skip running this tests on Windows.

ahartikainen:
Yes, let's skip this.

I tried, but I don't know, had even problems using install_cxx_toolchain.

Btw, why is there 2 functions for install_cmdstan? And why the other one follows the output on Windows, but the other one doesn't?

mitzimorris:
> Btw, why is there 2 functions for install_cmdstan? And why the other one follows the output on Windows, but the other one doesn't?

what do you mean, 2 functions?

I'm not sure how the code has evolved - there's function in file `utils.py` called `install_cmdstan` that calls main of script in file `install_cmdstan.py` - can things be done differently?

mitzimorris:
>  the other one follows the output on Windows, but the other one doesn't?

which follows output on Windows?  package function in `utils.py` or script?

ahartikainen:
Ok, I see what you mean.

we can basically do

    from .install_cmdstan import main

    main()

But we would then want to fill sys.argv.

What we could do, is to move the sys.argv handling to `if __name__ == "__main__"` block, or manually fill the sys.argv before calling the main.

Other option is to define function arguments and set them to none, if they are none, then handle sys.argv.

I would do the __name__ thing (move sys.argv handling out of the main function)

mitzimorris:
there seem to be problems with GitHub?  getting intermittant failures.

ahartikainen:
I think, there might be some problem with subprocess to subprocess error handling. And anyways, the compilation of CmdStan on Windows take 10-20 minutes (model compilation takes 60+ seconds).

cmdstanpy/utils.py
```diff
@@ -887,21 +897,45 @@ def install_cmdstan(
...
+        cmd.extend(['--overwrite'])
     if verbose:
-        cmd.extend(['--verbose', 'TRUE'])
+        cmd.extend(['--verbose'])
     proc = subprocess.Popen(
         cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, env=os.environ
     )
-    while proc.poll() is None:
-        print(proc.stdout.readline().decode('utf-8').strip())
+    if platform.system() != 'Windows':
```
ahartikainen, line 906: Why this is blocked on Windows?

mitzimorris:
> the compilation of CmdStan on Windows take 10-20 minutes (model compilation takes 60+ seconds).

wow.  

we need CI tests on Windows, because we've got lots of users on Windows, but not that many devs.

cmdstanpy/utils.py
mitzimorris: will remove that - was trying to see if that would fix Windows build problem - SO suggested that reading from stdout would cause pipes to block and processes to hang.  since unit tests continued to fail on Windows, I don't think this was the culprit.  however, if we can get rid of chain of subprocesses, that would be best.

mitzimorris:
hi @ahartikainen  - per your suggestion:

> move the sys.argv handling to if __name__ == "__main__" block

then rename  `main` to `install_cmdstan`  and expose it directly as part of package?

mitzimorris:
Oy vey!!!  (https://en.wikipedia.org/wiki/Oy_vey)

need to refactor script `install_cxx_toolchain.py` as well, as it's getting called from `install_cmdstan`.

mitzimorris:
@ahartikainen - as part of this PR, going to refactor install_cxx_toolchain - question:

always install toolchain into dir $HOME/_DOT_CMDSTAN(py)?
the toolchain location is added to the path, therefore it should always be in a known location - if the user wants a CmdStan install in a special dir, that's one thing,  cxx toolchain is different.

does this make sense?

ahartikainen:
So there are couple of things to consider.

### First install location:

The default loc should be the `.cmdstan`. There is a case to check `C:\RTools40` because that is the default folder used by rtools installer.

What do you think, if we first check `.cmdstan(py)` folders and then if compiler is not found, check C:\Rtools40?

### Use of rtools compiler

By default, we should assume user has compiler on their path.  And after 2.25/2.26 this might be true (when conda toolchain start to work again). Then calling `cmdstanpy.utils.cxx_toolchain_path(compiler_root_path)` is called then we should add correct folders to PATH (compiler_root_path) or we should check the default locations and add the first compiler found in the default folders.

But, can we also first check if gcc is found and if it is not, then call internally `cmdstanpy.utils.cxx_toolchain_path()`

So what does this mean.

So when users first install cmdstanpy

### Case 1: Default folders

    python -m cmdstanpy.install_cmdstan

This should first check if gcc is found. If not, install Rtools to default location, then install CmdStan.

Then everything should work as expected

### Case 2: Default folders, conda gcc for python stuff

Users needs to install rtools manually

    python -m cmdstanpy.install_cxx_toolchain
    python -m cmdstanpy.install_cmdstan --compiler

And when starting to code in python

    from cmdstanpy.utils import cxx_toolchain_path
    cxx_toolchain_path()

### Case 3: Non -default location

User install rtools manually

    python -m cmdstanpy.install_cxx_toolchain -dir my/path

# User needs to add rtools folders to path

https://stackoverflow.com/questions/19287379/how-do-i-add-to-the-windows-path-variable-using-setx-having-weird-problems

    setx my/path/rtools40/...

(this is not really easy, can we add possibility to add these if needed)

     python -m cmdstanpy.install_cxx_toolchain -dir my/path --add_global

which would call subprocess

    subprocess.call(["setx", correct paths])

Then, how do we tell install_cmdstan to use non-default path, if it is not add to PATH (?)

    python -m cmdstanpy.install_cmdstan --compiler_path my/path

After CmdStan is installed, either cxx_toolchain_path needs to be called

    from cmdstanpy.utils import cxx_toolchain_path
    cxx_toolchain_path()

or nothing special is needed, if global PATH is changed.

mitzimorris:
hi @ahartikainen - returning to this issue now - question w/r/t conda - is this necessary for CmdStanPy?
what about case that's just for C++ toolchain and CmdStan?

ahartikainen:
Hi, I think there are no issues anymore with latest CmdStan.

So basically conda users with conda compilers don't need to install RTools. And normal users without any compilers can install RTools and then activate them as the example shows.

mitzimorris:
closing - this is outdated - will revisit this issues in another PR.

PR #312, opened 2020-10-14
Enable per chain options

ahartikainen:
#### Submission Checklist

- [ ] Run unit tests

#### Summary

This PR fixes the bug which caused per chain options to fail.

Now user can set these values individually per chain: `stepsize`, `inits`, `metric_file`, `seed`.

#### Copyright and Licensing

Ari Hartikainen

### DIFF
--- a/cmdstanpy/stanfit.py
+++ b/cmdstanpy/stanfit.py
@@ -527,7 +527,15 @@ class CmdStanMCMC:
                 )
                 for key in dzero:
                     if (
-                        key not in ['id', 'diagnostic_file']
+                        key
+                        not in [
+                            'id',
+                            'diagnostic_file',
+                            'metric_file',
+                            'stepsize',
+                            'init',
+                            'seed',
+                        ]
                         and dzero[key] != drest[key]
                     ):
                         raise ValueError(
@@ -715,9 +723,7 @@ class CmdStanMCMC:
                 len(self.column_names),
                 order='A',
             )
-            self._draws_pd = pd.DataFrame(
-                data=data, columns=self.column_names
-            )
+            self._draws_pd = pd.DataFrame(data=data, columns=self.column_names)
         if params is None:
             return self._draws_pd
         mask = []
@@ -759,10 +765,10 @@ class CmdStanMCMC:
                 names.append(column_name)
                 idxs.append(i)
         return pd.DataFrame(
-            self._draws[
-                self._draws_warmup:, :, idxs
-            ].reshape((dim0, dims), order='A'),
-            columns=names
+            self._draws[self._draws_warmup :, :, idxs].reshape(
+                (dim0, dims), order='A'
+            ),
+            columns=names,
         )
 
     def stan_variables(self) -> Dict:
--- a/cmdstanpy/utils.py
+++ b/cmdstanpy/utils.py
@@ -941,9 +941,10 @@ class MaybeDictToFilePath:
                 for i, obj_item in enumerate(obj):
                     if not isinstance(obj_item, str):
                         err_msgs.append(
-                            'List element {} must be a filename string, found {}'.format(
-                                i, obj_item
-                            )
+                            (
+                                'List element {} must be a filename string,'
+                                ' found {}'
+                            ).format(i, obj_item)
                         )
                     elif not os.path.exists(obj_item):
                         missing_obj_items.append(
--- a/cmdstanpy/utils.py
+++ b/cmdstanpy/utils.py
@@ -908,7 +908,9 @@ def install_cmdstan(
 class MaybeDictToFilePath:
     """Context manager for json files."""
 
-    def __init__(self, *objs: Union[str, dict], logger: logging.Logger = None):
+    def __init__(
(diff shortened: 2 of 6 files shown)


cmdstanpy/stanfit.py
```diff
@@ -527,7 +527,15 @@ def validate_csv_files(self) -> None:
...
                     if (
-                        key not in ['id', 'diagnostic_file']
+                        key
+                        not in [
+                            'id',
+                            'diagnostic_file',
+                            'metric_file',
+                            'stepsize',
+                            'init',
+                            'seed',
```
ahartikainen, line 537: is that seed correct key?

cmdstanpy/utils.py
```diff
@@ -933,6 +935,20 @@ def __init__(self, *objs: Union[str, dict], logger: logging.Logger = None):
                 if not os.path.exists(obj):
                     raise ValueError("File doesn't exist {}".format(obj))
                 self._paths[i] = obj
+            elif isinstance(obj, list):
+                if not all(isinstance(obj_item, str) for obj_item in obj):
+                    raise ValueError("Not all list items are str")
```
ahartikainen, line 940: Maybe a bit better error msg?

cmdstanpy/utils.py
```diff
@@ -933,6 +935,20 @@ def __init__(self, *objs: Union[str, dict], logger: logging.Logger = None):
...
+                missing_obj_items = []
+                for obj_item in obj:
+                    if not os.path.exists(obj_item):
+                        missing_obj_items.append(obj_item)
+                if missing_obj_items:
+                    raise ValueError(
+                        "File doesn't exist: {}".format(
```
ahartikainen, line 947: Better error msg?

cmdstanpy/utils.py
mitzimorris: yes.  error messages should correspond to how user views these objects.  how about:

`'List element {} must be a filename string, found {}'.format(i, obj_item)`

cmdstanpy/utils.py
mitzimorris: if the calling code provides additional context, then this should be OK - there are a couple of kinds of args that can be either a dict or a json file, or list thereof.

mitzimorris:
hi @ahartikainen let me know when this is ready for review.
we need unit tests to test that the bug is fixed; also useful way to figure out what the sensible use cases are.
and if you can enumerate various use cases, I'd be happy to put more examples into the docs and notebooks.

ahartikainen:
Hi, this is ready now.

One use case, is to "continue" sampling.

```python
import cmdstanpy
import numpy as np
import ujson as json
import arviz as az

model_code = """
data {
    int N;
    real y[N];
}
parameters {
    real alpha;
    real<lower=0> sigma;
}
model {
    alpha ~ normal(0,1);
    sigma ~ normal(0,1);
    y ~ normal(alpha, sigma);
}
"""

model_path = "./normal_model.stan"
with open(model_path, "w") as f:
    print(model_code, file=f)

model = cmdstanpy.CmdStanModel(stan_file=model_path)
stan_data = {
    "N": 10,
    "y": np.random.randn(10) * 3 + 2
}

seed = 100
fit = model.sample(data=stan_data, chains=4, seed=seed, iter_warmup=1000, iter_sampling=100)

idata = az.from_cmdstanpy(posterior=fit)

def get_inits(fit, checkpoint_name="init_checkpoint"):
    params_bool = [not item.endswith("__") for item in fit.column_names]
    params = [item for item in fit.column_names if not item.endswith("__")]
    names = []
    for i, init in enumerate([item[params_bool].tolist() for item in fit.draws(inc_warmup=True)[-1]], 1):
        names.append(checkpoint_name + f"_chain_{i}" + ".json")
        with open(names[-1], "w") as f:
            json.dump(dict(zip(params, init)), f)
    return names

inits = get_inits(fit, checkpoint_name="init_checkpoint_round_0")

def get_stepsize(fit):
    return fit.stepsize.tolist()

stepsize = get_stepsize(fit)

def get_metric(fit, checkpoint_name="metric_checkpoint"):
    names = []
    for i, metric in enumerate([item.tolist() for item in fit.metric], 1):
        names.append(checkpoint_name + f"_chain_{i}" + ".json")
        with open(names[-1], "w") as f:
            json.dump({"inv_metric": metric}, f)
    return names

metric = get_metric(fit, checkpoint_name="metric_checkpoint_round_0")

seed += 4

new_fit = model.sample(
    data=stan_data,
    chains=4, 
    seed=seed, 
    inits=inits,
    step_size=stepsize,
    metric=metric,
    iter_warmup=0, 
    adapt_engaged=False, 
    iter_sampling=100
)

new_idata = az.from_cmdstanpy(posterior=new_fit)
idata = az.concat(idata, new_idata, dim="draw")
```

The seed can be handled with seed value + id, but then again user can use specific seed for each chain. Not saying when this is a good idea (e.g. some stuck chains, user filter outs some chains --> not a good idea)

funko-unko:
> One use case, is to "continue" sampling.

Hey, that's exactly my use case!

PR #321, opened 2020-11-23
fix test_utils tests

ahartikainen:
#### Submission Checklist

- [ ] Run unit tests
- [ ] Declare copyright holder and open-source license: see below

#### Summary

I don't know a good way to test multiple cmdstan versions without installing multiple versions.
Also changed a bunch of tests to use tempfolder.

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting (this will be you or your assignee, such as a university or company):

Ari Hartikainen

By submitting this pull request, the copyright holder is agreeing to license the submitted work under the following licenses:
- Code: BSD 3-clause (https://opensource.org/licenses/BSD-3-Clause)

PR #324, opened 2020-11-24
Additional initialisation options for `CmdStanModel.sample`

### LINKED ISSUE #49: function `sample` argument `inits` should allow initialization function
#### Summary:
The `sample` command `inits` is used to specify some or all initial parameter values.

Add functionality as in [RStan](https://mc-stan.org/rstan/reference/stan.html):
> Set initial values by providing a function that returns a ~list~ Dict for specifying the initial values of parameters for a chain. The function can take an optional parameter chain_id through which the chain_id (if specified) or the integers from 1 to chains will be supplied to the function for generating initial values.

tcbegley:
#### Submission Checklist

- [x] Run unit tests

#### Summary

Hi again, thanks for the great package, I'm enjoying playing around with it a lot.

This PR allows more flexible initialisation of parameters when using `CmdStanModel.sample`. In particular it allows initialisation via a function which optionally takes `chain_id` as an argument. I have tried to be faithful to the logic and error reporting present in [the relevant parts of cmdstanr](https://github.com/stan-dev/cmdstanr/blob/c84fab987d48b0ccc2ba635b631b75c5e12b1ea4/R/args.R#L703-L720)

The approach I took was to call the initialisation function if passed inside `sample` to create a list of dictionaries, then allow `MaybeDictToFilePath` to take care of the rest, mainly as `MaybeDictToFilePath` doesn't have visibility on `chain_ids`. This required also making sure that `MaybeDictToFilePath` can handle lists of dictionaries as well as single dictionaries. I have added a test for this new behaviour.

Very happy to take guidance on whether the approach here is right, whether I've added enough input validation and whether more test coverage is needed.

Fixes #49 

#### Copyright and Licensing

I am the copyright holder of this work and agree to the terms of the license.

### DIFF
--- a/cmdstanpy/model.py
+++ b/cmdstanpy/model.py
@@ -12,7 +12,8 @@ from concurrent.futures import ThreadPoolExecutor
 from multiprocessing import cpu_count
 from numbers import Real
 from pathlib import Path
-from typing import Any, Dict, List, Union
+from typing import Any, Callable, Dict, List, Union
+from inspect import signature
 
 from cmdstanpy.cmdstan_args import (
     CmdStanArgs,
@@ -38,6 +39,17 @@ from cmdstanpy.utils import (
     get_logger,
     scan_sampler_csv,
 )
+
+# compound type for sampler initialiser
+SamplerInitsType = Union[
+    float,
+    str,
+    Dict,
+    List[str],
+    List[Dict],
+    Callable[[], Dict],
+    Callable[[int], Dict],
+]
 
 
 class CmdStanModel:
@@ -428,7 +440,7 @@ class CmdStanModel:
         threads_per_chain: Union[int, None] = None,
         seed: Union[int, List[int]] = None,
         chain_ids: Union[int, List[int]] = None,
-        inits: Union[Dict, float, str, List[str]] = None,
+        inits: SamplerInitsType = None,
         iter_warmup: int = None,
         iter_sampling: int = None,
         save_warmup: bool = False,
@@ -512,6 +524,12 @@ class CmdStanModel:
             * dictionary - pairs parameter name : initial value.
             * string - pathname to a JSON or Rdump data file.
             * list of strings - per-chain pathname to data file.
+            * list of dictionaries - per-chain initialisation dictionaries
+            * function - a function returning a dictionary of initial parameter
+              values. The function can take an optional parameter chain_id
+              through which the chain_id (if specified) or the integers from 1
+              to chains will be supplied to the function for generating initial
+              values.
 
         :param iter_warmup: Number of warmup iterations for each chain.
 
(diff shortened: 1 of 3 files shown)


ahartikainen:
Thanks for the PR.

Is this a structure we want to support?

I know pystan has this, but what is the use-case? Does it make easy to get an idea what inits is used?

tcbegley:
Very fair questions. Here's a couple thoughts.

Use cases
- Restarting sampling from where a previous run left off. I have a use case which is roughly speaking one where I need to do cross-validation on a slow to sample model, and approximations like PSIS-LOO don't help. I've found that good initialisation from previous runs can really help shorten the needed warmup without introducing unacceptable amounts of bias. imo a list of dictionarys for per-chain initial values would be a very convenient way to achieve this.
- Initialising with a sample from the prior rather than uniform samples on the unconstrained scale. This is something I've found helpful before when the uniform initialisation on the unconstrained scale had a tendency to give initial values with extremely low probability and caused some of my chains to get stuck. A solution might have been to transform those parameters so that `unif(-2, 2)` initialisation was less likely to give problems, but keeping the model intact and passing in a custom initialisation function that could draw a sample from my priors was a convenient pattern for me in PyStan.

Arguments for these changes
- per-chain initialisation via list of dictionaries or a function offers a clean interface for achieving use-cases such as the above
- since cmdStanPy already supports per-chain initialisation via list of strings, there isn't really anything new here that wasn't possible before (I could always write my initialisation function, call it a few times and serialise the results to JSON), but it does reduce boilerplate code for the user
- feature parity with cmdStanR + familiarity for former PyStan users.

> Does it make easy to get an idea what inits is used?

Not necessarily as the case of initialisation by a list of dictionaries or a function inits are serialised to a temporary JSON file and then deleted once the context of the `MaybeDictToFilePath` is left. However

1. This is consistent with initialisation via a single dictionary as is currently supported
2. If knowing the initial values is vitally important, there is always the fallback of serialising the inits yourself and initialising via a list of strings
3. We also, regardless of how the sampler was initialised, have the option of inspecting the output of `.draws` which will show the first sample, which (unless I'm wrong about `.draws` starting at the first sample rather than the initial value) is not quite the initial values, but should be pretty close and maybe good enough for diagnose serious issues.

mitzimorris:
this has been on the TODO list for a long time - https://github.com/stan-dev/cmdstanpy/issues/49

mitzimorris:
@serban-nicusor-toptal hi Nic,  noticed that this build failed on Travis because tests are looking for CmdStan 2.24.1 but latest CmdStan is 2.25.0.  I can't figure out why this is happening - is something cached somewhere or otherwise configured?

serban-nicusor-toptal:
Hey @mitzimorris [here](https://github.com/stan-dev/cmdstanpy/blob/develop/.travis.yml) you can see the travis config, it doesn't reference any version for cmdstan. Maybe it's hardcoded somewhere in cmdstanpy ?

ahartikainen:
See https://github.com/stan-dev/cmdstanpy/pull/321

We never install 2.24 and cache has cleared.

mitzimorris:
@serban-nicusor-toptal - thanks for checking!

@ahartikainen I see problems with some unit tests - e.g. here: https://github.com/stan-dev/cmdstanpy/blob/fa66eeb012c978adf11f6ba8be2cad2542827dd7/test/test_utils.py#L181-L182

pretty sure this one's on me

mitzimorris:
this PR has gone stale - revive or close?

ahartikainen:
Let's close this PR.

If there is still need for this kind of functionality we can talk about it in Discourse.

I really don't think this approach (function that takes chain_id) solves any real problem (more like a legacy thing).

tcbegley:
> I really don't think this approach (function that takes chain_id) solves any real problem (more like a legacy thing).

I think in which case it's worth closing #49. The reason I originally opened this PR was because I saw an open issue from one of the maintainers of this library which suggested it was a wanted feature.

ahartikainen:
> > I really don't think this approach (function that takes chain_id) solves any real problem (more like a legacy thing).
> 
> I think in which case it's worth closing #49. The reason I originally opened this PR was because I saw an open issue from one of the maintainers of this library which suggested it was a wanted feature.

You are right! I think we have gone forward about what kind of API we want to have.
Also just to be explicit there was nothing wrong with the code written.

tcbegley:
> Also just to be explicit there was nothing wrong with the code written.

Kind of you to say so explicitly, but really not necessary, no hard feelings here 🙂

PR #334, opened 2020-12-18
stdout/stderr on popen  handling  (for windows)

akirayou:
#### Submission Checklist

- [x] Run unit tests (On Ubuntu)

#### Summary
Fix for windows (I tested it on python 3.8.6 /windows10 Japanese locale )
・Pseudo non-block IO for subprocess.Popen  stdout/stderr
・Use system.stdin.encoding insisted of 'utf-8' for bytes.decode('utf-8')

Some stan model have many warning message (ex. bad initial value of sampling),
In case of that cmdstanpy on windows will hangs up. 
Because Windows have not enough buffer for pipe-stderr, so you have to use non-block IO.
I add Pseudo non-block IO extension function. (Windows have no fnctl funcitons)

In many case,Windows don't use 'utf-8'  for terminal.
But 'utf-8' was hard coded.It's no problem for just sampling(cmdstan outputs ASCII only), 
but it's harmful for install_cmdstan/install_cxx_toolcain . 

Anyway now I can run cmdstanpy on windows and Ubuntu with this patch,

#### Copyright and Licensing

Akira NODA @ Shimadzu corp.

### DIFF
--- a/cmdstanpy/__init__.py
+++ b/cmdstanpy/__init__.py
@@ -39,7 +39,7 @@ from ._version import __version__  # noqa
 from .model import CmdStanModel  # noqa
 from .stanfit import CmdStanGQ, CmdStanMCMC, CmdStanMLE, CmdStanVB  # noqa
 from .utils import set_cmdstan_path  # noqa
-from .utils import cmdstan_path, install_cmdstan, set_make_env
+from .utils import cmdstan_path, install_cmdstan, set_make_env  # noqa
 
 __all__ = [
     'set_cmdstan_path',
--- a/cmdstanpy/install_cmdstan.py
+++ b/cmdstanpy/install_cmdstan.py
@@ -117,7 +117,9 @@ def install_version(
             if proc.returncode:
                 msgs = ['Command "make clean-all" failed']
                 if stderr:
-                    msgs.append(stderr.decode(sys.stdin.encoding, errors='ignore').strip())
+                    msgs.append(stderr.
+                                decode(sys.stdin.encoding, errors='ignore')
+                                .strip())
                 raise CmdStanInstallError('\n'.join(msgs))
             print('Rebuilding version {}'.format(cmdstan_version))
         cmd = [make, 'build']
@@ -132,7 +134,9 @@ def install_version(
         while proc.poll() is None:
             if not proc.wait_newline(0.5):
                 continue
-            output = proc.qout.readline().decode(sys.stdin.encoding, errors='ignore').strip()
+            output = proc.qout.readline().\
+                decode(sys.stdin.encoding, errors='ignore').\
+                strip()
             if verbose and len(output) != 0:
                 print(output, flush=True)
         proc.wait_log()
--- a/cmdstanpy/install_cmdstan.py
+++ b/cmdstanpy/install_cmdstan.py
@@ -29,10 +29,9 @@ import urllib.request
 from collections import OrderedDict
 from pathlib import Path
 from time import sleep
-import sys
 
 from cmdstanpy import _DOT_CMDSTAN, _DOT_CMDSTANPY
-from cmdstanpy.utils import validate_dir,proc_readline_ext
+from cmdstanpy.utils import validate_dir, proc_readline_ext
 
 EXTENSION = '.exe' if platform.system() == 'Windows' else ''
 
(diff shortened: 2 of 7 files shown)


mitzimorris:
there's two different issues here - general support for platform encoding, and problems with blocking on Windows.
could you file issues for both?  and then, let's do a PR for the encoding issue and let's discuss what do to about subprocess handling on windows?

mitzimorris:
ps.  and thank you for the work you've already done!

akirayou:
OK, I'll split this PR in 2 PR.It needs few days.
1st:  I'll PR encoding problem (Almost no effect for almost all users , but It's needed for checking the effect of  2nd PR on my PC)
2nd: (1st one merged then ) I'll PR sub process handling 
Is it OK?

The sub process handling problem is this.
https://stackoverflow.com/questions/61786127/why-can-popen-stdout-readline-deadlock-and-what-to-do-about-it

Linux have 64KB pipe buffer,but windows have only 4KB pipe buffer.
The buffer size means max size of STDERR message for cmdStan or make(install_cmdstan/make  some model).4KB is too small in many case.
My 2nd PR makes the STDERR size limitation  unlimited .

And I used this workaround for 2nd PR,Because I want to run same code on windows and Linux  (and Mac).
https://stackoverflow.com/questions/375427/a-non-blocking-read-on-a-subprocess-pipe-in-python

ahartikainen:
Will the buffer fill also for the current develop?

But either way I like the changes. Very neat way to handle the encoding. Will that work in general or only when we expect output from shell or e.g. Stan where everything is ascii? (I assume yes)

Also, should we empty stderr buffer too?

akirayou:
>Will the buffer fill also for the current develop?
Yes ,but it happens only on the stan  model which have many warning message. 
 
> Stan where everything is ascii? (I assume yes)
I forgot this.Almost all case OK.

 But if you use  non ASCII compatible (ex. UTF-16) encoding terminal and if cmdstan will output ASCII-binary message.It will fail. (Almost no one use UTF-16 because of this. )

It's depends on cmdstan's build option.
if CXXFLAGS " -fexec-charset=[some terminal character encoding]"  is used .There is no problem.
But there is no CXXFLAGS, cmdstan will outputs raw ASCII.

Same problem some timeshappens for each .exe in build tools. In this reason. 
Almost all user choose ASCII compatible character encoding  for terminal encoding.

It means almost case ASCII can read as  UTF-8(linux)/SJIS(japanese win)/Big5(Taiwan/chineese win)/KS(Korea win) and so on.
But for example SJIS can not read as  UTF-8 nor ASCII.
[FYI]The Makefile of cmdstan outputs terminal's  encoded multibyte string(i.e. SJIS Japanese localized message) .So compile() method have to support terminal character encodings. But I don't know for sampe() method.

akirayou:
I've investigated  the encoding  little more.

In windows you can set non-ascii  super set character code with chcp command.
I tried chcp 37 (EBCDIC) and chcp 65000 (UTF-7). 
Result: The mingw32-make build failed in both case.

[FYI]In Ubuntu  there is no EBCDIC/UTF-7/UTF-16 language pack ,So I can not test it. (May be only IBM main frame user ) MacOS only have ASCII superset (in /usr/share/locale)

Conclusion:
This  is stancmd's limitation. stancmd needs ASCII superset character code. 
And you can assume terminal character code must be ASCII superset character code.
So it's reasonable way to use terminal character code (=sys.stdin.encoding ) for cmdstanpy's pipe.


ASCII super set character code:
Character code which contains ASCII code , There is no need to translate  ASCII to the superset character code.

mitzimorris:
are there still problems for Japanese locale?

akirayou:
Yes the problem on decode() function still remains on the develop branch. So I'm using another branch one now.
https://github.com/akirayou/cmdstanpy/commits/for_shimadzu 

I think you can close this PR as rejected, if there is no other non-ascii windows users.

ahartikainen:
Is there somewhere said officially that `sys.stdin.encoding` is the correct way to handle this?

akirayou:
There is no official way to get terminal encoding (code page) on windows python. So I used win32 API.
You can set sys.stdin.encoding manually with official way.But It's global setting.I think it's risky way. 
 https://docs.python.org/3/using/cmdline.html#envvar-PYTHONIOENCODING 
set PYTHONIOENCODING=cp932
set PYTHONLEGACYWINDOWSSTDIO=True

Even more unfortunately, there is few the UTF-8 output commands and many local code page command are used in the Makefile of building cmdstan. (I don't know about when buiding the model)
So even if you guess or manually set the terminal encoding,you need "errors=" parameter in decode() function. 

I think just adding errors parameter is even good workaround, 
Some part of messages will not readable.But almost no one read the error message.

PR #337, opened 2021-01-14
Add extract function

ahartikainen:
#### Submission Checklist

- [x] Run unit tests

#### Summary

Add option to extract ndim objects.

#### Copyright and Licensing

Ari Hartikainen

### DIFF
--- a/cmdstanpy/utils.py
+++ b/cmdstanpy/utils.py
@@ -1063,10 +1063,15 @@ class TemporaryCopiedFile:
         if self._tmpdir:
             shutil.rmtree(self._tmpdir, ignore_errors=True)
 
-            
-def extract(fit: CmdStanMCMC, parameters: Optional[List]=None, inc_diagnostics: Optional[bool]=False, inc_warmup: Optional[bool]=False) -> Dict[str, Any]:
+
+def extract(
+    fit: CmdStanMCMC,
+    parameters: Optional[List] = None,
+    inc_diagnostics: Optional[bool] = False,
+    inc_warmup: Optional[bool] = False,
+) -> Dict[str, Any]:
     """Extract ndim dictionary from CmdStanPy fit.
-    
+
     Parameters
     ----------
     fit: CmdStanMCMC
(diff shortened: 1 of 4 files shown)


ahartikainen:
@mitzimorris we should now discuss what should be the name of this function, what output shape is desired and where to put this.

I think we need similar function for CmdStanR.

mitzimorris:
will take a look at the code.

option 1:
extract is to `stan_variables` as `draws` is to `draws_pd` -
therefore, rename `stan_variables` to `stan_variables_pd` and rename `extract` to `stan_variables`
also function rename `stan_variable` to `stan_variable_pd` and make corresponding `stan_variable` function
that returns a single variable as np.ndarray.

option 2:
create utility function "as_pd" that takes either an np.ndarray or dict thereof and returns pd.dataframe.

mitzimorris:
thinking about it more - I think naming this `stan_variable` and renaming current `stan_variable` function to  `stan_variable_pd` would be the best thing to do here.
will do code review now.

cmdstanpy/utils.py
```diff
@@ -1061,3 +1061,74 @@ def __enter__(self):
...
+        col_base, *col_tail = col.replace(']', '').replace('[', ',').split(',')
+        if col_tail:
+            # gather nD array locations
+            column_groups[col_base].append(tuple(map(int, col_tail)))
+        # gather raw data locations for each parameter
+        column_locs[col_base].append(i)
+    dims = {}
+    for colname, col_dims in column_groups.items():
+        # gather parameter dimensions (assumes dense arrays)
+        dims[colname] = tuple(np.array(col_dims).max(0))
```
mitzimorris, line 1111: this is equivalent to utility function `parse_var_dims`?   does this handle multi-dim arrays?
e.g., Stan variable that is an array of matrices?
```
int d;
  matrix [d,d] d_m;
  matrix [d,d] d_m_ar[d];
```

which version is more efficient?

as per comment - this only works on dense arrays

mitzimorris (changes requested):
let's discuss design a bit more.

mitzimorris:
I like the way that the `extract` function is generalized, but this should be used by CmdStanMCMC et al classes and each class should have class method `stan_variable`, `stan_variables` which return a numpy ndarray and dict.
please forget earlier comments about pandas dataframes - we definitely want to work with numpy.ndarrays, no?

also, would it be good to have helper function `as_int` for ndarrays that should be int arrays?
this is needed for int varaiable declared in the generated quantities block.  however the CSV gets read in,
the reals should be close enough to ints that reading them in as reals and converting to ints won't be problematic.

bottom line:  this is good - if you don't have time, I can refactor - your call.

mitzimorris:
OK, branched from where you were at - https://github.com/stan-dev/cmdstanpy/tree/feature/extract-refactor

needs more unit tests and docs, but I think this does what we want.

PR #356, opened 2021-03-05
broaden type check to include numpy scalars (in addition to `ndarray`)

meereeum:
#### Submission Checklist

- [x] Run unit tests

#### Summary

JSON serialization currently falls down if the `data` dictionary values include numpy scalars (with `TypeError: $SCALAR is not JSON serializable`). Fix type-check to catch these, in addition to arrays (the `tolist()` conversion still works as expected to convert to builtin `int`/`float`).

#### Copyright and Licensing

I am the copyright holder of this work and agree to the terms of the license.

### DIFF
--- a/cmdstanpy/utils.py
+++ b/cmdstanpy/utils.py
@@ -370,7 +370,7 @@ def jsondump(path: str, data: Dict) -> None:
     """Dump a dict of data to a JSON file."""
     data = data.copy()
     for key, val in data.items():
-        if isinstance(val, np.ndarray):
+        if type(val).__module__ == 'numpy':
             val = val.tolist()
             data[key] = val
     with open(path, 'w') as fd:
--- a/test/test_sample.py
+++ b/test/test_sample.py
@@ -10,6 +10,8 @@ import unittest
 from multiprocessing import cpu_count
 from time import time
 
+import numpy as np
+
 import pytest
 from testfixtures import LogCapture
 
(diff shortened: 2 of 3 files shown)


mitzimorris:
could you please file an issue for this and say a little bit more about the data that is problematic?
I'm not sure this fix is going to do the right thing for your data, so more context here would help.
we definitely want to be able to handle numpy data types correctly.

meereeum:
Sure, just imagine passing `N=np.int32(100)` as data to your favorite model. This is the default mode for data generated by, e.g., `pybrms`. I'm using this code on my personal branch as a prerequisite to be able to run `brms`-type models. There may be other ways to check for numpy scalars, but limiting to `np.ndarray` type for proper JSON serialization is too limiting.

ahartikainen:
This looks very neat solution.

mitzimorris:
the question is, has this been tested?  I think what will happen is that the input will be a list when the Stan program will expect N to be a scalar; therefore, I don't think this will work.  however, a unit test should be easy to construct using the example bernoulli model. 

while I really appreciate this PR, the process is:  file an issue, provide a test case showing how things should work, and then fix the bug.  it's a real PITA, but in the long run, it's worth it.

ahartikainen:
It will be scalar.

`np.int32(3).tolist() == 3`

mitzimorris (changes requested):
please add two tests: 

- add unit test here:  https://github.com/stan-dev/cmdstanpy/blob/eba7d3b55a5f81dd422d7f696637fec672c864de/test/test_utils.py#L206. test is scalar
-  add functional test here:   https://github.com/stan-dev/cmdstanpy/blob/eba7d3b55a5f81dd422d7f696637fec672c864de/test/test_sample.py#L153-L161

ahartikainen (approved):

mitzimorris (approved):

meereeum:
thanks for seeing this through, @mitzimorris

PR #366, opened 2021-03-23
Add refresh argument to CmdStanModel.sample

teddygroves:
#### Submission Checklist

- [x] Run unit tests

#### Summary

This change allows the user to set the `CmdStanArgs` argument `refresh` by adding a matching argument to the method `CmdStanModel.sample`. 

The main reason for doing this is to control the tqdm progress bar's granularity: for example setting `refresh` to 1 makes the progress bar update every iteration. Someone also might want to do this in order to tell from the `stdout.txt` file exactly at what iteration a message was printed.

#### Copyright and Licensing

Teddy Groves

### DIFF
--- a/cmdstanpy/cmdstan_args.py
+++ b/cmdstanpy/cmdstan_args.py
@@ -628,6 +628,12 @@ class CmdStanArgs:
                     'invalid path for output files,'
                     ' cannot write to dir: {}'.format(self.output_dir)
                 ) from exc
+        if self.refresh is not None:
+            if not isinstance(self.refresh, int) or self.refresh < 1:
+                raise ValueError(
+                    'Argument refresh must be a positive integer value, '
+                    'found {}.'.format(self.refresh)
+                )
 
         if self.sig_figs is not None:
             if (
--- a/cmdstanpy/cmdstan_args.py
+++ b/cmdstanpy/cmdstan_args.py
@@ -548,7 +548,7 @@ class CmdStanArgs:
         sig_figs: str = None,
         save_diagnostics: bool = False,
         save_profile: bool = False,
-        refresh: str = None,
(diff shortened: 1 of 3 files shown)


mitzimorris:
agreed we should expose the `refresh` argument to CmdStan, but we should do it in a more general way.

technically, `refresh` is available to all methods -  https://mc-stan.org/docs/2_26/cmdstan-guide/command-line-interface-overview.html#output-control-arguments.  not that I like the CmdStan argument hierarchy, but inasmuch as the code has been organized to follow it, will continue to do so.

this means that the implementation will look more like the work done in this PR:  

- in file `cmdstan_args.py` - add argument `refresh` to class `CmdStanArgs`,  in method `compose_command`, add as part of args following keyword `output`

- in file `model.py` - add argument to methods - `sample`, `optimize`, etc.

(nothing needed to do in `stanfit.py` or `utils.py` - everything should just work for this one)

- minimal unit tests to check code.  as long as `refresh` is a positive int, all good.

teddygroves:
I'll update the other `model`methods and add some unit tests.

>     * in file `cmdstan_args.py` - add argument `refresh` to class `CmdStanArgs`,  in method `compose_command`, add as part of args following keyword `output`

I think this is already done - `CmdStanArgs` and `compose_command` seem to already be able to handle the `refresh` argument. Is there anything that needs to be changed there?

mitzimorris (approved):
looks great - just one more thing - in `CmdStanArgs.__init__` :  `refresh` should have type hint `int`, not `str`.

mitzimorris:
>  think this is already done - CmdStanArgs and compose_command seem to already be able to handle the refresh argument. Is there anything that needs to be changed there?

you're correct - that was implemented - just never exposed.  great!

Issue #345, opened 2021-02-05
allow user to create CmdStanMCMC et al from output csv files


mitzimorris:
#### Summary:

If a user wants to work with a saved sample, it should be possible instantiate a fitted model object
from CSV files and user inputs.

#### Description:

We're doing this in both standalone generated quantities and in unit tests -

https://github.com/stan-dev/cmdstanpy/blob/68be06ed75d7d85b3ce22b11352b34e5d621cbdf/cmdstanpy/model.py#L853

Example from unit test:
```
       # construct fit using existing sampler output
        exe = os.path.join(DATAFILES_PATH, 'logistic' + EXTENSION)
        rdata = os.path.join(DATAFILES_PATH, 'logistic.data.R')
        sampler_args = SamplerArgs(iter_sampling=100)
        cmdstan_args = CmdStanArgs(
            model_name='logistic',
            model_exe=exe,
            chain_ids=[1, 2, 3, 4],
            seed=12345,
            data=rdata,
            output_dir=DATAFILES_PATH,
            sig_figs=17,
            method_args=sampler_args,
        )
        runset = RunSet(args=cmdstan_args)
        runset._csv_files = [
            os.path.join(DATAFILES_PATH, 'logistic_output_1.csv'),
            os.path.join(DATAFILES_PATH, 'logistic_output_2.csv'),
            os.path.join(DATAFILES_PATH, 'logistic_output_3.csv'),
            os.path.join(DATAFILES_PATH, 'logistic_output_4.csv'),
        ]
        retcodes = runset._retcodes
        for i in range(len(retcodes)):
            runset._set_retcode(i, 0)
        fit = CmdStanMCMC(runset)
```

In order to make this a user-facing utility, we need to determine
the minimum number of piece of information user needs to supply
in addition to list of Stan CSV files from a run, plus method.

#### Additional Information:
Provide any additional information here.

mitzimorris:
the CmdStanMCMC et al objects don't provide much functionality other than `summary` and `stan_variables`, given this, what are the compelling use cases?

ahartikainen:
This code needs better settings handling.
(from https://github.com/arviz-devs/arviz/blob/main/arviz/tests/external_tests/test_data_cmdstanpy.py )
```
from cmdstanpy import CmdStanMCMC
from cmdstanpy.stanfit import RunSet
from cmdstanpy.model import CmdStanArgs, SamplerArgs

args = CmdStanArgs(
                "dummy.stan",
[...]

PR #402, opened 2021-07-06
Feature/345 from csv to cmd stan mcmc et al

mitzimorris:
#### Submission Checklist

- [x] Run unit tests

#### Summary

Created class method `from_csv` for CmdStanMCMC, CmdStanMLE, and CmdStanVB objects to reconstitute inference object from saved CSV files.  Utility function `validate_csvfiles` generalizes the step of finding the (list of) csvfiles and doing a quick parse of the header of the first file.  The list of files and config information is then used by the class methods to create the object itself.

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting (this will be you or your assignee, such as a university or company):   Columbia University

### DIFF
--- a/cmdstanpy/stanfit.py
+++ b/cmdstanpy/stanfit.py
@@ -465,8 +465,8 @@ class CmdStanMCMC:
         for file in os.listdir(dir):
             if file.endswith(".csv"):
                 csvfiles.append(os.path.join(dir, file))
-        num_chains = len(csvfiles)
-        if num_chains == 0:
+        chains = len(csvfiles)
+        if chains == 0:
             raise ValueError('No CSV files found in directory {}'.format(dir))
         config_dict = {}
         try:
@@ -494,10 +494,10 @@ class CmdStanMCMC:
         cmdstan_args = CmdStanArgs(
             model_name=config_dict['model'],
             model_exe=config_dict['model'],
-            chain_ids=None,
+            chain_ids = [x + 1 for x in range(chains)],
             method_args=sampler_args,
         )
-        runset = RunSet(args=cmdstan_args)
+        runset = RunSet(args=cmdstan_args, chains=chains)
         runset._csv_files = csvfiles
         for i in range(len(runset._retcodes)):
             runset._set_retcode(i, 0)
--- a/cmdstanpy/stanfit.py
+++ b/cmdstanpy/stanfit.py
@@ -15,7 +15,7 @@ import numpy as np
 import pandas as pd
 
 from cmdstanpy import _TMPDIR, _CMDSTAN_WARMUP, _CMDSTAN_SAMPLING, _CMDSTAN_THIN
-from cmdstanpy.cmdstan_args import CmdStanArgs, Method
+from cmdstanpy.cmdstan_args import SamplerArgs, CmdStanArgs, Method
 from cmdstanpy.utils import (
     EXTENSION,
     check_sampler_csv,
@@ -26,6 +26,7 @@ from cmdstanpy.utils import (
     get_logger,
     parse_sampler_vars,
     parse_stan_vars,
+    scan_config,
     scan_generated_quantities_csv,
     scan_optimize_csv,
     scan_variational_csv,
@@ -445,6 +446,63 @@ class CmdStanMCMC:
         )
         # TODO - hamiltonian, profiling files
         return repr
+
+    @classmethod
+    def from_csv(cls, dir: str):
+        """
+        Instantiate CmdStanMCMC from sampler csvfiles
+        :param dir: directory path
+        """
+        if dir is None:
+            raise ValueError('Must specify path to directory with Stan CSV files.')
+        csvfiles = []
+        for file in os.listdir(dir):
+            if file.endswith(".csv"):
+                csvfiles.append(os.path.join(dir, file))
+        num_chains = len(csvfiles)
+        if num_chains == 0:
+            raise ValueError('No CSV files found in directory {}'.format(dir))
+        config_dict = {}
+        try:
+            with open(csvfiles[1], 'r') as fd:
+                scan_config(fd, config_dict, 0)
+        except (IOError, OSError, PermissionError) as e:
+            raise ValueError(
+                'Cannot read CSV file: {}'.format(csvfiles[1])
+            ) from e
+        if not 'model' in config_dict or not 'method' in config_dict:
+            raise ValueError(
+                "File {} is not a Stan CSV file.".format(csvfiles[1])
+                )
+        if config_dict['method'] != 'sample':
+            raise ValueError(
+                "File {} isn't a Stan sampler CSV file, "
+                "found method = {} ".format(csvfiles[1], config_dict['method'])
+                )
+        sampler_args = SamplerArgs(
+            iter_sampling=config_dict['num_samples'],
+            iter_warmup=config_dict['num_warmup'],
+            thin=config_dict['thin'],
+            save_warmup=config_dict['save_warmup']
+        )
+        cmdstan_args = CmdStanArgs(
+            model_name=config_dict['model'],
+            model_exe=config_dict['model'],
+            chain_ids=list(range(1,num_chains+1)),
+            method_args=sampler_args,
+        )
+        runset = RunSet(args=cmdstan_args)
+        runset._csv_files = csvfiles
+        for i in range(len(runset._retcodes)):
+            runset._set_retcode(i, 0)
+        fit = CmdStanMCMC(runset)
+        try:
+            fit.draws()
+        except (IOError, OSError, PermissionError) as e:
+            raise ValueError(
+                'An error occured processing the CSV files:\n\t{}'.format(str(e))
+            ) from e
+        return fit
 
     @property
     def chains(self) -> int:
--- a/cmdstanpy/stanfit.py
+++ b/cmdstanpy/stanfit.py
@@ -1450,8 +1450,8 @@ def from_csv(
             for i in range(len(runset._retcodes)):
                 runset._set_retcode(i, 0)
             fit = CmdStanVB(runset)
-        return fit
     except (IOError, OSError, PermissionError) as e:
         raise ValueError(
             'An error occured processing the CSV files:\n\t{}'.format(str(e))
         ) from e
+    return fit
--- a/cmdstanpy/stanfit.py
+++ b/cmdstanpy/stanfit.py
@@ -1,6 +1,7 @@
 """Container objects for results of CmdStan run(s)."""
 
 import copy
+import glob
 import logging
 import math
 import os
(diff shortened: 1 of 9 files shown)


cmdstanpy/utils.py
```diff
@@ -390,7 +390,7 @@ def rdump(path: str, data: Dict) -> None:
             print(line, file=fd)
 
 
-def rload(fname: str) -> dict:
+def rload(fname: str) -> Dict:
```
ahartikainen, line 393:
```suggestion
def rload(fname: str) -> dict:
```

I think py39 recommends using base items and not typing items, but really I don't care.

ahartikainen (commented):
Reason for class method (which doesn't use the class) is that users find the method more easily?

What if we had one from_csv function which has a keyword for the `method`? It could even be optional and validate_csvfiles could do a bit of magic trying to find the correct class, if method is not defined (e.g. method=None).

cmdstanpy/utils.py
```diff
@@ -1017,6 +1017,47 @@ def install_cmdstan(
...
+    if len(csvfiles) == 0:
+        raise ValueError('No CSV files found in directory {}'.format(dir))
+    try:
+        with open(csvfiles[0], 'r') as fd:
+            scan_config(fd, config_dict, 0)
+    except (IOError, OSError, PermissionError) as e:
+        raise ValueError('Cannot read CSV file: {}'.format(csvfiles[0])) from e
+    if 'model' not in config_dict or 'method' not in config_dict:
+        raise ValueError("File {} is not a Stan CSV file.".format(csvfiles[0]))
+    if config_dict['method'] != method:
```
WardBrian, line 1051:
```suggestion
    if method is not None and config_dict['method'] != method:
```

Changing this would allow a relatively simple dispatch function which calls validate_csv and then calls one of the three class methods based on the method keyword

WardBrian (commented):
I agree that a top-level from_csv seems like a good idea. Particularly because 'method' already is part of the dictionary, it seems very plausible to return the appropriate type. I'm not opposed to leaving a from_csv in each class, however

mitzimorris:
thanks - agreed, top-level "from_csv" makes sense.  will refactor accordingly

cmdstanpy/utils.py
mitzimorris: we need to be consistent - elsewhere found type hint as `Dict`.
whatever is best practice is fine by me.

found this:  https://mypy.readthedocs.io/en/stable/cheat_sheet_py3.html
```
# In Python 3.8 and earlier, the name of the collection type is
# capitalized, and the type is imported from 'typing'
x: List[int] = [1]
```

cmdstanpy/stanfit.py
```diff
@@ -1457,3 +1337,121 @@ def save_csvfiles(self, dir: str = None) -> None:
         :param dir: directory path
         """
         self.runset.save_csvfiles(dir)
+
+
+def from_csv(
+    dir: str = None, method: str = None
+) -> Union[CmdStanMCMC, CmdStanMLE, CmdStanVB]:
```
WardBrian, line 1344: This is a pretty clear place where [typing literals and overloading](https://mypy.readthedocs.io/en/stable/literal_types.html) would be very helpful. Any users of mypy will appreciate if we annotate this in such a way that the return type isn't a union in the unambiguous cases

cmdstanpy/stanfit.py
```diff
@@ -1457,3 +1337,121 @@ def save_csvfiles(self, dir: str = None) -> None:
...
+    :param dir: directory path
+    :param method: CmdStan method name
+
+    :return: either a CmdStanMCMC, CmdStanMLE, or CmdStanVB object
+    """
+    if dir is None:
+        raise ValueError('Must specify directory of Stan CSV files.')
+    if not os.path.exists(dir):
+        raise ValueError('Directory {} not found.'.format(dir))
+    if method is None or method not in ['sample', 'optimize', 'variational']:
```
WardBrian, line 1358: I still believe it would be a good idea to allow method=None and figure it out from the files in this case. It would require moving some of the checks around, but I think it would be a good idea.

If we don't want to do that, I think we should remove the default assignments to None in the method signature. That will promote not supplying the arguments up to a syntax level rather than a valueerror level.

WardBrian (commented):
A few comments but this is looking very good!

cmdstanpy/stanfit.py
mitzimorris: makes sense.

cmdstanpy/stanfit.py
```diff
@@ -1330,3 +1337,113 @@ def save_csvfiles(self, dir: str = None) -> None:
         :param dir: directory path
         """
         self.runset.save_csvfiles(dir)
+
+
+def from_csv(dir: str = None) -> Union[CmdStanMCMC, CmdStanMLE, CmdStanVB]:
```
WardBrian, line 1342: It may be overly nitpicky, but I think including the method keyword as optional would be helpful in a statically typed context. It would be annoying to have to disambiguate the union type after every call if you know the method ahead of time. 

Other than the failing tests, this looks good to me

mitzimorris: happy to add some flexibility here.    changed logic, added checks and tests, also missing test files.  this time for sure?

cmdstanpy/stanfit.py
mitzimorris: if I'm understanding correctly, you're suggesting the following?

```
SampleMethod = Literal["sample"]
OptimizeMethod = Literal["optimize"]
@overload
def from_csv(dir: str, method: SampleMethod) -> CmdStanMCMC:
@overload
def from_csv(dir: str, method: OptimizeMethod) -> CmdStanMLE:

def from_csv(
    dir: str = None, method: str = None
) -> Union[CmdStanMCMC, CmdStanMLE, CmdStanVB]:
```

WardBrian: This is my only remaining comment.

Here is what I believe a useful set of type signatures would be:

```python
# add this to the existing imports
from typing import Literal, overload

@overload
def from_csv(dir: str, method: Literal["sample"]) -> CmdStanMCMC: ...
@overload
def from_csv(dir: str, method: Literal["optimize"]) -> CmdStanMLE: ...
@overload
def from_csv(dir: str, method: Literal["variational"]) -> CmdStanVB: ...
@overload
def from_csv(dir: str, method: str) -> Union[CmdStanMCMC, CmdStanMLE, CmdStanVB]: ...

# this one is as existing and has actual implementation
def from_csv(
    dir: str = None, method: str = None
) -> Union[CmdStanMCMC, CmdStanMLE, CmdStanVB]:
```

WardBrian (commented):
This looks good to me minus the mypy/typing nice-to-have I included below. If you don't want such a complicated type structure in the file that's understandable and we can merge.

Just to check, we have no interest in instantiating a CmdStanGQ object like this, right?

WardBrian:
Actually, after checking the docs it seems Literal isn't available from `typing` until python 3.8 -- it was in `typing_extensions` prior. I would make a comment that it would be nice to do something like that one day, but leave it as is for now and it's good to go in my eyes.

mitzimorris:
> Just to check, we have no interest in instantiating a CmdStanGQ object like this, right?

been asking myself the same question.  easy to do:  assemble just the generated quantities variables.  not easy to do: assemble generated quantities and other model variables.

mitzimorris:
> Actually, after checking the docs it seems Literal isn't available from typing until python 3.8 -- it was in typing_extensions prior. I would make a comment that it would be nice to do something like that one day, but leave it as is for now and it's good to go in my eyes.

I'm all for more typing - that's why initial design was for class methods.   agreed - nice to do in the future and thanks, I learned something today!

cmdstanpy/stanfit.py
```diff
@@ -1330,3 +1337,133 @@ def save_csvfiles(self, dir: str = None) -> None:
         :param dir: directory path
         """
         self.runset.save_csvfiles(dir)
+
+
+def from_csv(
+    dir: str = None, method: str = None
```
ahartikainen, line 1343: I would just ask for a list of paths for csv files. Using a directory is error-prone when there are multiple different runs in the same folder.

ahartikainen (commented):
One comment about the `from_csv` API.

It could have a possibility for both (list of filepaths and dir) or even a glob statement?

mortonjt:
Hi @mitzimorris thank you for inviting me to review.

I believe that this PR addresses one of the issues we had passing around `CmdStanMCMC` objects, by explicitly allowing for these models to instantiated from file paths, that should address this problem.

Looking back at my initial comments at #345 it looks like I misunderstood the functionality of `CmdStanMCMC`; if I understand the source code correctly, it looks like the `CmdStanMCMC` object contains static MCMC draws, meaning that it currently isn't possible to resume the MCMC sampler to further explore the posterior distribution.  If so the use cases that I highlighted earlier aren't relevant to this PR.

Having some way resume training *could* lead to interesting future use cases -- you can imagine scenarios where you may either want to continue exploring the posterior distribution, or may want to update your model with new / private datasets.  However, it looks like this type of functionality is outside the scope of the `CmdStanMCMC` object.

mitzimorris:
> Having some way resume training _could_ lead to interesting future use cases -- you can imagine scenarios where you may either want to continue exploring the posterior distribution, or may want to update your model with new / private datasets. However, it looks like this type of functionality is outside the scope of the `CmdStanMCMC` object.

automatically resuming adaptation is a much-requested feature.  here's a good description of what would need to be done - https://discourse.mc-stan.org/t/current-state-of-checkpointing-in-stan/12348/13?u=mitzimorris

cmdstanpy/stanfit.py
mitzimorris: how about:
- required argument `dir`
- optional arguments either `pattern` - the glob or `names` - a list of csv filenames.
if only `dir` is specified, then get all csvfiles in the directory.

WardBrian: I think the simplest api is something like

`def from_csv(path: str | list[str], method: str = None)`

If path is a list, it must be a list of individual files. If it is not a list, it would be a glob if it contains a '*' character, or a dir otherwise. This would generate a little more boilerplate but avoids the issue of having mutually exclusive arguments

cmdstanpy/utils.py
ahartikainen: Then it is ok

cmdstanpy/stanfit.py
ahartikainen: Also, I think we should recommend users trying pathlib

    paths = list(pathlib.Path("path/to/folder").glob("my_run_x_*.csc"))

And then manually check that the files are correct. Otherwise debugging why the files are not seen could be hard.

cmdstanpy/stanfit.py
```diff
@@ -1340,39 +1341,68 @@ def save_csvfiles(self, dir: str = None) -> None:
...
         )
-    if not os.path.exists(dir):
-        raise ValueError('Directory {} not found.'.format(dir))
+
     csvfiles = []
-    for file in os.listdir(dir):
-        if file.endswith(".csv"):
-            csvfiles.append(os.path.join(dir, file))
+    if isinstance(path, list):
+        csvfiles = list
```
ahartikainen, line 1375:
```suggestion
        csvfiles = path
```

cmdstanpy/stanfit.py
```diff
@@ -1340,39 +1341,68 @@ def save_csvfiles(self, dir: str = None) -> None:
...
+        else:
+            raise ValueError('Invalid path specification: {}'.format(path))
+    else:
+        raise ValueError('Invalid path specification: {}'.format(path))
+
     if len(csvfiles) == 0:
-        raise ValueError('No CSV files found in directory {}'.format(dir))
+        raise ValueError('No CSV files found in directory {}'.format(path))
+    for file in csvfiles:
+        if not (os.path.exists(file) and file.endswith('.csv')):
```
ahartikainen, line 1400:
```suggestion
        if not os.path.exists(file):
```
Do we want to force `.csv`?

ahartikainen (commented):
This is starting to look good.

rok-cesnovar:
Just FYI: cmdstanr does this with `as_cmdstan_fit` functions that return the fit objects. Seems cmdstanpy has gone another direction, not sure we need to coordinate on it though.

mitzimorris:
@rok-cesnovar - agree, not sure we need to coordinate.
unlike `as_cmdstan_fit`, `from_csv` creates a RunSet object, so all methods should work.

cmdstanpy/stanfit.py
mitzimorris: good question.  certainly not necessary - if it's not a Stan CSV file, it won't parse.

(joke for @ahartikainen :  clearly my time in Sweden made me overly cautious on behalf of the user)

cmdstanpy/stanfit.py
mitzimorris: thanks, good catch.  need more unit tests.

WardBrian (approved):
This looks ready to me if all the questions are resolved

mitzimorris:
questions resolved

mortonjt:
Thanks @mitzimorris !

PR #404, opened 2021-07-08
Support xarray draws object

WardBrian:
#### Submission Checklist

- [x] Run unit tests

#### Summary

Add an optional dependency on [xarray](https://xarray.pydata.org/en/stable/index.html) and add the ability to call `.draws_xr()` on a CmdStanMCMC object. Modeled after `.draws_pd()`

screenshot

This allows more natural processing of multi-dimensional quantities, and should interop slightly easier with packages like arviz which use xarray extensively. 

#### Copyright and Licensing

### DIFF
--- a/cmdstanpy/stanfit.py
+++ b/cmdstanpy/stanfit.py
@@ -14,7 +14,14 @@ from typing import Dict, List, Tuple
 import numpy as np
 import pandas as pd
 
-from cmdstanpy import _TMPDIR, _CMDSTAN_WARMUP, _CMDSTAN_SAMPLING, _CMDSTAN_THIN
+try:
+    import xarray as xr
+
+    XARRAY_INSTALLED = True
+except ImportError:
+    XARRAY_INSTALLED = False
+
+from cmdstanpy import _CMDSTAN_SAMPLING, _CMDSTAN_THIN, _CMDSTAN_WARMUP, _TMPDIR
 from cmdstanpy.cmdstan_args import CmdStanArgs, Method
 from cmdstanpy.utils import (
     EXTENSION,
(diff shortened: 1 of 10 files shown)


cmdstanpy/stanfit.py
```diff
@@ -958,6 +965,61 @@ def draws_pd(
...
+        dims = ("draws", "chains")
+        for var in vars:
+            draw1 = 0
+            if not inc_warmup and self._save_warmup:
+                draw1 = self.num_draws_warmup
+            col_idxs = self._metadata.stan_vars_cols[var]
+            var_dims = dims + tuple(
+                f"{var}_dim_{i}" for i in range(len(self.stan_vars_dims[var]))
+            )
+            data[var] = (var_dims, np.squeeze(self._draws[draw1:, :, col_idxs]))
```
ahartikainen, line 1017: What is this step doing? One needs to be really careful when using squeeze and usually, it doesn't work like it is supposed to.

ahartikainen (commented):
Hi, added some comments.

cmdstanpy/stanfit.py
WardBrian: The first statement generates dimension names like "mu_dim_0", "mu_dim_1" if the stan variable has multiple dimensions. The squeeze exists because if there is only a single column index, the return of `self._draws[draw1:, :, col_idxs]` has shape (chains, dims, 1) but this is not desirable

cmdstanpy/stanfit.py
```diff
@@ -958,6 +965,61 @@ def draws_pd(
...
+
+        return xr.Dataset(data, coords=coordinates, attrs=attrs).transpose(
+            'chains', 'draws', ...
```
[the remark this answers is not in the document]
WardBrian: The first order reason is that it made more sense to me that the underlying dimensionality be with chain first, so that if you selected model.draws_xr().alpha[0] you get all the draws for a single chain, rather than the first draw for each of the four chains.

Secondly, the first thing I did after writing this code was test it by using a couple arviz functions, which all complained if I didn't do a transpose, so I figured it would be a good ideal.

```python
/home/brian/miniconda3/envs/stan/lib/python3.9/site-packages/arviz/data/base.py:169: UserWarning: More chains (1000) than draws (4). Passed array should have shape (chains, draws, *shape)
  warnings.warn(
```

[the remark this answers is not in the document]
ahartikainen: Ok, sounds reasonable.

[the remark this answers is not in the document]
ahartikainen: (arviz should move to named dimensions only at some points, but until that is reality, going this is ok)

cmdstanpy/stanfit.py
ahartikainen: What happens when there is dimension size 1?

WardBrian: You mean if either of chain or draws is size one? I suppose this could happen for chain. I will look into a refactor that avoids squeeze

ahartikainen: If there is a vector of size 1

WardBrian: I think that would still show up as having a dimension in stan_var_dims, as opposed to the scalar values which have unit as their dimension, so it should be possible to handle the two separately. 

I think if forced to choose between all scalars having a phony dimension and length-1 vectors being reduced to scalars, I would pick the latter

mitzimorris: this is correct - we're parsing the column names, a scalar Stan variable name cannot contain '['.  if there is a Stan vector or array of length 1, it will have column name `foo[1]` and therefore will have a dimension.

cmdstanpy/stanfit.py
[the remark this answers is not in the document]
mitzimorris: if you wanted something other than all the draws for a variable, you would be looking at a chain's worth of draws to check for R-hat or the like - does this change the underlying data layout?  the draws nd array is column major so that you get the right memory locality - https://mc-stan.org/cmdstanpy/api.html#cmdstanpy.CmdStanMCMC.draws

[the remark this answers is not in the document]
WardBrian: For 4 chains and 1000 draws the output is an array of 4 1000-long arrays, rather than the 4000-long the existing draws would be. I believe this would still support the same memory locality

mitzimorris (changes requested):
this looks good.  unit tests, please.

cmdstanpy/stanfit.py
ahartikainen: Ok, great.

So length 1 chain / draw cases needs to be resolved.

WardBrian: Yes, I will refactor to accommodate

WardBrian:
> unit tests, please.

Indeed I should add some - any suggestions for what we’d like to test? So far as I could tell browsing through the existing tests, it seems like all we guarantee about `.draws_pd` is a few constraints about the size/shape of the output

mitzimorris:
> all we guarantee about .draws_pd is a few constraints about the size/shape of the output

right.  the `draws` function has more tests.   if we assume that the draws array is good, we only need to test that we've done the right thing on the conversion.

the real test is to see if this interfaces with Arviz - cf https://github.com/arviz-devs/arviz/blob/ae83908c25ac800f97a27c7bf146ce8984ab4bd2/arviz/data/io_cmdstanpy.py#L600

ahartikainen:
I would not change anything in arviz until we stop support older versions.

WardBrian:
> curious to see if this interfaces with Arviz

I believe it will certainly make it easier. One of my motivations for this was to make it easier to use CmdStanPy with the user-facing Arviz functions.

Consider something like plotting the highest density interval. At present, the shortest way I could find to do so was 

```python
# doesn't work, no way to specific parameter:
az.plot_hdi(d2.year, model)
# works:
az.plot_hdi(d2.year, hdi_data=az.hdi(model).mu)
```
There is now a slightly more natural option:
```python
az.plot_hdi(d2.year, m4_7.draws_xr().mu)
```

test/data/.gitignore
```diff
@@ -0,0 +1,8 @@
+# this line ignores everything
+*
+!/**/
+# and then this 'unignores' anything with an extension
```
mitzimorris, line 4: cool!

mitzimorris (approved):
looks good!

WardBrian:
> unit tests not happy, though - need more config?

Whoops, I thought the CI would install requirements-optional as well. Should be fixed now once it re-runs

Issue #407, opened 2021-07-14
method `sample` has parameter `validate_csv` - is this useful?


mitzimorris:
#### Summary:
Set up tests to profile performance to see whether or not
`validate_csv = False` makes any appreciable difference whatsoever.
If not, we should eliminate this option.  This will simplify logic in the CmdStanMCMC
methods.  Right now, there are checks all over the place, e.g.

```
if self._metadata is None:
            self.validate_csv_files()
```

If we always call validate_csv_files during initialization, the code would be much cleaner.

#### Description:
The CmdStanMCMC class processes the sample CSV output files in two stages:

 - `validate_csv` scans a set of CSV output files, makes sure that config and size and shape of the data are compatible.  by default, this is done during initialization.

 - `assemble_drawset` does a 2nd pass to get the draws and convert them from ascii csv to float.  this happens the first time that the user tries to access either the stepsize, metric, or draws.  this is not done automatically to save both memory and processing time.

The CmdStanModel method `sample` has boolean parameter:  `validate_csv`:

>  :param validate_csv: If ``False``, skip scan of sample csv output file.
>             When sample is large or disk i/o is slow, will speed up processing.
>             Default is ``True`` - sample csv files are scanned for completeness
>             and consistency.
> 

I'm not sure this is actually true or useful.  When the sample is large but disk i/o is good,
the scan reads the file line-by-line and most lines then undergo a `split` operation.
When the sample isn't large, even slow disk i/o is fine.
When the sample is large and disk i/o is slow, then the sampler run itself will be slow,
and the fraction of time spent validating the output CSV files might not be significant.
Likewise, for complex models, most of the time will be spent doing the sampling,
not validating the CSV files.

#### Additional Information:
Provide any additional information here.

WardBrian:
Placing my comments here rather than in #408: I think it is entirely reasonable to remove this argument. When so many of the useful functions just end up calling `validate_csv_files` immediately anyway, it seems worthwhile to just call in the constructor. Even if there is a performance hit, which as you point out is dubious, I think it happening at initialization rather than at some semi-arbitrary point downstream when a method is called is for the best.

mitzimorris:
incorporating this as part of #408

PR #408, opened 2021-07-14
Feature/405 extend stan variables

mitzimorris:
#### Submission Checklist

- [x] Run unit tests

#### Summary

This PR generalizes the `stan_variables` function to optimize and variational methods, which requires the following:

- expose the `InferenceMedatadata` class
- populate CmdStanMCMC, CmdStanMLE, and CmdStanVB objects attribute `_metadata` during object initialization
- expose attr `_metadata` as property `metadata`
- generalize the term `sampler` to `method` when referring to output columns which end in `__`

#### Copyright and Licensing

### DIFF
--- a/cmdstanpy/stanfit.py
+++ b/cmdstanpy/stanfit.py
@@ -394,7 +394,6 @@ class InferenceMetadata:
         """
         return copy.deepcopy(self._stan_vars_cols)
 
-
     @property
     def stan_vars_dims(self) -> Dict[str, Tuple[int]]:
         """
@@ -492,7 +491,6 @@ class CmdStanMCMC:
         thinned sampling iterations.
         """
         return int(math.ceil((self._iter_sampling) / self._thin))
-
 
     @property
     def metadata(self) -> InferenceMetadata:
--- a/cmdstanpy/stanfit.py
+++ b/cmdstanpy/stanfit.py
@@ -1132,9 +1132,6 @@ class CmdStanMLE:
         shape = ()
         if len(col_idxs) > 0:
             shape = self._metadata.stan_vars_dims[name]
-        print('xs: {}'.format(xs))
-        print('len xs: {}'.format(len(xs)))
-        print('shape: {}'.format(shape))
         return np.array(xs).reshape(shape)
 
     def stan_variables(self) -> Dict[str, np.ndarray]:
--- a/cmdstanpy/stanfit.py
+++ b/cmdstanpy/stanfit.py
@@ -1135,7 +1135,7 @@ class CmdStanMLE:
         if len(col_idxs) > 0:
             shape = self._metadata.stan_vars_dims[name]
         return np.ndarray(
-            shape=shape, buffer=np.array(xs)
+            np.array(xs).reshape(shape)
         )
 
     def stan_variables(self) -> Dict[str, np.ndarray]:
@@ -1384,7 +1384,7 @@ class CmdStanVB:
         if len(col_idxs) > 0:
             shape = self._metadata.stan_vars_dims[name]
         return np.ndarray(
-            shape=shape, buffer=np.array(xs)
+            np.array(xs).reshape(shape)
         )
 
     def stan_variables(self) -> Dict[str, np.ndarray]:
--- a/cmdstanpy/stanfit.py
+++ b/cmdstanpy/stanfit.py
@@ -504,6 +504,39 @@ class CmdStanMCMC:
         return self._metadata
 
     @property
+    def sampler_vars_cols(self) -> Dict:
+        """
+        Deprecated - use "metadata.method_vars_cols" instead
+        """
+        self._logger.warning(
+            'property "sampler_vars_cols" has been deprecated, '
+            'use "metadata.method_vars_cols" instead.'
+        )
+        return self.metadata.method_vars_cols
+
+    @property
+    def stan_vars_cols(self) -> Dict:
+        """
+        Deprecated - use "metadata.stan_vars_cols" instead
+        """
+        self._logger.warning(
+            'property "stan_vars_cols" has been deprecated, '
+            'use "metadata.stan_vars_cols" instead.'
+        )
+        return self.metadata.method_vars_cols
+
+    @property
+    def stan_vars_dims(self) -> Dict:
+        """
+        Deprecated - use "metadata.stan_vars_dims" instead
+        """
+        self._logger.warning(
+            'property "stan_vars_dims" has been deprecated, '
+            'use "metadata.stan_vars_dims" instead.'
+        )
+        return self.metadata.stan_vars_dims
+
+    @property
     def column_names(self) -> Tuple[str]:
         """
         Names of all outputs from the sampler, comprising sampler parameters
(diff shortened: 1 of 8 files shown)


mitzimorris:
@ahartikainen and @WardBrian - this PR still needs to be extended to the output of generated quantities.

the question is how much refactoring to do in one PR.  right now, this adds a lot of pretty similar code to the classes that wrap a runset, i.e. CmdStanMCMC et al.

things could be simplified by simply exposing the "InferenceMetadata" object via method `metadata` on all the objects.

we might also be able to take things one step further by generalizing `stan_variable` and `stan_variables` to work with either a draws array or a vector of per-parameter estimates.

also there's this issue:  https://github.com/stan-dev/cmdstanpy/issues/407

what do y'all think?

cmdstanpy/stanfit.py
```diff
@@ -1116,6 +1115,28 @@ def optimized_params_dict(self) -> OrderedDict:
         """Returns optimized params as Dict."""
         return OrderedDict(zip(self.column_names, self._mle))
 
+    def stan_variable(self, name: str) -> np.ndarray:
```
WardBrian, line 1118: Similar to the existing `stan_variable` in the MCMC object, these new methods will want fairly detailed docstrings, describing at least the dimensionality of the return type.

cmdstanpy/stanfit.py
```diff
@@ -485,12 +488,8 @@ def column_names(self) -> Tuple[str, ...]:
         and quantities of interest. Corresponds to Stan CSV file header row,
         with names munged to array notation, e.g. `beta[1]` not `beta.1`.
         """
-        if not self._validate_csv and self._metadata is None:
-            self._logger.warning(
-                'csv files not yet validated, run method validate_csv_files()'
-                ' in order to retrieve sample metadata.'
-            )
-            return None
+        if self._metadata is None:
```
WardBrian, line 491: Even if we don't go the extra step of calling validate_csv_files in the constructors, I think these changes are great -- unexpectedly returning None can be a real headache

cmdstanpy/stanfit.py
```diff
@@ -1323,6 +1371,28 @@ def variational_params_dict(self) -> OrderedDict:
...
+            raise ValueError('unknown name: {}'.format(name))
+        col_idxs = list(self._metadata.stan_vars_cols[name])
+        vals = list(self._variational_mean)
+        xs = [vals[x] for x in col_idxs]
+        shape = ()
+        if len(col_idxs) > 0:
+            shape = self._metadata.stan_vars_dims[name]
+        return np.ndarray(
+            shape=shape, buffer=np.array(xs)
+        )
```
WardBrian, line 1385: I've never seen this form of creating an ndarray before - does it have an advantage over `np.array(xs).reshape(shape)`? Merely curious - the docs seem to suggest `np.ndarray` is primarily an internal constructor, and my assumption is it does two allocations?

cmdstanpy/stanfit.py
```diff
@@ -509,61 +508,47 @@ def num_unconstrained_params(self) -> int:
...
-                'csv files not yet validated, run method validate_csv_files()'
-                ' in order to retrieve sample metadata.'
-            )
-            return None
+        if self._metadata is None:
+            self.validate_csv_files()
         return self._metadata.sampler_vars_cols
 
+    @property
+    def stan_vars_cols(self) -> Dict:
```
WardBrian, line 532: I realize that there is a large body of code in the package which _doesn't_ do this, but the type hint convention is to specify the types in the dictionary (or other collection). For example, I believe the full return type of this function would be `Dict[str, Tuple[int]]`. This may be worth a separate PR to fix up all the existing types, which I'd be happy to do if we wanted

WardBrian (commented):
This looks very useful to me! I made a few remarks:

cmdstanpy/stanfit.py
mitzimorris: this is solely due to my lack of fluency in Python.  thanks for catching this.

cmdstanpy/utils.py
```diff
@@ -672,7 +672,8 @@ def parse_method_vars(names: Tuple[str]) -> Dict[str, Tuple[int]]:
     return {v: tuple([k]) for (k, v) in enumerate(names) if v.endswith('__')}
 
 
-def parse_stan_vars(names: Tuple[str]) -> (Dict[str, Tuple[int]], Dict[str, Tuple[int]]):
+def parse_stan_vars(names: Tuple[str]) ->
+    (Dict[str, Tuple[int]], Dict[str, Tuple[int]]):
```
WardBrian: I believe this should be 

`Tuple[Dict[str, Tuple[int,...], Dict[str, Tuple[int,...]` 

`Tuple[int]` is the type of a tuple with exactly one integer element, the ... makes it variable length of all ints. Ugly, I know. 

It may be worth adding something like mypy to the CI just to catch our own declarations

mitzimorris: mypy found 139 errors.  will try not to introduce new ones in this PR.

WardBrian: I can go through and address existing ones after this PR

mitzimorris:
OK, this is ready for code review.  are the existing unit tests sufficient?
doing CmdStanGQ next.

cmdstanpy/stanfit.py
WardBrian: This still needs to be resolved here and in the other new functions

cmdstanpy/stanfit.py
```diff
@@ -1116,6 +1118,36 @@ def optimized_params_dict(self) -> OrderedDict:
...
+            raise ValueError('unknown name: {}'.format(name))
+        col_idxs = list(self._metadata.stan_vars_cols[name])
+        vals = list(self._mle)
+        xs = [vals[x] for x in col_idxs]
+        shape = ()
+        if len(col_idxs) > 0:
+            shape = self._metadata.stan_vars_dims[name]
+        return np.ndarray(
+            shape=shape, buffer=np.array(xs)
+        )
```
WardBrian, line 1139: This should also change to `np.array(xs).reshape(shape)`

mitzimorris: done!

cmdstanpy/stanfit.py
```diff
@@ -1135,7 +1135,7 @@ def stan_variable(self, name: str) -> np.ndarray:
         if len(col_idxs) > 0:
             shape = self._metadata.stan_vars_dims[name]
         return np.ndarray(
```
WardBrian, line 1137: There's no need to wrap a call to np.array in np.ndarray, so the full return statement for both of these should just be `return np.array(xs).reshape(shape)`

WardBrian (changes requested):
The tests any everything look good to me, just another minor nitpick

cmdstanpy/stanfit.py
```diff
@@ -1134,9 +1132,10 @@ def stan_variable(self, name: str) -> np.ndarray:
         shape = ()
         if len(col_idxs) > 0:
             shape = self._metadata.stan_vars_dims[name]
-        return np.ndarray(
-            np.array(xs).reshape(shape)
-        )
+        print('xs: {}'.format(xs))
+        print('len xs: {}'.format(len(xs)))
+        print('shape: {}'.format(shape))
```
WardBrian, line 1137: This seems like debugging code that was left in

mitzimorris: fixed

WardBrian (approved):
This looks good to me now

PR #412, opened 2021-07-26
Typing fixes

WardBrian:
#### Submission Checklist

- [x] Run unit tests

#### Summary

This PR mainly makes changes to the optional typing to bring them up to the standards used by mypy and other linters, as discussed here https://github.com/stan-dev/cmdstanpy/pull/408#discussion_r670471957. Any example of this is turning something like 

`dims: list = None`

into

`dims: Optional[List[int]] = None`

The package now very nearly passes `mypy --strict`, with all of the remaining errors being benign. Fixing them requires using `numpy.typing`, which isn't available to us until we make the numpy minimum version 1.21.

Some minor semantic changes were made to make the types clearer. These should have no outward-facing impact, and all existing tests still pass. This mainly included changing the default values of some variables to match their eventual types, and changing calls like `isinstance(x, number.Integral)` to the more typical `isinstance(x, int)`.

We may want to discuss including mypy in the CI checks or in the pre-commit configuration if we would like to enforce this behavior going forward

#### Copyright and Licensing

### DIFF
--- a/cmdstanpy/__init__.py
+++ b/cmdstanpy/__init__.py
@@ -25,7 +25,7 @@ _DOT_CMDSTANPY = '.cmdstanpy'
 _DOT_CMDSTAN = '.cmdstan'
 
 
-def _cleanup_tmpdir():
+def _cleanup_tmpdir() -> None:
     """Force deletion of _TMPDIR."""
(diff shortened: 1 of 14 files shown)


mitzimorris (approved):
This looks awesome!!!
Typing is a huge win for code readability and maintainability - many thanks.

PR #413, opened 2021-07-27
Feature/409 stan variables standalone gq

mitzimorris:
#### Submission Checklist

- [x] Run unit tests

#### Summary

Refactor class `CmdStanGQ`

- `CmdStanGQ` attribute `mcmc_sample` is the CmdStanMCMC object corresponding to CmdStanModel method `generate_quantities` argument `mcmc_sample`.  

- Properties `metadata` and `mcmc_sample` allow access to metadata and mcmc_sample object, respectively.

- Add methods `stan_variable`, `stan_variables`, `draws`, `draws_pd`, `draws_xr`.   All methods have optional boolean parameter `inc_warmup`.  Draws methods have additional parameter `inc_sample` so user can deal with sample plus quantities as a single numpy.ndarray, pandas.DataFrame, or xarray, as needed.  Unit tests check shape and contents.

- Deprecate properties `generated_quantities`, `generated_quantities_pd`,  `sample_plus_quantities`, and `sample_plus_quantities_pd` in favor of using `draws` et al methods.

- Add helper methods `flatten_chains` and `build_xarray_data` used by both CmdStanMCMC and CmdStanGQ objects to manage the drawset.  The former is in `utils.py` and could be exposed as a general-purposes utility function for end users.

- Per issue #416, CmdStanMCMC and CmdStanGQ no longer store pd.DataFrame of draws, just the ndarray corresponding to all draws from all chains.

- Found cut-and-paste error in deprecating CmdStanMCMC properties in favor of metadata properties, added more unit tests to demonstrate correctness.

#### Copyright and Licensing

Columbia University

### DIFF
--- a/cmdstanpy/model.py
+++ b/cmdstanpy/model.py
@@ -930,10 +930,6 @@ class CmdStanModel:
 
         :return: CmdStanGQ object
         """
-        sample_csv_files = []
-        sample_drawset_pd = None
-        chains = 0
-        save_warmup = False
         if isinstance(mcmc_sample, CmdStanMCMC):
             sample_csv_files = mcmc_sample.runset.csv_files
         elif isinstance(mcmc_sample, list):
@@ -942,8 +938,8 @@ class CmdStanModel:
                     'Expecting list of Stan CSV files, found empty list'
                 )
             try:
-                sample_fit = from_csv(mcmc_sample)
                 sample_csv_files = mcmc_sample
+                sample_fit = from_csv(sample_csv_files)
                 mcmc_sample = sample_fit
             except ValueError as e:
                 raise ValueError(
@@ -952,7 +948,6 @@ class CmdStanModel:
                         repr(e), '\n\t'.join(mcmc_sample)
                     )
                 ) from e
-
         else:
             raise ValueError(
                 'MCMC sample must be either CmdStanMCMC object'
@@ -965,7 +960,6 @@ class CmdStanModel:
                 'Sample contains saved wormup draws which will be used '
                 'to generate additional quantities of interest.'
             )
-
         generate_quantities_args = GenerateQuantitiesArgs(
             csv_files=sample_csv_files
         )
--- a/cmdstanpy/model.py
+++ b/cmdstanpy/model.py
@@ -36,7 +36,6 @@ from cmdstanpy.utils import (
     cmdstan_path,
     do_command,
     get_logger,
-    scan_sampler_csv,
 )
 
 
--- a/cmdstanpy/model.py
+++ b/cmdstanpy/model.py
@@ -928,6 +928,7 @@ class CmdStanModel:
         :return: CmdStanGQ object
         """
         if isinstance(mcmc_sample, CmdStanMCMC):
+            mcmc_fit = mcmc_sample
             sample_csv_files = mcmc_sample.runset.csv_files
         elif isinstance(mcmc_sample, list):
             if len(mcmc_sample) < 1:
@@ -937,7 +938,7 @@ class CmdStanModel:
             try:
                 sample_csv_files = mcmc_sample
                 sample_fit = from_csv(sample_csv_files)
-                mcmc_sample = sample_fit
+                mcmc_fit = sample_fit
             except ValueError as e:
                 raise ValueError(
                     'Invalid sample from Stan CSV files, error:\n\t{}\n\t'
@@ -950,9 +951,9 @@ class CmdStanModel:
                 'MCMC sample must be either CmdStanMCMC object'
                 ' or list of paths to sample Stan CSV files.'
             )
-        chains = mcmc_sample.chains
-        chain_ids = mcmc_sample.chain_ids
-        if mcmc_sample.metadata.cmdstan_config['save_warmup']:
+        chains = mcmc_fit.chains
+        chain_ids = mcmc_fit.chain_ids
+        if mcmc_fit.metadata.cmdstan_config['save_warmup']:
             self._logger.warning(
                 'Sample contains saved warmup draws which will be used '
                 'to generate additional quantities of interest.'
@@ -989,7 +990,7 @@ class CmdStanModel:
                     msg, runset.__repr__()
                 )
                 raise RuntimeError(msg)
-            quantities = CmdStanGQ(runset=runset, mcmc_sample=mcmc_sample)
+            quantities = CmdStanGQ(runset=runset, mcmc_sample=mcmc_fit)
         return quantities
 
     def variational(
--- a/cmdstanpy/model.py
+++ b/cmdstanpy/model.py
@@ -22,6 +22,7 @@ from cmdstanpy.cmdstan_args import (
 )
 from cmdstanpy.compiler_opts import CompilerOptions
 from cmdstanpy.stanfit import (
+    from_csv,
     CmdStanGQ,
     CmdStanMCMC,
     CmdStanMLE,
@@ -930,64 +931,40 @@ class CmdStanModel:
         :return: CmdStanGQ object
         """
         sample_csv_files = []
-        sample_drawset = None
+        sample_drawset_pd = None
         chains = 0
-
+        save_warmup = False
         if isinstance(mcmc_sample, CmdStanMCMC):
             sample_csv_files = mcmc_sample.runset.csv_files
-            sample_drawset = mcmc_sample.draws_pd()
-            chains = mcmc_sample.chains
-            chain_ids = mcmc_sample.chain_ids
         elif isinstance(mcmc_sample, list):
             if len(mcmc_sample) < 1:
-                raise ValueError('MCMC sample cannot be empty list')
-            sample_csv_files = mcmc_sample
-            chains = len(sample_csv_files)
-            chain_ids = [x + 1 for x in range(chains)]
+                raise ValueError(
+                    'Expecting list of Stan CSV files, found empty list'
+                )
+            try:
+                sample_fit = from_csv(mcmc_sample)
+                sample_csv_files = mcmc_sample
+                mcmc_sample = sample_fit
+            except ValueError as e:
+                raise ValueError(
+                    'Invalid sample from Stan CSV files, error:\n\t{}\n\t'
+                    ' while processing files\n\t{}'.format(
+                        repr(e), '\n\t'.join(mcmc_sample)
+                    )
+                ) from e
+
         else:
             raise ValueError(
                 'MCMC sample must be either CmdStanMCMC object'
-                ' or list of paths to sample csv_files.'
+                ' or list of paths to sample Stan CSV files.'
             )
-        try:
-            if sample_drawset is None:  # assemble sample from csv files
-                config = {}
-                # scan 1st csv file to get config
-                try:
-                    config = scan_sampler_csv(sample_csv_files[0])
-                except ValueError:
-                    config = scan_sampler_csv(sample_csv_files[0], True)
-                conf_iter_sampling = None
-                if 'num_samples' in config:
-                    conf_iter_sampling = int(config['num_samples'])
-                conf_iter_warmup = None
-                if 'num_warmup' in config:
-                    conf_iter_warmup = int(config['num_warmup'])
-                conf_thin = None
-                if 'thin' in config:
-                    conf_thin = int(config['thin'])
-                sampler_args = SamplerArgs(
-                    iter_sampling=conf_iter_sampling,
-                    iter_warmup=conf_iter_warmup,
-                    thin=conf_thin,
-                )
-                args = CmdStanArgs(
-                    self._name,
-                    self._exe_file,
-                    chain_ids=chain_ids,
-                    method_args=sampler_args,
-                )
-                runset = RunSet(args=args, chains=chains, chain_ids=chain_ids)
-                runset._csv_files = sample_csv_files
-                sample_fit = CmdStanMCMC(runset)
-                sample_drawset = sample_fit.draws_pd()
-        except ValueError as exc:
-            raise ValueError(
-                'Invalid mcmc_sample, error:\n\t{}\n\t'
-                ' while processing files\n\t{}'.format(
-                    repr(exc), '\n\t'.join(sample_csv_files)
-                )
-            ) from exc
+        chains = mcmc_sample.chains
+        chain_ids = mcmc_sample.chain_ids
+        if mcmc_sample.metadata.cmdstan_config['save_warmup']:
+            self._logger.warning(
+                'Sample contains saved wormup draws which will be used '
+                'to generate additional quantities of interest.'
+            )
 
         generate_quantities_args = GenerateQuantitiesArgs(
             csv_files=sample_csv_files
@@ -1021,7 +998,7 @@ class CmdStanModel:
                     msg, runset.__repr__()
                 )
                 raise RuntimeError(msg)
-            quantities = CmdStanGQ(runset=runset, mcmc_sample=sample_drawset)
+            quantities = CmdStanGQ(runset=runset, mcmc_sample=mcmc_sample)
         return quantities
 
     def variational(
--- a/cmdstanpy/model.py
+++ b/cmdstanpy/model.py
@@ -954,7 +954,7 @@ class CmdStanModel:
         chain_ids = mcmc_sample.chain_ids
         if mcmc_sample.metadata.cmdstan_config['save_warmup']:
             self._logger.warning(
-                'Sample contains saved wormup draws which will be used '
+                'Sample contains saved warmup draws which will be used '
                 'to generate additional quantities of interest.'
             )
         generate_quantities_args = GenerateQuantitiesArgs(
--- a/cmdstanpy/model.py
+++ b/cmdstanpy/model.py
@@ -9,9 +9,8 @@ import subprocess
 from collections import OrderedDict
 from concurrent.futures import ThreadPoolExecutor
 from multiprocessing import cpu_count
-from numbers import Real
 from pathlib import Path
-from typing import Any, Dict, List, Union
+from typing import Any, Dict, List, Optional, Union
 
 from cmdstanpy.cmdstan_args import (
     CmdStanArgs,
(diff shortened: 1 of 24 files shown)


cmdstanpy/utils.py
```diff
@@ -1020,6 +1020,25 @@ def install_cmdstan(
     return True
 
 
+def flatten_chains(draws_array: np.ndarray) -> np.ndarray:
```
WardBrian, line 1046: I am happy with this in Utils. It may be nice to write a function which goes the other way both for round-trip testing and for some users who really dig into the api.

test/test_utils.py
```diff
@@ -736,5 +737,17 @@ def test_exit(self):
             do_command(args, HERE)
 
 
+class FlattenTest(unittest.TestCase):
+    def test_good(self):
+        array_3d = np.empty((200, 4, 4))
+        flattened = flatten_chains(array_3d)
+        self.assertEqual(flattened.shape, (800, 4))
```
WardBrian, line 744: I think it is important not just to test the shape, but also that the stacking behavior is in the desired (column-major?) order. This should be rather simple to test by using something other than `np.empty` like a few concatenated `np.arange` calls, and then assertions like 'element 400 is equal to element 0' and 'element 2 is not equal to element 0'

cmdstanpy/stanfit.py
```diff
@@ -1635,3 +1917,27 @@ def from_csv(
             'An error occured processing the CSV files:\n\t{}'.format(str(e))
         ) from e
     return fit
+
+
+def add_var_to_xarray(
```
WardBrian, line 1948: I don't think this is the best name for what this function does. Perhaps something like `build_xarray_data`?

cmdstanpy/stanfit.py
```diff
@@ -1635,3 +1917,27 @@ def from_csv(
...
+    data: Dict[str, xr.DataArray],
```
WardBrian, line 1949: This type is incorrect I believe - it seems like the dictionary is `Dict[str, Tuple[Tuple[str,...], np.ndarray]`. It's not terribly far off from an xrarry DataArray object (you have everything needed to instantiate one) but it isn't actually calling xr.DataArray (which is a good thing!)

cmdstanpy/stanfit.py
```diff
@@ -1231,20 +1222,25 @@ class CmdStanGQ:
...
             )
         self.runset = runset
-        self._metadata = None
+        self._logger = logger or get_logger()
         self.mcmc_sample = mcmc_sample
         self._generated_quantities = None
-        self._column_names = scan_generated_quantities_csv(
-            self.runset.csv_files[0]
-        )['column_names']
+        self._generated_quantities_pd = None
```
WardBrian: To stay consistent with what we just merged in #412, I recommend not making this None if you can avoid it. 

See: https://github.com/stan-dev/cmdstanpy/blob/fa339ab1709780589a0144c5d87afde550ad6b77/cmdstanpy/stanfit.py#L475-L478

You can't then use a test like `if self._generated_quantities_pd is None` unfortunately, but there are easy proxies of this (for numpy arrays, we never expect the shape to be equal to `(0,)` unless it is an unpopulated dummy array, for example)

cmdstanpy/stanfit.py
```diff
@@ -1231,20 +1222,25 @@ class CmdStanGQ:
     Container for outputs from CmdStan generate_quantities run.
     """
 
-    def __init__(self, runset: RunSet, mcmc_sample: pd.DataFrame) -> None:
+    def __init__(
+        self,
+        runset: RunSet,
+        mcmc_sample: CmdStanMCMC,
+        logger: logging.Logger = None,
```
WardBrian, line 1260: Should be `Optional[logging.Logger]` to be precise

cmdstanpy/stanfit.py
```diff
@@ -1285,61 +1338,284 @@ def generated_quantities(self) -> np.ndarray:
...
-            data=self._generated_quantities, columns=self.column_names
+        if self._generated_quantities_pd is None:
+            self._generated_quantities_pd = pd.DataFrame(
+                data=flatten_chains(self._generated_quantities),
+                columns=self.column_names,
+            )
+        return self._generated_quantities_pd
+
+    def generated_quantities_xr(
+        self, vars: List[str] = None, inc_warmup: bool = False
```
WardBrian, line 1390: Second arg should be `Optional[List[str]]`

WardBrian (commented):
This is a great change. A few comments, several of which are nitpicky typing things

test/test_utils.py
mitzimorris: agree - added more sets

cmdstanpy/stanfit.py
mitzimorris: agree - renamed

cmdstanpy/stanfit.py
mitzimorris: fixed

cmdstanpy/stanfit.py
mitzimorris: fixed

cmdstanpy/stanfit.py
mitzimorris: fixed

mitzimorris:
unit tests contained an error (user error, blush) - testing `generated_quantities_xr` which has signature: 
```
def generated_quantities_xr(
        self, vars_in: Optional[List[str]] = None, inc_warmup: bool = False
    ) -> "xr.Dataset":
```
type hints are fine, but Python was happy to run the following, where arg is `str`, not `List[str]`
should we also allow single string value as well as a list of strings here?

the statement:
```
xr_var = bern_gqs.generated_quantities_xr(vars_in='y_rep')
```
resulted in error message:  `KeyError: 'y'`
because the for loop:
```
for var in vars:
```
iterated over each character in the string.

and do the same for `vars_xr`?

WardBrian:
This is a pretty standard issue with the fact that strings are perfectly happy to just act like a sequence of characters. xarray itself feature several foot guns that stem from the same thing, if I recall correctly. 

I’m more or less ambivalent on such a change. To me it seems like a pretty obvious instance of junk in, junk out, but it’s also very easy to accommodate:

```python
if isinstance(vars_in, str):
    vars = [vars_in]

```

so we may as well do so.

mitzimorris:
function `draws_xr` now has argument name `vars_in` - which goes against name used for this argument in other similar functions - I think the argument name should be `vars` and then we use a different name internally - I suggest `vars_list`.

WardBrian:
That also sounds good to me

mitzimorris:
we now have 3 different places in the code where we have an argument which is a variable name or names -
- `stan_variable` has argument `name`, only returns one variable's worth of information
- `draws_pd` has argument `params`, list of variable names
-  `draws_xr` has argument `vars` - list of variable names

I've always disliked use of `params` and prefer `vars`.  can we swap out argument names before 1.0 release, and if so, how?

WardBrian:
I prefer vars (var_names is fine but longer, so why?)

unfortunately, I believe changing keyword names is always an immediately breaking change - no way to depreciate that I know of. You can do something hacky like *add* a keyword and then warn if the old one is set to any non-default value, but that will end up breaking more things in the long run if there is more than one optional param to the function

mitzimorris:
in which case, c'est la vie.  next API.

WardBrian:
It’s a matter of opinion, but I think that’s totally within scope of a switch from 0.x to 1.0. It really only breaks calls where people are directly specifying the name of their argument, which I think is probably a lot rarer than people just using them positionally (which would continue to work)

WardBrian:
I've thought about it more, and I'm pretty sure it is possible to depreciate a keyword argument safely like so:

```python
def draws_pd(
    self, params: Optional[List[str]] = None, inc_warmup: bool = False
) -> pd.DataFrame:
    pass # function code here

# ========= becomes =======>

def draws_pd(
    self, vars: Optional[List[str]] = None, inc_warmup: bool = False,
    *, # this means everything after this is 'keyword only'
    params: Optional[List[str]] = None
):
    if params is not None:
        if vars is not None:
            raise ValueError("Cannot use both vars and (depreciated) params")
        logger.warning("Keyword params is depreciated, use 'vars' instead")
        vars = params
    pass # function code here
```

This syntax means a call like `fit.draws_pd(['alpha'], True)` works as before, as does `fit.draws_pd([params=['alpha'], inc_warmup=True)` and `fit.draws_pd([vars=['alpha'], inc_warmup=True)`, but passing three positional arguments should still be a TypeError and passing keywords for both vars and params throws a ValueError. 

I think that is safe enough to allow the refactoring we need. Happy to discuss more however.  haven't been able to find any discussion of people using this trick before, which could mean it's a new thought or could mean it has an obvious flaw I'm overlooking - or, most likely, compatibility isn't on a lot of peoples minds.

mitzimorris:
yet another question - would like to make `draws_pd` and `generate_quantities_pd` have same behavior.  this would require changing property to method.  this could break things, but I doubt there's that much code out there uses this.

WardBrian:
Not sure I follow - what’s different at the moment?

mitzimorris:
generate_quantities_pd is a property - doesn't take arguments.   but it should have args `vars` and `include_warmup`.

cmdstanpy/model.py
```diff
@@ -927,66 +927,36 @@ def generate_quantities(
...
             raise ValueError(
                 'MCMC sample must be either CmdStanMCMC object'
-                ' or list of paths to sample csv_files.'
+                ' or list of paths to sample Stan CSV files.'
+            )
+        chains = mcmc_sample.chains
+        chain_ids = mcmc_sample.chain_ids
+        if mcmc_sample.metadata.cmdstan_config['save_warmup']:
+            self._logger.warning(
+                'Sample contains saved wormup draws which will be used '
```
ahartikainen, line 957:
```suggestion
                'Sample contains saved warmup draws which will be used '
```

cmdstanpy/model.py
```diff
@@ -927,66 +927,36 @@ def generate_quantities(
...
+                ) from e
         else:
             raise ValueError(
                 'MCMC sample must be either CmdStanMCMC object'
-                ' or list of paths to sample csv_files.'
+                ' or list of paths to sample Stan CSV files.'
+            )
+        chains = mcmc_sample.chains
+        chain_ids = mcmc_sample.chain_ids
+        if mcmc_sample.metadata.cmdstan_config['save_warmup']:
```
ahartikainen, line 955: Is this (dict key) always present?

test/test_generate_quantities.py
```diff
@@ -99,67 +110,158 @@ def test_gen_quanties_mcmc_sample(self):
...
+        with LogCapture() as log:
+            logging.getLogger()
+            bern_gqs = model.generate_quantities(
+                data=jdata, mcmc_sample=bern_fit
+            )
+        log.check_present(
+            (
+                'cmdstanpy',
+                'WARNING',
+                'Sample contains saved wormup draws which will be used to '
```
ahartikainen, line 189:
```suggestion
                'Sample contains saved warmup draws which will be used to '
```

WardBrian:
Changing a property to a function can be rather confusing for any existing users (`someobject.formerproperty` is still valid code, it is just now a function object rather than the expected value) - perhaps for the time being introduce the function version under a different name and issue a depreciation warning on the property? 

It should be reasonable to re-implement the property as a call to the new function, so we can tell users exactly what to replace their code with in the warning

cmdstanpy/model.py
mitzimorris: yes, if it's a sample, this will be captured from CSV file header.

mitzimorris:
OK, will implement as `generated_quantities_draws`. and `generated_quantities_draws_pd` - or `gq_draws` and `gq_draws_pd`.  writing out `generated_quantities` make for pretty long names.

cmdstanpy/model.py
mitzimorris: good catch - fixed

WardBrian:
I think `gq` is definitely fine as the prefix for the new methods. Especially since they will be calls with arguments, you can end up with a very long line otherwise

test/test_generate_quantities.py
mitzimorris: good catch - fixed

mitzimorris:
we now have `gq_draws`, `gq_draws_pd`, and `gq_draws_xr`. 
the methods `sample_plus_quantities` et al are pretty verbose - leaving unless someone has a good, short, and descriptive prefix for `_draws`.

WardBrian:
Just so I can say I asked - would there be any merit to gq_draws (et al) simply having an `include_sample` flag, or similar, rather than it being a separate function?

mitzimorris:
yes - this is a brilliant solution!

mitzimorris:
tests are failing because of added test `pandas.testing.assert_series_equal` arg added in 1.3.0
will try for workaround in pandas 1.2

mitzimorris:
Updated PR description; this is ready for review.

At this point, both CmdStanMCMC and CmdStanGQ have methods to access draws as structured variables via `stan_variables` which returns a dict of per-variable numpy.ndarrays and methods `draws` et al. which return a single data structure, either a numpy.ndarray, pandas.DataFrame, or xarray DataSet.

Lots of methods have been deprecated, and hope to drop them when we declare 1.0

cmdstanpy/stanfit.py
```diff
@@ -1260,20 +1265,24 @@ class CmdStanGQ:
...
+    ) -> None:
         """Initialize object."""
         if not runset.method == Method.GENERATE_QUANTITIES:
             raise ValueError(
                 'Wrong runset method, expecting generate_quantities runset, '
                 'found method {}'.format(runset.method)
             )
         self.runset = runset
-        self._metadata = None
+        self._logger = logger or get_logger()
```
WardBrian, line 1281: Some other places in this PR are moving _away_ from storing the logger, but this is still adding it. Intentional?

WardBrian (commented):
Just one comment for the moment, related to #415 

Also, I think we should throw [@functools.cache](https://docs.python.org/3/library/functools.html#functools.cache) on utils.get_logger, that way we don't have to evaluate that if condition each and every call and the performance should be identical to what we were doing previously

WardBrian:
Do you want to include documentation updates in this PR or a separate one? The page on the GQ functionality lists all of it's return object's available methods/properties (we will also need to update the page for MLE and the Variational)

cmdstanpy/stanfit.py
mitzimorris: thanks, oversight - will remove.

mitzimorris:
> Do you want to include documentation updates in this PR or a separate one? The page on the GQ functionality lists all of it's return object's available methods/properties (we will also need to update the page for MLE and the Variational)

we need to do docuentation in a separate PR.  there's a lot to clean up there.

test/test_sample.py
```diff
@@ -604,9 +613,9 @@ def test_validate_big_run(self):
         self.assertEqual(fit.step_size.shape, (2,))
         self.assertEqual(fit.metric.shape, (2, 2095))
         self.assertEqual((1000, 2, 2102), fit.draws().shape)
-        phis = fit.draws_pd(params=['phi'])
+        phis = fit.draws_pd(vars=['phi'])
```
WardBrian, line 616: should we also now test with `vars='phi'` since that is allowed? More of a general comment for all tests of these kinds

cmdstanpy/stanfit.py
```diff
@@ -1288,87 +1295,462 @@ def __repr__(self) -> str:
...
+                    self.mcmc_sample.metadata.stan_vars_cols[var],
+                    0,
+                    self.mcmc_sample.draws(inc_warmup=inc_warmup),
+                )
+
+        return xr.Dataset(data, coords=coordinates, attrs=attrs).transpose(
+            'chain', 'draw', ...
+        )
+
+    def stan_variable(self, name: str, inc_warmup: bool = False) -> np.ndarray:
```
WardBrian, line 1672: Are we okay with `name` as a param or would we prefer `var` to go with the changes to the _pd functions etc?

Also, I assume we will want another (small) pr to change/depreciate the parameter names in MCMC et al's draws functions?

cmdstanpy/stanfit.py
```diff
@@ -1288,87 +1295,462 @@ def __repr__(self) -> str:
...
+        Returns the generated quantities draws as a xarray Dataset.
+        :param vars: optional list of variable names.
+        :param inc_warmup: When ``True`` and the warmup draws are present in
+            the MCMC sample, then the warmup draws are included.
+            Default value is ``False``.
+        """
+        if not XARRAY_INSTALLED:
+            raise RuntimeError(
+                "xarray is not installed, cannot produce draws array"
+            )
```
WardBrian, line 1604: I think I've found a way to test this, but I will open a separate PR after this is merged to address it. Testing something with an optional dependency can be very tricky

WardBrian (commented):
Last couple questions before approval

cmdstanpy/stanfit.py
mitzimorris: did this as part of this PR.

test/test_sample.py
mitzimorris: string keyword is tested earlier, line 564

mitzimorris:
incorporated comments into this PR, excepting xarray pkg test.   also noticed considerable inconsistency in error message formatting - changed error messages in `stanfit.py` so that first word is capitalizied.  another pass would be required it we want sentence-final period everywhere.

WardBrian:
I may be in the minority on this, but I don't find periods in error messages to be required. So long as the error tells you something meaningful, formatting is just nice to have.

Is there anything else you think this PR needs? I'm happy to give it one last once-over if you think it's good

mitzimorris:
>  formatting is just nice to have.
agreed.

a once over, or not, as you have time and patience.

mitzimorris:
updating documentation in a separate PR.

WardBrian (approved):
I went through again and didn't spot anything more. I think this is a great change

Issue #410, opened 2021-07-20
need utility function "stan_json" 


mitzimorris:
#### Summary:
per discussion on Discourse:  https://discourse.mc-stan.org/t/various-observations-on-rstan-cmdstanr-pystan-and-cmdstanpy-after-teaching-with-all-in-parallel/23455/13

the utility function `jsondump` should be made public, w/ better name.

#### Description:

> I can’t comment on the internal data-file-writing functionality as it’s not in the API. It appears to call python json because it throws an error on numpy arrays in just the same way. It would be nice to have it available to all users. (I don’t know if “expose” is strictly the right term for that.) 

#### Additional Information:
Provide any additional information here.

rok-cesnovar:
FWIW, cmdstanr exposes `write_stan_json`: https://mc-stan.org/cmdstanr/reference/write_stan_json.html

WardBrian:
Do we need to change the functionality of the function at all, or just rename and expose it?

Do we want the same with `rdump`, or is that unnecessary?

mitzimorris:
exposing this to users will require more checking the input data dictionary items and documenting allowed input types.
we should link documentation to this:  https://mc-stan.org/docs/2_27/cmdstan-guide/json.html

rok-cesnovar:
CmdStanR does not use or handle `rdump`, except if a user supplies an existing rdump file to the `$sample()` call. The JSON reader is faster in CmdStan and there is no real reason to handle it. R has faster JSON readers/writers so no sense of duplicating work.

Thus I dont think CmdStanPy needs to add that.

mitzimorris:
> Thus I dont think CmdStanPy needs to add that.

sounds good to me.  let's not expose rdump.

PR #418, opened 2021-08-03
Expose write_stan_json utility

WardBrian:
#### Submission Checklist

- [x] Run unit tests

#### Summary
Addresses #410 by exposing `jsondump` as `write_stan_json` and **removing** `rdump`. 

As part of this I am allowing the more general `Mapping` type (which essentially means `.items()` is implemented) rather than strictly `Dict`, so you can now do things like `write_stan_json(path, dataframe)` or even `write_stan_json(path, xarray_dataset)`

I also added additional checks modeled on the cmdstanr function of the same name for clarity. These checks consist of:
* Verify the type of each value is numeric, a boolean, or a collection with a meaningful error if not.
* Verify that the values present are finite (meaning not NaN or inf/-inf)

Checking these two issues in particular mean that the call to `ujson.dump` should never itself fail

#### Copyright and Licensing

Simons Foundation

### DIFF
--- a/cmdstanpy/__init__.py
+++ b/cmdstanpy/__init__.py
@@ -45,7 +45,7 @@ from .stanfit import (  # noqa
     from_csv,
 )
 from .utils import set_cmdstan_path  # noqa
-from .utils import cmdstan_path, install_cmdstan, set_make_env
+from .utils import cmdstan_path, install_cmdstan, set_make_env, write_stan_json
 
 __all__ = [
     'set_cmdstan_path',
@@ -58,4 +58,5 @@ __all__ = [
     'CmdStanVB',
     'CmdStanModel',
     'from_csv',
+    'write_stan_json',
 ]
--- a/cmdstanpy/model.py
+++ b/cmdstanpy/model.py
@@ -10,7 +10,7 @@ from collections import OrderedDict
 from concurrent.futures import ThreadPoolExecutor
 from multiprocessing import cpu_count
 from pathlib import Path
-from typing import Any, Dict, List, Optional, Union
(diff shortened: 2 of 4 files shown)


mitzimorris:
as part of this PR, per discussion on the issue, I think we should get rid of utils methods `_rdump_array` and `rdump`

mitzimorris (approved):
LGTM!

funko-unko:
Wait, I don't think it's reasonable to throw on NaNs or infs. Stan appears to handle this without problems, why should this not be allowed?

WardBrian:
JSON has no standard way of serializing NaN or infinite values. Stan has defined a standard it uses, but this is not supported by the standard json libraries in Python (`json` or `ujson`). Without the check we currently have, the function would just throw an error a few lines later when it tried to output to the file.'

In theory we could go through all the values and replace them with the strings Stan accepts, but this requires looping over a (possibly deeply nested) structure and calling `isnan` and `isfinite` on every value. I decided to model this PR on CmdStanR's function of the same name, which similarly throws an error on non-finite floating point values.

While Stan in theory can deserialize these values, is there a legitimate reason for their use in a model? I have to assume they end up causing problems when one actually tries to run HMC or another inference algorithm

nsiccha:
I've used it to denote missing values, but I guess there are other ways to do that.

WardBrian:
Because it's in the stan spec, we should probably support it eventually. This PR was more about cleaning up and structuring our existing functionality, which didn't support it, but it is a deficiency in the API. If you'd like to open an issue for it, supporting it later should be fine

Issue #415, opened 2021-07-28
code cleanup - why do we have loggers as class attributes?


mitzimorris:
#### Summary:
Remove class attributes `logger` ~~and change calls `self._logger.info` etc. to `logging.info`~~ and instead use helper function `get_logger()`https://github.com/stan-dev/cmdstanpy/blob/5aac1d4c189e7d83839b463a7794a996f68370eb/cmdstanpy/utils.py#L45-L50

#### Description:
Some classes have logger attributes.  According to Python docs, better not to - cf.[ "Logging Cookbook"](https://docs.python.org/3/howto/logging-cookbook.html#using-loggers-as-attributes-in-a-class-or-passing-them-as-parameters):

> **Using loggers as attributes in a class or passing them as parameters**
> 
> While there might be unusual cases where you’ll need to do this, in general there is no point because loggers are singletons. Code can always access a given logger instance by name using logging.getLogger(name), so passing instances around and holding them as instance attributes is pointless. Note that in other languages such as Java and C#, loggers are often static class attributes. However, this pattern doesn’t make sense in Python, where the module (and not the class) is the unit of software decomposition.

mitzimorris:
@ahartikainen and @WardBrian - is the above advice appropriate for CmdStanPy?

WardBrian:
I am not 100% familiar with logging best practice, but I think you would then use `logger.getLogger('cmdstan').info(...)`, not `logging.info(...)`, as logging on the root logger is not as useful in general. I think this would be equivalent to what we do right now as default behavior, just perhaps a bit cleaner object model without having to pass around the logging object.

Not really a general remark, but worth noting that the way it is currently handled makes the output in [cmdstanjupyter](https://github.com/WardBrian/CmdStanJupyter) very simpler as I can pass a logger rather than having to mess with the cmdstan one. But it's far from necessary to do it this way (I can just call logger.getLogger('cmdstan') in that package as well).

mitzimorris:
I like the idea of being able to call `logger.getLogger('cmdstan')` everywhere, and letting other packages do the same.

For other packages, how do we document this, and shouldn't name 'cmdstan' be 'cmdstanpy'?

mitzimorris:
OK, now I see what's going on - we have function `get_logger`:  https://github.com/stan-dev/cmdstanpy/blob/5aac1d4c189e7d83839b463a7794a996f68370eb/cmdstanpy/utils.py#L45-L50


we should use this everywhere.

[...]

PR #419, opened 2021-08-04
Code cleanup - mypy enforcement and typo correction

WardBrian:
#### Submission Checklist

- [x] Run unit tests

#### Summary

* Finishes #415. Uses `get_logger` universally, caches `get_logger`, and emits deprecation warnings if old keyword is used, to be removed in 1.0
* Fixes some typos that made it through review ('depreciated' vs 'deprecated')
* Fix some mypy errors **and add mypy to the CI**. This may be controversial, as it will end up rejecting a lot of first passes, but if we'd like to commit to having a decently typed package, that needs to be enforced automatically somewhere.
* Add tests that mock xarray not being installed to ensure correct behavior of `.draws_xr` in that case 

#### Copyright and Licensing

Simons Foundation

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -19,6 +19,13 @@ repos:
     rev: 3.9.2
     hooks:
       - id: flake8
+  - repo: https://github.com/pre-commit/mirrors-mypy
+    rev: v0.910
+    hooks:
+      - id: mypy
+        # Copied from setup.cfg
+        exclude: ^test/
+        additional_dependencies: [ numpy >= 1.21 ]
   # local uses the user-installed pylint, this allows dependency checking
   - repo: local
     hooks:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -35,12 +35,11 @@ jobs:
           pip install -r requirements-test.txt
           pip install codecov
 
-      - name: Run flake8 and pylint
+      - name: Run flake8, pylint, mypy
         run: |
-          mkdir run_tests
-          cd run_tests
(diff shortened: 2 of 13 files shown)


.pre-commit-config.yaml
```diff
@@ -19,6 +19,13 @@ repos:
     rev: 3.9.2
     hooks:
       - id: flake8
+  - repo: https://github.com/pre-commit/mirrors-mypy
+    rev: v0.910
```
mitzimorris, line 23: so this won't take effect until we hit 0.910?

mitzimorris (approved):
LGTM!

.pre-commit-config.yaml
WardBrian: That's a revision number for mypy, not our package. It tells pre-commit which version to fetch from their github releases

mitzimorris:
this looks great.  yes, I can't spell.  is there a way to spellcheck all strings in a program?

.pre-commit-config.yaml
mitzimorris: thanks.  understood.

mitzimorris:
just checking:  ready to merge?

WardBrian:
I only caught the previous error because I had misspelled my new error messages, and a search of the project revealed the same misspelling already present.

If you're serious, there does seem to be [this python package](https://pypi.org/project/spelling/) which can do spell checking in a CI. I'm downloading it now to try it out.

If we don't actually care that much about minor spelling mistakes, this should be ready

WardBrian:
Okay, so that spelling program actually did catch some errors, but there are two reasons we shouldn't use it, IMO:

1. It's very hard to configure (ignoring a word requires creating `wordlist.txt` _in the project folder_ (e.g. in cmdstan/)
2. 'depreciate' actually is a word, so the spell checker wouldn't have caught these errors anyway

mitzimorris (approved):
great!

PR #428, opened 2021-08-10
Check for tbb.dll on Windows before editing PATH

WardBrian:
#### Submission Checklist

- [x] Run unit tests

#### Summary

This is addressing https://github.com/stan-dev/cmdstan/issues/1027 from our side. At the moment, if `tbb.dll` is in the PATH, we still add `stan/lib/stan_math/lib/tbb` to the PATH. If the original TBB was in `stan/math/...`, this is fine, but if it was anywhere else (such as `<conda_env>/Library/bin`), we end up with two copies of tbb.dll in the path. 

This is probably not a big deal, except that `where.exe`, used in make/program, outputs _both_ places, and the logic in that makefile is fragile enough that this triggers a failure there. 

This is a particularly nasty bug for users, because it means you get a 'model failed to compile' error, but if you run the code a second time it will tell you that the `.exe` already exists, and indeed, it does and it works as intended. 

Proposed solution:  using `ctypes`, we can see if 'tbb' can be resolved in the current path setup and only add it if it is not.

The CmdStan makefile could (should) also be updated.

**Alternative**

In general, I think having this code in the `__init__` of CmdStanModel is a bit sloppy, as it gets run every single model, not just once per instance. I think, if it was up to me, we wouldn't do this at all, but simply trap the 'tbb not installed' error CmdStan already outputs and either do it there, or print a meaningful message to the user - automatically editing the path behind the scenes is poor practice, IMO

#### Copyright and Licensing

Simons Foundation


WardBrian:
@mitzimorris I would be interested in getting either this fix or the alternative change I suggest in the next version, if possible - it currently leads to some strange behavior with the conda-installed cmdstan, which copies `tbb.dll` into a folder on the PATH rather than adding `stan/lib/stan_math/lib/tbb` to the PATH

WardBrian:
I've also opened https://github.com/stan-dev/cmdstan/pull/1031 to make this more robust on the CmdStan side, but my comments here still apply

ahartikainen (approved):
I don't see any problems with this setup.

Just to be sure, this is using PATH defined in os.environ and not system PATH?

WardBrian:
I'm afraid that is version dependent. ctypes will use the system PATH on 3.8+ alongside anything passed to `os.add_dll_directory()`, prior it was os.environ path. 

This means on 3.8 and 3.9 the above check won't stop us from adding the same thing to the path multiple times, but it will still prevent us adding a second unique path if tbb exists in the first place

ahartikainen:
So if users has added tbb to path in python, this will not work?

Can we call subprocess and check if tbb.dll can be found?

WardBrian:
We can use a subprocess to call `where.exe tbb.dll`, but at that point I do seriously question if we want this running on every single model instantiation

ahartikainen:
Is there a problem calling where.exe multiple times?

WardBrian:
There's no problem from a functionality standpoint, it just doesn't strike me as particularly lightweight to be spawning a subprocess just to check if the user's path is set properly before we create an object. At the very least such a thing would belong in `compile()`, IMO

ahartikainen:
How long does that call take? I mean, we already call other stuff with subprocess, so adding one more call doesn't seem too big of a deal.

Do we only need to do this when compile call is called? And the compiled model then has info where tbb is (or is that also check when sampling is called?)

WardBrian:
I don't believe `where` takes any appreciable time, but of course it depends on how complicated your path variable is. At the moment the only subprocess calls in the init function are through the (optional) call to compile.

I'm not 100% sure what the behavior of the linker is here - e.g., if a compiled program needs a path to the dll it was compiled against to still function. I don't have my Windows machine available to test it at the moment, but I believe you're correct. Windows in general doesn't store paths inside compiled executables, so it likely dynamically resolves the library at runtime as well as compile time.

Regardless, this is highlighting why I don't think this behavior should be in the initializer even more - if we want to be sure this code is run before a model is used, and if we would like to now add the restriction that it is run _no more than once_ (both for path cleanliness and general code organization), we shouldn't be doing it in the object, but rather during the package load or in another utility like how the cmdstan path is handled. 

Furthermore, if we _do_ think that it's necessary for the compiled program to have a path to the dll at runtime to function, we should heavily encourage that the user set this path in a outside-Python non-dynamic way (such as `make install-tbb`, or how conda sets this up automatically), or else any model compiled by CmdStanPy would not work if you tried to run it in a shell outside Python

WardBrian:
@ahartikainen I've switched to the subprocess version for the time being. We can discuss more long-term how we want to handle looking for TBB on windows, but I didn't want it to hold up the next version release

ahartikainen:
Hi, I like this solution. It is very "applied" solution and will make life easier for our users.

WardBrian:
It looks like `subprocess.run` didn't get `capture_output` until 3.7, so I'll have to slightly refactor

ahartikainen:
What python versions do we support?

WardBrian:
We run CI on 3.6, 3.7, 3.8, and 3.9. Python 3.6 currently is still supported, reaching end-of-life this December (2021-12-23)

WardBrian:
Hopefully the third time is the charm - switched to using our internal do_command wrapper

mitzimorris (approved):
nice!

Issue #281, opened 2020-08-17
expose argument `save_iterations` on `optimize` method; return all iterations


mitzimorris:
#### Summary:
The CmdStan `optimize` method has argument `save_iterations` - when true, the output csv file contains estimates from all iterations.

#### Description:
Add argument `save_iteration` to optimize method; return all iterations from output csv file via method `iterations`.

#### Additional Information:
Provide any additional information here.

PR #450, opened 2021-09-03
Project/release 1.0 281 optimize

mitzimorris:
#### Submission Checklist

- [x] Run unit tests

#### Summary

Addresses issue #281 - the call to the `optimize` method now has argument `save_iterations`, default is `False`.  When set to `True`, will set CmdStan command line arg `save_iterations=1`, which causes the optimizer to write intermediate iterations to the Stan CSV output file.

The intermediate iterations are available either as a numpy.ndarray object over all data columns in the Stan CSV file, or as a pandas.DataFrame.   The `stan_variable` methods don't expose the intermediate iterations.  The reasoning here is that intermediate iterations are only of interest when trying to debug a model during development, or to debug an algorithm - it's not a very common use and doesn't really warrent adding more logic to `stan_variable`.  That said, it wouldn't be that hard to extend this functionality to `stan_variable` - can be added now or later.

Also brings CmdStanPy into line with CmdStanR w/r/t allowing user to retrieve estimates even is optimizer fails to converge - https://github.com/stan-dev/cmdstanr/issues/314 by adding argument `require_converged` - when `False`, use can retrieve MLE estimates (same as behavoir for method `variational`).   If the optimizer fails to converge, the resulting CmdStanMLE object will return a warning when the user accesses the estimates.

Changes to the code base:

- cmdstan_args.py - add argument `save_iterations` to `OptimizeArgs`
- model.py - add logic to handle add'l args and to handle failure to converge
- stanfit.py, class `CmdStanMLE` - logic to capture whether or not optimizer converged, and new methods to return all iterations. 
- utils.py - changed parsing logic for parsing Stan CSV files output by optimizer.

#### Copyright and Licensing


WardBrian:
Don't think I can review this until Tuesday, but it will be the first thing I do then! Right off the bat, I think `stan_variables` should feature a flag to include these like we do with `inc_sample` in GQ and `include_warmup` in MCMC, no?

cmdstanpy/stanfit.py
```diff
@@ -1397,24 +1401,56 @@ def stan_variable(
...
+                return self._all_iters[:, col_idxs].reshape(  # type: ignore
+                    dims, order='F'
+                )
+            else:
+                mle = np.expand_dims(self._mle, axis=0)  # hack for col indexing
+                return (
+                    mle[0, col_idxs]
+                    .reshape(dims, order='F')  # type: ignore
+                    .squeeze(axis=0)
+                )
```
ahartikainen: What is the original shape for mle? and what is the output shape?

cmdstanpy/utils.py
```diff
@@ -609,7 +609,7 @@ def scan_optimize_csv(path: str, save_iters: bool = False) -> Dict[str, Any]:
             if save_iters:
                 all_iters[i, :] = [float(x) for x in xs]
             if i == iters - 1:
-                mle = np.array([float(x) for x in xs], dtype=float)
+                mle = np.array([float(x) for x in xs])
```
ahartikainen, line 612:
```suggestion
                mle = np.array(xs, dtype=float)
```

Why do we have a list comprehension here?

cmdstanpy/stanfit.py
mitzimorris: in Stan, a multi-dim variable is output by the Stan 'write_array' function column major, i.e., Fortran order - cf the header row from test model `multidim_vars.stan` -

```
rep.1.1.1,y_rep.2.1.1,y_rep.3.1.1,y_rep.4.1.1,y_rep.5.1.1,y_rep.1.2.1,y_rep.2.2.1,y_rep.3.2.1,y_rep.4.2.1,y_rep.5.2.1,y_rep.1.3.1,y_rep.2.3.1,y_rep.3.3.1,y_rep.4.3.1,y_rep.5.3.1,y_rep.1.4.1,y_rep.2.4.1,y_rep.3.4.1,y_rep.4.4.1,y_rep.5.4.1,y_rep.1.1.2,y_rep.2.1.2,y_rep.3.1.2,y_rep.4.1.2,y_rep.5.1.2,y_rep.1.2.2,y_rep.2.2.2,y_rep.3.2.2,y_rep.4.2.2,y_rep.5.2.2,y_rep.1.3.2,y_rep.2.3.2,y_rep.3.3.2,y_rep.4.3.2,y_rep.5.3.2,y_rep.1.4.2,y_rep.2.4.2,y_rep.3.4.2,y_rep.4.4.2,y_rep.5.4.2,y_rep.1.1.3,y_rep.2.1.3,y_rep.3.1.3,y_rep.4.1.3,y_rep.5.1.3,y_rep.1.2.3,y_rep.2.2.3,y_rep.3.2.3,y_rep.4.2.3,y_rep.5.2.3,y_rep.1.3.3,y_rep.2.3.3,y_rep.3.3.3,y_rep.4.3.3,y_rep.5.3.3,y_rep.1.4.3,y_rep.2.4.3,y_rep.3.4.3,y_rep.4.4.3,y_rep.5.4.3,
```

I'll check to see if we need to add more tests to make sure this is really working as expected

cmdstanpy/utils.py
mitzimorris: not needed - didn't know np.array could convert strings to float.

ahartikainen: Yeah, it is not obvious

WardBrian (approved):
LGTM!

PR #457, opened 2021-09-09
fix shaping issues

ahartikainen:
#### Submission Checklist

- [ ] Run unit tests

#### Copyright and Licensing

Ari Hartikainen

### DIFF
--- a/test/test_optimize.py
+++ b/test/test_optimize.py
@@ -287,6 +287,36 @@ class CmdStanMLETest(unittest.TestCase):
         self.assertTrue('frac_60' in vars_iters)
         self.assertEqual(vars_iters['frac_60'].shape, (8,))
 
+    def test_variables_shape(self):
+        stan = os.path.join(DATAFILES_PATH, 'shape.stan')
+        model = CmdStanModel(stan_file=stan)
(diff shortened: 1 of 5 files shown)


test/test_optimize.py
```diff
@@ -285,6 +287,36 @@ def test_variables_3d(self):
...
+        for var, shape in {
+            "x": float,
+            "a": float,
+            "b": (2,),
+            "c": (2,3),
+            "d": (2,),
+            "e":(2,3),
+            "f": (4,2,3),
+            "g":(4,5,2,3),
+        }.items():
```
WardBrian, line 315: Is this really sufficient for testing that the shape is in the correct ordering? E.g. the failing example was still a 3x3 matrix, but the values were not in the correct order.

I also think we should duplicate a similar test for `sample` with iterations=1, and `variational`.

mitzimorris:
regarding tests - since this is a bug, I think it would be good to add the model reported in the issue and then add new unit tests which specifically call out this bug and test that both the shapes and contents are OK.  I can do this, working from Ari's branch.  @ahartikainen - OK by you?

PR #460, opened 2021-09-23
[1.0] Remove deprecations

WardBrian:
Closes #430. This should probably be the last thing we merge before cutting a release, hence the draft status.

#### Submission Checklist

- [x] Run unit tests
- [x] Declare copyright holder and open-source license: see below

#### Summary

- Removed all deprecated features

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting (this will be you or your assignee, such as a university or company):
Simons Foundation

By submitting this pull request, the copyright holder is agreeing to license the submitted work under the following licenses:
- Code: BSD 3-clause (https://opensource.org/licenses/BSD-3-Clause)

PR #461, opened 2021-09-24
Project/release 1.0 subprocess stream outputs

mitzimorris:
#### Submission Checklist

- [x] Run unit tests

#### Summary

This PR addresses problems that have been reported where subprocesses hang, as well as improve the behavoir of the progress bar during both sampling and installation.

Going on the theory that the reports of processes hanging were due to the stdout/stderr pipes filling up, the `do_command` was refactored to use polling to grab lines from stdout while the subprocess is running, instead of waiting for the process to complete.

This PR also refactors the `CmdStanModel` class method `_run_cmdstan` so that outputs can be streamed to the console during sampling, allowing folks who want to debug their Stan models using Stan `print` stmt to do so.  In order to make this play nicely with the tqdm progress bar, the logic needed to update the progress bar was refactored into class `SamplerProgress`.  A `SamplerProgress` object can be used as the iterable for the tqdm progress bar.  This requires a little extra bookkeeping on the outside in order to pass the total number of iterations into the progress bar on instantiation.  During warmup, the progress bar color (sic) is orange, during sampling, the color bar is blue.   This works on my machine, both in a Jupyter notebook and in the terminal.  ymmv.

The `install_cmdstan.py` program was refactored to use the `do_command` method, and I added progress bars everywhere; however on my machine this only seems to show up when I run the program itself.  I'm not sure what the problem is, also not sure if this ever worked.

#### Copyright and Licensing


cmdstanpy/install_cmdstan.py
```diff
@@ -211,21 +214,125 @@ def install_version(
...
+                pbar.close()
+        print('Installed {}'.format(cmdstan_version))
+=======
             clean_all(verbose)
             print('Rebuilding version {}'.format(cmdstan_version))
         build(verbose)
         print('Test model compilation')
         compile_example()
     print('Installed {}'.format(cmdstan_version))
+>>>>>>> 84f97454020f633a63843a634e0508e88d837ce2
```
WardBrian, line 335: Looks like a merge snuck into a commit

mitzimorris:
perhaps sampler progress bar should be implemented as before - via custom callback, e.g. like `utils.py` `wrap_url_progress_hook` - will experiment.

mitzimorris:
this is ready for review.

WardBrian:
I’ll look at this first thing tomorrow

mitzimorris:
tqdm is a total PITA and the documentation is sketchy.  we probably want to supress the warning that shows up in the jupyternotebooks about autonotebook.  however, I think we should use autonotebook because it keeps our code simpler.

cmdstanpy/install_cmdstan.py
```diff
@@ -337,12 +303,25 @@ def retrieve_version(version: str, progress: bool = True) -> None:
...
     try:
+        print('Extracting distribution')
         tar = tarfile.open(file_tmp)
         target = os.getcwd()
         if platform.system() == 'Windows':
             # fixes long-path limitation on Windows
             target = r'\\?\{}'.format(target)
-        tar.extractall(target)
+
+        if progress:
```
WardBrian, line 313: This currently does nothing for me

cmdstanpy/install_cmdstan.py
```diff
@@ -306,7 +272,7 @@ def retrieve_version(version: str, progress: bool = True) -> None:
             if progress:
                 progress_hook: Optional[
                     Callable[[int, int, int], None]
-                ] = wrap_progress_hook()
+                ] = wrap_url_progress_hook()
```
WardBrian, line 275: This currently does nothing for me in console or jupyter

cmdstanpy/model.py
```diff
@@ -1231,132 +1225,115 @@ def variational(
...
-            runset._set_retcode(idx, proc.returncode)
+            while proc.poll() is None:
+                if proc.stdout is not None:
+                    line = proc.stdout.readline()
+                    fd_out.write(line)
+                    line = line.strip()
+                    if show_console and len(line) > 1:
+                        print('chain {}: {}'.format(idx + 1, line))
+                    if show_progress and progress_hook is not None:
+                        progress_hook(line)
```
WardBrian, line 1275: This might be a good place for the rarely-used `while... else` statement so we can update the progress hooks at least once if the model completes super quickly (I think this is why my above output is strange on the bernoulli example). 

Was fixing this the main reason to switch to a callback? I for one think the class idea was cleaner, if both can get us what we want

cmdstanpy/stanfit.py
```diff
@@ -263,7 +254,11 @@ def stdout_files(self) -> List[str]:
     @property
     def stderr_files(self) -> List[str]:
         """List of paths to CmdStan stderr transcripts."""
-        return self._stderr_files
+        get_logger().warning(
```
WardBrian, line 257: Does this deprecation imply a release before 1.0, or will be be carrying a deprecation into 1.0?

cmdstanpy/stanfit.py
```diff
@@ -939,10 +923,9 @@ def diagnose(self) -> Optional[str]:
         """
         cmd_path = os.path.join(cmdstan_path(), 'bin', 'diagnose' + EXTENSION)
         cmd = [cmd_path] + self.runset.csv_files
-        result = do_command(cmd=cmd)
-        if result:
-            get_logger().info(result)
-        return result
+        result = StringIO()
+        do_command(cmd=cmd, show_console=True, fd_out=result)
+        return result.getvalue()
```
WardBrian, line 928: I guess this always was the behavior, but printing out and returning seems strange to me. In repls/jupyter this leads to duplicated output

test/test_install_cmdstan.py
```diff
@@ -37,6 +37,8 @@ def test_retrieve_version(self):
         with self.assertRaises(ValueError):
             retrieve_version('')
 
+    from unittest import mock
```
WardBrian, line 40: If we're adding this import we can change line 42 to just `@mock...`, but I'd just as soon not import this directly

cmdstanpy/install_cmdstan.py
```diff
@@ -161,23 +135,12 @@ def compile_example() -> None:
...
-    if proc.returncode:
-        msgs = ['Failed to compile example model bernoulli.stan']
-        if stderr:
-            msgs.append(stderr.decode('utf-8').strip())
-        raise CmdStanInstallError('\n'.join(msgs))
+    try:
+        do_command(cmd=cmd, show_console=verbose)
+    except RuntimeError as e:
+        raise CmdStanInstallError(
+            'Command "make clean-all" failed\n%s' % str(e)
```
WardBrian, line 142: While we did disable the pylint warning, I do think we should try to use f-strings going forward for this type of thing
```suggestion
           f'Command "make clean-all" failed\n{e}'
```

WardBrian (changes requested):
This is really great. A lot of comments, but hopefully most are quick fixes

cmdstanpy/stanfit.py
mitzimorris: agreed.  changing the return type would make sense - we could change it to bool for problems/no problems.
changing return type of a method seems like bad behavoir, but that's why we're still in beta.

test/test_install_cmdstan.py
mitzimorris: it was @mock before but pylint complained.

cmdstanpy/stanfit.py
WardBrian: I think if I had to pick one behavior over the other I'd pick returning the string, and removing the side effect

test/test_install_cmdstan.py
WardBrian: Odd. We definitely shouldn't need the @unittest. if we're importing mock directly, so we can either use @unittest.mock without the import, or @mock with the import

cmdstanpy/stanfit.py
mitzimorris: good question.  we should just get rid of them - it's not anything that's exposed to the end user.

mitzimorris: since all console outputs go to stdout file, it could be renamed -  "msgs", "console_out" ????

WardBrian: `out_stream_msgs`? If it's primarily an internal function I don't mind being a little verbose

test/test_install_cmdstan.py
WardBrian: These poor github users must get pinged a lot.

cmdstanpy/model.py
mitzimorris: yes, this seemed to fix, but evidently not entirely - will go back to SamplerProgress class and try to get it right.

cmdstanpy/model.py
```diff
@@ -1231,132 +1225,115 @@ def variational(
...
+                if line.startswith("Iteration"):
+                    if 'Sampling' in line:
+                        pbar.colour = 'blue'
+                    pbar.update(1)
+                    pbar.postfix[0]["value"] = line
+                elif line.startswith("Elapsed"):
+                    pbar.close()
+
+        except (ImportError, ModuleNotFoundError):
+            print("tqdm was not downloaded, progressbar not shown")
```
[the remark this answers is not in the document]
mitzimorris: agreed - will try again.

cmdstanpy/model.py
```diff
@@ -564,7 +579,8 @@ def sample(
         sig_figs: Optional[int] = None,
         save_latent_dynamics: bool = False,
         save_profile: bool = False,
-        show_progress: Union[bool, str] = False,
+        show_progress: bool = False,
+        show_console: bool = False,
```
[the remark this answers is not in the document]
mitzimorris: show_console yes, show_progress no - optimize and variational are there to provide fast approximations.

cmdstanpy/model.py
```diff
@@ -331,9 +339,16 @@ def compile(
...
-                        msg = do_command(cmd, cmdstan_path())
-                        if msg is not None and 'Warning or error:' in msg:
-                            msg = msg.split("Warning or error:", 1)[1].strip()
+                        tmp = io.StringIO()
+                        do_command(
+                            cmd=cmd,
+                            cwd=cmdstan_path(),
+                            show_console=True,
+                            fd_out=tmp,
+                        )
```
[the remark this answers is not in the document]
mitzimorris: agreed, will fix.

mitzimorris:
thanks for the feedback - will try to get to this tonight.

cmdstanpy/model.py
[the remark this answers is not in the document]
WardBrian: Agree, that was my original intent in asking -- show_console only.

cmdstanpy/stanfit.py
ahartikainen: Yes, I think returning string is a better idea?

cmdstanpy/utils.py
```diff
@@ -925,56 +925,77 @@ def read_rdump_metric(path: str) -> List[int]:
...
+    get_logger().debug('cmd: %s', ' '.join(cmd))
     try:
         proc = subprocess.Popen(
             cmd,
             cwd=cwd,
+            bufsize=1,
             stdin=subprocess.DEVNULL,
             stdout=subprocess.PIPE,
-            stderr=subprocess.PIPE,
+            stderr=subprocess.STDOUT,  # avoid buffer overflow
```
ahartikainen, line 951: So what happened with the buffer overflow? This was due to C++ error output?

ahartikainen: Or could we just check both pipes when doing proc.poll ?

cmdstanpy/stanfit.py
ahartikainen, line 928:
```suggestion
        with StringIO() as fd_out:
            do_command(cmd=cmd, show_console=True, fd_out=fd_out)
            result = fd_out.getvalue()
        return result
```

ahartikainen (commented):
What if we tried to use `flynt` (once only needed?) to fix our fstrings?

https://pypi.org/project/flynt/

WardBrian:
> What if we tried to use `flynt` (once only needed?) to fix our fstrings?
> 
> https://pypi.org/project/flynt/

See some comments on #460 - flynt doesn't play super well with line limits and fails with mixed " and ' in the source, so it ends up leaving behind a lot of old-style strings. I tried it last week with very little success, but it would be great to move everything over.

cmdstanpy/utils.py
mitzimorris: C++ compiler error novels seems most likely.  Sending stderr to stdout greatly simplifies the logic and as we end of parsing both to get assorted error messages, there's no reason not to combine them.

test/test_install_cmdstan.py
ahartikainen: fixed the @ calls :)

ahartikainen:
Ok, no for flynt, let see it again in 6mo

cmdstanpy_tutorial.ipynb
```diff
@@ -40,158 +41,219 @@
...
+    "version of CmdStan *newer* than 2.27.0, but not the latest,\n",
+    "you can install it in the standard conda way by specifying\n",
+    "``cmdstan==VERSION`` in the install command.\n",
+    "\n",
+    "#### Pip (non-Conda) users \n",
+    "\n",
+    "CmdStan can also be installed from PyPI via URL: https://pypi.org/project/cmdstanpy/ or from the\n",
+    "command line using ``pip``:\n",
+    "\n",
+    ".. code-block:: bash\n",
```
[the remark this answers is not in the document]
mitzimorris: fixked

cmdstanpy/model.py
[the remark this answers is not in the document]
mitzimorris: the progress bar callback hook is much more responsive using a reporter class, FWIW.
I've tweaked/hacked this so that:

a) the progress bar remains after sampling has completed
b) in the terminal window, messages written to stdout expose intermediate versions of the progress bar (at least they do on my machine).  to hack this, I write out chains worth of padded blank lines, which leaves space between commands, but also, leaves things looking OK everywhere.

also, can't get chains to display in oder - what's there is as good as I can get it.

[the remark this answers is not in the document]
WardBrian: This seems much much better for me now in testing. Order was a stretch, especially with doing everything threaded.  Thanks!

WardBrian:
Looks much better to me now (though I preferred the yellow/blue as opposed to the two shades of blue). It's so slick that I'm wondering if we should enable it by default. That might mean we would have to actually require tqdm (which would make all the import-checking unnecessary), though. 

Ping me when you want me to re-review

cmdstanpy/model.py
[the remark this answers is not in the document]
mitzimorris: added this to all methods

cmdstanpy/model.py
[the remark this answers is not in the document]
mitzimorris: fixed.

cmdstanpy/install_cmdstan.py
```diff
@@ -107,25 +94,12 @@ def build(verbose: bool = False) -> None:
...
-        msgs = ['Command "make build" failed']
-        if stderr:
-            msgs.append(stderr.decode('utf-8').strip())
-        raise CmdStanInstallError('\n'.join(msgs))
+    try:
+        do_command(cmd=cmd, show_console=verbose)
+    except RuntimeError as e:
+        raise CmdStanInstallError(
+            'Command "make build" failed\n%s' % str(e)
+        ) from None
```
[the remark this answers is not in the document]
mitzimorris: done - disabled pylint "raise-missing-from" instead.

cmdstanpy/install_cmdstan.py
```diff
@@ -306,7 +272,7 @@ def retrieve_version(version: str, progress: bool = True) -> None:
             if progress:
```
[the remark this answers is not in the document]
mitzimorris: in theory yes, but it's been this way for a very long time, so not changing.

cmdstanpy/install_cmdstan.py
mitzimorris: what do you mean?  how does Windows handle unpacking a tgz file?

WardBrian: I'm running on ubuntu at the moment and 

```python
>>> import cmdstanpy
>>> cmdstanpy.install_cmdstan("2.26.0", progress=True)
```

displays no progress bars during download or unpacking

cmdstanpy/install_cmdstan.py
mitzimorris: this works for me when I invoke the script itself from the terminal:
```
python3 cmdstanpy/install_cmdstan.py --overwrite --progress
```
but it doesn't work when it's invoked via utils function `install_cmdstan`.  not sure what the problem is.

WardBrian: My guess is TQDM detects if it is being redirected/piped and disables itself. Probably finally biting us that install_cmdstan the function really just uses subprocess to call install_cmdstan the script. With the recent refactor it shouldn't be too hard to change this if we wanted to, we could just write a main function which takes in a dict and call it directly from the code. For the script, we could just call `main(argparse.parse(sys.argv))` or similar

WardBrian: I think this is worthwhile to fix but I can do that in a later PR if that's easier

WardBrian: Yeah it would actually be pretty simple, we do more or less the same thing to call the install_cxx_toolchain script from inside install_cmdstan. We do it by overwriting sys.argv though, which is a bit of a hack. It would be better to simply make `main()` take in the list and then we can call it in the `if __name__ == "__main__"` block with `sys.argv[1:]`

mitzimorris: removed option.

cmdstanpy/install_cmdstan.py
mitzimorris: it works on a Mac (Catalina),  but only after:
 - installing develop version as latest version for system library
 - calling the python script from the comnand line.

but if this is too limited, OK with getting rid of it.

mitzimorris:
ready for re-review.

mitzimorris:
@WardBrian - making progress bars on by default sounds good - if nothiing else, we'll quickly find out which platforms are problematic.

also, regarding #462, maybe for install, verbose  should be on by default.
I got pretty radical last night trying to simplify behavior of `install_cmdstan'.
will folks be using the conda installer more than this script?   how verbose is conda installer?

WardBrian:
I'm not sure how often people will be using conda or this comparatively. CmdStan has 1164 downloads on Conda since we got it on there, but I have no comparison point for that number. As for verbosity, the compilation happens on the Conda Forge servers, not on the end user machine, so it's pretty minimal in that regard.

cmdstanpy/install_cmdstan.py
WardBrian: I was referring to the progress bar not the tar unpacking, may have been unclear. I think this is covered by #462

cmdstanpy/model.py
```diff
@@ -330,29 +340,37 @@ def compile(
...
+                        finally:
+                            console = sys_stdout.getvalue()
+
+                    if '--warn-pedantic' in console:  # warn, don't fail
+                        lines = console.split('\n')
+                        warnings = [x for x in lines if x.startswith('Warning')]
+                        get_logger().warning('\n'.join(warnings))
+
+                    if 'Syntax error' in console:
+                        get_logger().warning(console)
```
WardBrian, line 360: I'm concerned these conditionals are overfitting the warnings stanc3 emits. We emit warnings even without `--warn-pedantic`, and there are compile errors other than syntax-based ones (like semantic errors)

cmdstanpy/model.py
```diff
@@ -856,58 +865,56 @@ def sample(
...
+                    # pylint: disable=unused-import
+                    from tqdm.auto import tqdm  # type: ignore # noqa
+                except ImportError:
+                    get_logger().warning(
+                        'Package tqdm not installed, cannot show progress '
+                        'information. Please install tqdm with '
+                        "'pip install tqdm'"
+                    )
+                    show_progress = False
+                    show_console = True
```
WardBrian, line 895: I don't think turning on show_console if show_progress fails is a good idea, we should just emit the warning (I might either want progress bars or no output, which this currently doesn't allow).

It would be nice if we had the makevar equivalent of ?= for this, but currently this would even overwrite if I passed `show_console=False`

cmdstanpy/model.py
```diff
@@ -1231,132 +1248,117 @@ def variational(
...
-                pbar.close()
-
-        except Exception as e:  # pylint: disable=broad-except
-            get_logger().warning(
-                'Chain %s: Failed to read the progress on the fly. Error: %s',
-                idx,
-                repr(e),
+            from tqdm.auto import tqdm  # type:ignore
+
+            pbar: Any = tqdm(
```
WardBrian, line 1343:
```suggestion
            pbar: tqdm = tqdm(
```

I don't think this needs to be Any, we can give it the class annotation

cmdstanpy/stanfit.py
```diff
@@ -295,37 +286,26 @@ def get_err_msgs(self) -> str:
...
-                    and os.path.exists(self._stdout_files[i])
-                    and os.stat(self._stdout_files[i]).st_size > 0
-                ):
+                if self._args.method == Method.SAMPLE:
                     with open(self._stdout_files[i], 'r') as fd:
                         contents = fd.read()
                         # pattern matches initial "Exception" or "Error" msg
                         pat = re.compile(r'^E[rx].*$', re.M)
                         errors = re.findall(pat, contents)
+                        print(errors)
```
WardBrian, line 298: Is this a debug print that snuck in?

cmdstanpy/utils.py
```diff
@@ -1217,14 +1232,17 @@ def pushd(new_dir: str) -> Iterator[None]:
...
 
-def wrap_progress_hook() -> Optional[Callable[[int, int, int], None]]:
+def wrap_url_progress_hook() -> Optional[Callable[[int, int, int], None]]:
+    """Sets up tqdm callback for url downloads."""
     try:
-        from tqdm import tqdm
+        from tqdm.auto import tqdm  # type:ignore
 
-        pbar = tqdm(
+        pbar: Any = tqdm(
```
WardBrian, line 1240:
```suggestion
        pbar: tqdm = tqdm(
```

Same as above

docsrc/examples/MCMC Sampling.ipynb
```diff
@@ -151,62 +152,105 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "fit = model.sample(data=data_file, iter_warmup=100000, iter_sampling=100000, show_progress=True)\n"
+   ]
```
WardBrian, line 198: Note that we need to add tqdm to `docsrc/env.yml`. Even then, I don't think the rendered version of this notebook will contain the progress bars

docsrc/examples/MCMC Sampling.ipynb
```diff
@@ -217,9 +261,14 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
    "outputs": [],
-   "metadata": {}
+   "source": []
   }
```
WardBrian, line 272: Can we delete this new blank cell?

cmdstanpy_tutorial.ipynb
```diff
@@ -324,23 +621,89 @@
...
+    "print(diags)"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
    "outputs": [],
-   "metadata": {}
+   "source": []
```
WardBrian, line 696: We should delete this trailing blank cell

cmdstanpy/install_cmdstan.py
```diff
@@ -161,68 +129,53 @@ def compile_example() -> None:
...
     except ValueError as e:
         raise CmdStanInstallError(
             "Failed to rebuild CmdStan. Are you sure it is installed?"
         ) from e
 
 
-def install_version(
-    cmdstan_version: str, overwrite: bool = False, verbose: bool = False
-) -> None:
+def install_version(cmdstan_version: str, overwrite: bool = False) -> None:
```
WardBrian, line 157: So is verbose gone or is verbose on by default? I do not think it should be on by default

test/test_install_cmdstan.py
WardBrian:
```suggestion

```

This import is not necessary for me to pass pylint locally (in fact it's highlighted as unused)

WardBrian (changes requested):
A few more comments.

General note: If we want progress bars by default, we should move tqdm to requirements.txt, add it to setup.py, update the documentation, and set show_progress/progress to `True` in the function calls.

WardBrian:
I forgot to highlight them in my above review but I also think we should avoid introducing new `%` or `.format` style strings. There are a few in the diff here

cmdstanpy/model.py
mitzimorris: is there a way to run the Stanc3 parser to see what error messages are available?
and shouldn't this be documented somewhere - user's guide?

WardBrian: Not sure I follow your question entirely, but stanc3 outputs 3 kinds of error intentionally (see [this file](https://github.com/stan-dev/stanc3/blob/master/src/middle/Errors.ml) if you don't mind a little ocaml):

1. Syntax errors from either the lexer or parser, which begin "Syntax error"
2. Semantic (type) errors, which begin "Semantic error"
3. A FileNotFound error (I think we do our own check that the file exists, so this shouldn't happen?)

There are also some  internal compiler errors which should never leak into userspace, but could. They usually begin "This should never happen. Please file a bug"

WardBrian: If we want to put it somewhere in the doc, I think the cmdstan guide page on `stanc` would be good, underneath the command line arguments.

Also, for warnings, we mostly emit deprecation warnings if warn-pedantic is not enabled, but there are a few other warnings that get emitted. I know there is a warning about int division with `/` instead of `%/%` that is outside pedantic mode, and another warning for if your program is entirely empty.

WardBrian:
This might be silly, but could we approximate a loading bar for actually building cmdstan by counting the number of lines we expect (something like `make build | wc -l` on a reference machine) and then counting up to that number? It might vary a bit from version to version but I think we'd be able to get pretty close, and we always know the final line is something like ' ---- cmdstan built ---', right?

mitzimorris:
we can definitely create a callback hook that would do this.  lovely idea.

cmdstanpy/model.py
mitzimorris: ran a few of the models in Stanc3 repo dir `test/integration/bad/` - looks like semantic error messages are just like syntactic error messages, only they start with "Semantic error".  will check for this as well.

cmdstanpy/install_cmdstan.py
mitzimorris: I just reverted everything back to the way it was - verbose is off by default.

cmdstanpy/model.py
mitzimorris: got rid of line 895

cmdstanpy/model.py
mitzimorris: revamped logic, based on current Stanc3 behavoir (stanc3/test/integration/bad)

cmdstanpy/stanfit.py
mitzimorris: good catch - gone now.

cmdstanpy/utils.py
mitzimorris: fixed.

cmdstanpy/model.py
mitzimorris: fixed

WardBrian:
Great! On my machine:

```
$ make build 2>&1 | wc -l
141
```

We could probably just count up to 150 to account for any variance here?

docsrc/examples/MCMC Sampling.ipynb
mitzimorris: question:  why does `docsrc/env.yml` list `pip` twice?

docsrc/examples/MCMC Sampling.ipynb
mitzimorris: gone

cmdstanpy_tutorial.ipynb
mitzimorris: done

docsrc/examples/MCMC Sampling.ipynb
WardBrian: The first `pip` actually installs pip itself, the second starts a list of packages to be installed via pip. The list we use is basically equivalent to running `pip install -e .` in the cmdstanpy root directory

mitzimorris:
changes made per comments:

- added `tqdm` as requirement
- for sampling, `show_progress=True` by default.
- created progress bar for `make build` - on my machine, the first operation takes 20 seconds, so at first, not very reassuring.  then it starts rolling.

cmdstanpy/install_cmdstan.py
```diff
@@ -10,17 +10,16 @@
    -v, --version <release> : version, defaults to latest release version
    -d, --dir <path> : install directory, defaults to '$HOME/.cmdstan
    --overwrite: flag, when specified re-installs existing version
-   --verbose: flag, when specified prints output from CmdStan build process
```
WardBrian, line 13: Should readd this to docstring

cmdstanpy/install_cmdstan.py
```diff
@@ -352,10 +385,48 @@ def retrieve_version(version: str, progress: bool = True) -> None:
...
+            help="flag, add C++ compiler to path (Windows only)",
+        )
+    return vars(parser.parse_args(sys.argv[1:]))
+
+
 def main(args: Dict[str, Any]) -> None:
     """Main."""
 
     version = latest_version()
+    print(version)
```
WardBrian, line 429:
```suggestion

```

cmdstanpy/model.py
```diff
@@ -330,30 +337,54 @@ def compile(
...
+                        syntax_errors = [
+                            x for x in lines if x.startswith('Syntax error')
+                        ]
+                        semantic_errors = [
+                            x for x in lines if x.startswith('Semantic error')
+                        ]
+                        exceptions = [
+                            x
+                            for x in lines
+                            if x.startswith('Uncauth exception')
```
WardBrian, line 362:
```suggestion
                            if x.startswith('Uncaught exception')
```

We might also want to add messages that start with "Exception:" - this is what an internal ocaml exception would start with.

cmdstanpy/model.py
```diff
@@ -330,30 +337,54 @@ def compile(
...
+                            or len(exceptions) > 0
+                        ):
+                            get_logger().warning(
+                                'Stan program failed to compile:'
+                            )
+                            get_logger().warning(console)
+                        elif len(warnings) > 0:
                             get_logger().warning(
-                                "stanc3 has produced warnings:\n%s", msg
+                                '%d from Stan compiler:', len(warnings)
```
WardBrian, line 375: Reword?

cmdstanpy/model.py
```diff
@@ -330,30 +337,54 @@ def compile(
...
+                        ]
+                        if (
+                            len(syntax_errors) > 0
+                            or len(semantic_errors) > 0
+                            or len(exceptions) > 0
+                        ):
+                            get_logger().warning(
+                                'Stan program failed to compile:'
+                            )
+                            get_logger().warning(console)
```
WardBrian, line 372:
```suggestion
                            get_logger().error(console)
```

cmdstanpy/utils.py
```diff
@@ -121,11 +122,12 @@ def validate_cmdstan_path(path: str) -> None:
     Throws exception if specified path is invalid.
     """
     if not os.path.isdir(path):
-        raise ValueError('no such CmdStan directory {}'.format(path))
+        raise ValueError(
+            'No CmdStan directory, ' 'path {} does not exist.'.format(path)
```
WardBrian, line 126: Why are the strings in this file split like this?

cmdstanpy_tutorial.ipynb
```diff
@@ -40,158 +41,302 @@
...
+    "#### Pip (non-Conda) users \n",
+    "\n",
+    "CmdStan can also be installed from PyPI via URL: https://pypi.org/project/cmdstanpy/ or from the\n",
+    "command line using `pip`:\n",
+    "\n",
+    "```\n",
+    "pip install --upgrade cmdstanpy\n",
+    "```\n",
+    "\n",
+    "The optional packages are\n",
```
WardBrian, line 98: Update to reflect TQDM not being optional

docsrc/hello_world.rst
```diff
@@ -184,44 +184,36 @@ to a specified directory.
...
-  
-      pip install ipywidgets
-      jupyter nbextension enable --py widgetsnbextension
-  
-  Jupyter Lab still needs widgets manager.
-  
-  .. code-block:: bash
-  
-      jupyter labextension install @jupyter-widgets/jupyterlab-manager
+Users can enable progress bar for the sampling if the ``tqdm`` package
```
WardBrian, line 190: Update to reflect required nature of TQDM

WardBrian (commented):
Truly just nitpicky stuff now. The progress bar on the make command looks great, I think that's a huge win for users.

cmdstanpy/install_cmdstan.py
```diff
@@ -150,51 +143,77 @@ def build(verbose: bool = False) -> None:
...
+        if verbose:
+            do_command(cmd)
+        else:
+            do_command(cmd, fd_out=io.StringIO())
+    except RuntimeError as e:
+        # pylint: disable=raise-missing-from
+        raise CmdStanInstallError(f'Command "make clean-all" failed\n{e}')
+
+
+def rebuild_cmdstan(cmdstan_version: str, verbose: bool = False) -> None:
```
WardBrian, line 203: Sorry, missed this on first review - I think requiring the version argument to rebuild is pretty un-ergonomic (users might not know the exact version, and supplying the wrong version doesn't do anything besides break the progress bar, right?)

cmdstanpy/install_cmdstan.py
```diff
@@ -150,51 +143,77 @@ def build(verbose: bool = False) -> None:
...
+        bar_format="{desc} ({elapsed}) | {bar} | {postfix[0][value]}",
+        postfix=[dict(value=f'Building {cmdstan_version} {pad}')],
+        colour='blue',
+        desc='',
+        position=0,
+    )
+    success = f'--- CmdStan {cmdstan_version} built ---'
+
+    def build_progress_hook(line: str) -> None:
+        if line == '__DONE__' or line == success:
```
WardBrian, line 163: Could we do this with a simple regex instead to make this function version-agnostic and resolve the rebuild issue? We lose the ability to postfix the version we're building but I think its fine if the bar just says "Building Cmdstan" instead of "Building 2.27.0"

WardBrian (changes requested):

cmdstanpy/install_cmdstan.py
WardBrian: Also, I think `rebuild` should take a progress argument now and pass it along. It can default to True

cmdstanpy/install_cmdstan.py
WardBrian: Alt: We can refactor `utils.cmdstan_version_at` into two functions, one which parses the makefile to get the version and the other that has the current functionality, and then use the first in `build`. I think I prefer the regex/fuzzy solution, but either seems fine to me.

cmdstanpy/install_cmdstan.py
mitzimorris: added to docstring, and further edits to better explain options and defaults.

cmdstanpy/utils.py
mitzimorris: I blame 'black'.  or gremlins.  fixed.

cmdstanpy/model.py
mitzimorris: much better - thanks!

docsrc/hello_world.rst
mitzimorris: updated.  commented out section on Jupyter notebook add'l installs - not sure if these are needed anymore, as linked issue is closed.
how to test?

cmdstanpy/install_cmdstan.py
mitzimorris: I simplified the condition to eliminate the need for the regex, also, changed "do_command" to send final line of stdout to pbar (as needed) and it seems to just work and is in fact, the right thing to do.

cmdstanpy/install_cmdstan.py
mitzimorris: thanks for this - completely agree - it was a hack and now it's better.

cmdstanpy/utils.py
```diff
@@ -157,8 +157,7 @@ def cmdstan_path() -> str:
             cmdstan_dir = os.path.expanduser(os.path.join('~', _DOT_CMDSTANPY))
             if not os.path.exists(cmdstan_dir):
                 raise ValueError(
-                    'no CmdStan installation found, '
-                    'run command line script "install_cmdstan"'
+                    'No CmdStan installation found, ' 'run "install_cmdstan".'
```
WardBrian, line 160:
```suggestion
                    'No CmdStan installation found, run "install_cmdstan".'
```

cmdstanpy/utils.py
```diff
@@ -168,8 +167,7 @@ def cmdstan_path() -> str:
         latest_cmdstan = get_latest_cmdstan(cmdstan_dir)
         if latest_cmdstan is None:
             raise ValueError(
-                'no CmdStan installation found, '
-                'run command line script "install_cmdstan"'
+                'No CmdStan installation found, ' 'run "install_cmdstan".'
```
WardBrian, line 170:
```suggestion
                'No CmdStan installation found, run "install_cmdstan".'
```

WardBrian (commented):
Found two last gremin changes, then I think we can approve!

WardBrian (approved):
LGTM assuming tests pass!

docsrc/hello_world.rst
mitzimorris: some amount of setup is going to be required in order to make this fly in Jupyter notebooks.  not sure what or how much.

WardBrian:
So, if you try to run a progress bar in notebook without ipywidgets, you get

```
ImportError: IProgress not found. Please update jupyter and ipywidgets. See https://ipywidgets.readthedocs.io/en/stable/user_install.html
```

Which is not a terrible error message. But, the problem is we don't report that, we report `RuntimeError: Error during sampling: `, which is pretty bad.


Now, here's what I propose we do:

in `__init__.py`:

```python
_SHOW_PROGRESS = True

def _disable_progress(e: Exception):
    global _SHOW_PROGRESS
    if _SHOW_PROGRESS:
        _SHOW_PROGRESS = False
        get_logger().error(
            "Error in progress bar initialization:\n"
            f"\t{e}\n"
            "Disabling progress bars for this session"
        )

```

Then, at each call site of TQDM, we do something like this (shown for model.py, the same would need to be done in the url hook and the build and unzipping):

```python

    # pylint: disable=no-self-use
    def _wrap_sampler_progress_hook(
        self, chain_id: int, total: int
    ) -> Optional[Callable[[str], None]]:
        """Sets up tqdm callback for CmdStan sampler console msgs."""
        try:
            pbar: Any = tqdm(
                total=total,
                bar_format="{desc} |{bar}| {elapsed} {postfix[0][value]}",
                postfix=[dict(value="Status")],
                desc=f'chain {chain_id}',
                colour='yellow',
            )
        except Exception as e:
            _disable_progress(e)

            def sampler_progress_hook(_: str) -> None:
                return

        else:

            def sampler_progress_hook(line: str) -> None:
                if line == "Done":
                    pbar.postfix[0]["value"] = 'Sampling completed'
                    pbar.close()
                elif line.startswith("Iteration"):
                    if 'Sampling' in line:
                        pbar.colour = 'blue'
                    pbar.update(1)
                    pbar.postfix[0]["value"] = line

        return sampler_progress_hook
```


----

This makes the error non-fatal, shows the correct message, **and** only shows it once per session.

Screenshot from 2021-10-05 09-53-35

If we wanted, we could also replace all our `if show_progress` statements with `if show_progress and _SHOW_PROGRESS` to completely skip re-running this code, but that's not necessary - it will safely return an empty callback every time without that.

How does this sound @mitzimorris?

cmdstanpy/model.py
```diff
@@ -1341,22 +1355,37 @@ def _wrap_sampler_progress_hook(
...
+                        pbar.postfix[0]["value"] = 'Sampling completed'
+                        pbar.update(total - pbar.n)
+                        pbar.close()
+                    elif line.startswith("Iteration"):
+                        if 'Sampling' in line:
+                            pbar.colour = 'blue'
+                            pbar.update(1)
+                            pbar.postfix[0]["value"] = line
+                # pylint: disable=broad-except
+                except Exception as e:
```
WardBrian, line 1388: I don’t think it’s possible for this code to throw an exception due to user configuration, only due to a bug in our usage, so I don’t think trapping it is a good idea

cmdstanpy/model.py
```diff
@@ -894,6 +902,12 @@ def sample(
                 refresh = _CMDSTAN_REFRESH
             iter_total = iter_total // refresh + 2
 
+            if show_progress:
```
WardBrian, line 905: I think we can simplify this to 
`show_progress = show_progress and _SHOW_PROGRESS`

Additionally, asserting types at runtime is usually frowned upon if it can be avoided. I think adding a type annotation in init.py can avoid this

cmdstanpy/model.py
```diff
@@ -755,24 +762,25 @@ def sample(
...
-        :param show_progress: If True, display progress bar to track
-            progress for warmup and sampling iterations.  Default is ``True``.
+        :param show_progress: If ``True``, display progress bar to track
+            progress for warmup and sampling iterations.  Default is ``True``,
+            unless package tqdm progress bar encounter errors.
 
-        :param show_console: If True, stream CmdStan messages sent to stdout
-            and stderr to the console.  Default is False.
+        :param show_console: If ``True``, stream CmdStan messages sent to stdout
+            and stderr to the console.  Default is False.  When
```
WardBrian, line 783:
```suggestion
            and stderr to the console.  Default is False. 
```

cmdstanpy_tutorial.py
```diff
@@ -29,32 +32,34 @@
...
 
 # #### Run sampler diagnostics
 
-fit.diagnose()
-
-
+print(fit.diagnose())
+
+# #### Visualize posterior density of parameter "theta"
+if sys.version[0:3] != '3.8':  # https://github.com/conda/conda/issues/9589
```
WardBrian, line 58: I don’t understand the linked issue as a reason for this check? And I don’t think such a thing belongs in the tutorial code

cmdstanpy/utils.py
```diff
@@ -1211,24 +1225,40 @@ def install_cmdstan(
...
+            try:
+                if pbar.total is None:
+                    pbar.total = total_size
+                    pbar.reset()
+                downloaded_size = count * block_size
+                pbar.update(downloaded_size - pbar.n)
+                if pbar.n >= total_size:
+                    pbar.close()
+            # pylint: disable=broad-except
+            except Exception as e:
```
WardBrian, line 1260: Once again, this try/except seems like a bad idea to me. Try/excepts aren’t free operations and this will only get triggered if there’s an issue in our code. TQDM import errors should all happen during the constructor call

cmdstanpy/utils.py
```diff
@@ -1211,24 +1225,40 @@ def install_cmdstan(
...
-    ) -> None:
-        if pbar.total is None:
-            pbar.total = total_size
-            pbar.reset()
-        downloaded_size = count * block_size
-        pbar.update(downloaded_size - pbar.n)
-        if pbar.n >= total_size:
-            pbar.close()
+    try:
+        pbar: Any = tqdm(
```
WardBrian, line 1229: I think we can make the type of all of these `tqdm` Instead of `Any`

WardBrian (changes requested):
I think we also need to protect the use in the for loop that does tarball unpacking. Luckily we know that will have to happen after the url hooks, so we can assume `_SHOW_PROGRESS` will be false already if there are problems

WardBrian:
I’m not sure that putting the disable function in a different module than the global variable actually has the correct semantics. I believe when you say “global” that means you’re editing the version imported in utils, and I can’t find much reassurance online that it is actually propagating back to the original (or that, if it is, that behavior is guaranteed). 

If you want them in separate modules, I believe you can assign `cmdstanpy._SHOW_PROGRESS = False`
but it is worth testing that this actually works as intended

cmdstanpy/install_cmdstan.py
```diff
@@ -147,28 +149,46 @@ def build(verbose: bool = False, progress: bool = True) -> None:
...
+                        pbar.postfix[0]["value"] = line
+                    else:
+                        pbar.postfix[0][
+                            "value"
+                        ] = f'{line[:8]} ... {line[-20:]}'
+                        pbar.set_description('Compiling')
+                        pbar.update(1)
+
+        # pylint: disable=broad-except
+        except Exception as e:
```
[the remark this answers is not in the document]
mitzimorris: on Windows, in Jupyter, `pbar.close` seems to be crashing things.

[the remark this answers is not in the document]
WardBrian: This particular section of code wrapping a definition in try/except, not the body. In either case, I think it’s a non-starter if we would actually need to have a try/except inside every usage of tqdm rather than just at instantiation. If that would be necessary I think we should go back to the janky soft dependency with it disabled by default

cmdstanpy_tutorial.py
mitzimorris: OK, will remove altogether.

cmdstanpy/utils.py
mitzimorris: it's not an import error - it's that Jupyter notebook doesn't have ipywidgets installed and enabled.  so things blow up.  can this be checked for?

WardBrian: For me, if I try to use TQDM without ipywidgets, the failure occurs exactly when I call `tqdm()` and raises a python `ImportError`. What error are you getting?

WardBrian:
Seems like our `pushd` needs to make folders if they don’t exist

mitzimorris:
> Seems like our `pushd` needs to make folders if they don’t exist

that shouldn't be necessary - args passed to `pushd` are checked by `validate_dir`

mitzimorris:
the new 2.28.0 tarball has top-level directory 'cmdstan' - expecting 'cmdstan-2.28.0' - that's why this is failing.

WardBrian:
Oh, weird. @serban-nicusor-toptal and @rok-cesnovar - was this intentional to change the folder structure of the tarball?

serban-nicusor-toptal:
Unintentional. It's now fixed, can you please give it another go and let me know if it's fine ?

WardBrian:
Seems resolved @serban-nicusor-toptal

mitzimorris:
the code is pretty brittle - will add more checks to make sure the tarfile matches its tag.

WardBrian:
While it’s hard to argue against more stable code, I don’t think it’s unreasonable to make assumptions about tarballs we ourselves are generating. The worst thing that happens is we just fix it on release day

mitzimorris:
agreed, but now that we've seen this problem, catch it with an informative error message - debugging this one was no fun.

WardBrian:
I'm getting some intermittent testing failures on the new show_console tests for my local machine. I think it's due to the fact that we're using `.readline()` in _run_cmdstan, which will only ever return one line. If more than one line was created by the subprocess that loop (which will be system/run dependent) it won't output the latter ones. 

We can instead use `.readlines()` and a simple loop. I believe this is the correct behavior:

```python
            while proc.poll() is None:
                if proc.stdout is not None:
                    lines = proc.stdout.readlines()
                    for line in lines:
                        fd_out.write(line)
                        line = line.strip()
                        if show_console and len(line) > 1:
                            print('chain {}: {}'.format(idx + 1, line))
                        if show_progress and progress_hook is not None:
                            progress_hook(line)
```

test/test_optimize.py
```diff
@@ -142,17 +144,15 @@ def test_rosenbrock(self):
         self.assertAlmostEqual(mle.optimized_params_pd['x'][0], 1, places=3)
         self.assertAlmostEqual(mle.optimized_params_dict['x'], 1, places=3)
 
-        self.assertEqual(
-            mle.stan_variable('x', inc_iterations=True).shape, (36,)
-        )
-        self.assertEqual(mle.optimized_iterations_np.shape, (36, 3))
+        last_iter = mle.optimized_iterations_np.shape[0] - 1
```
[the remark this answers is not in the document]
mitzimorris: since `optimize` function hasn't changed in forever, it should take 36 iterations to find MLE.
but really not sure about what the algorithm does to know whether or not it should always take exactly 36 iterations, given seed and all other initialization.

test/test_variational.py
```diff
@@ -178,32 +180,20 @@ def test_variational_good(self):
             variational.column_names,
             ('lp__', 'log_p__', 'log_g__', 'mu[1]', 'mu[2]'),
         )
-
-        self.assertAlmostEqual(
-            variational.variational_params_np[3], 31.0418, places=2
-        )
+        # not testing values, just shapes
```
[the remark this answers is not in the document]
mitzimorris: I think that what CmdStanPy should test is that it correctly parses all relevant information out of the CSV file and puts it in the right slot with the right shape. 

the Stan and Math lib unit and integration tests should make sure that algorighms produce correct estimates.

does this make sense?

mitzimorris:
YATA!  I can confirm that on my Windows machine, running a jupyter notebook:
a) without ipywidgets, tqdm fail is captured once, and subsequence calls to `sample` run cleanly (sans progress bar)
b) with ipywidgets installed, per github issue instructions, all is good.

the regular unit tests test that the tqdm progress bar displays properly in the terminal window.

I think this still needs:
  - better install instructions for jupyter notebooks progress bar fans
  - better progress bar handling
  - cleanup of unit test hack to do whatever is required to keep state clean

mitzimorris:
> I'm getting some intermittent testing failures on the new show_console tests for my local machine. I think it's due to the fact that we're using `.readline()` in _run_cmdstan, which will only ever return one line. If more than one line was created by the subprocess that loop (which will be system/run dependent) it won't output the latter ones.
> 
> We can instead use `.readlines()` and a simple loop. I believe this is the correct behavior:
> 
> ```python
>             while proc.poll() is None:
>                 if proc.stdout is not None:
>                     lines = proc.stdout.readlines()
>                     for line in lines:
>                         fd_out.write(line)
>                         line = line.strip()
>                         if show_console and len(line) > 1:
>                             print('chain {}: {}'.format(idx + 1, line))
>                         if show_progress and progress_hook is not None:
>                             progress_hook(line)
> ```

I don't think this was the problem.   reverting this change, because calling `readlines` instead of `readline` causes everything to block.  the call to subprocess is configured:

```
proc = subprocess.Popen(
            cmd,
            cwd=cwd,
            bufsize=1,
            stdin=subprocess.DEVNULL,
            stdout=subprocess.PIPE,
            stderr=subprocess.STDOUT,  # avoid buffer overflow
            env=os.environ,
            universal_newlines=True,
        )
```

from the docs:  https://docs.python.org/3/library/subprocess.html

> bufsize will be supplied as the corresponding argument to the open() function when creating the stdin/stdout/stderr pipe file objects:
> 
> - 0 means unbuffered (read and write are one system call and can return short)
> - 1 means line buffered (only usable if universal_newlines=True i.e., in a text mode)
> - any other positive value means use a buffer of approximately that size
> - negative bufsize (the default) means the system default of io.DEFAULT_BUFFER_SIZE will be used.
>

WardBrian:
I can't recreate the test failure I was getting yesterday, so I think you were right @mitzimorris

mitzimorris:
still getting intermittant failures in unittest - this one:
test/test_variational.py::VariationalTest::test_single_row_csv

WardBrian:
Test ordering seems to work for me locally if I use set indexes. We should just set the index of the progress bar test to 0 and the index of the failure test to 9999 or some other arbitrarily high value and

mitzimorris:
I set the index to -1, which should mean test last.

docsrc/hello_world.rst
```diff
@@ -184,44 +184,48 @@ to a specified directory.
...
+Jupyter Lab Notebook requirements
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   
-      jupyter labextension install @jupyter-widgets/jupyterlab-manager
+In a Jupyter notebook, this package requires the `ipywidgets <https://ipywidgets.readthedocs.io/en/latest/index.html>`_ package.
+You must enable the ipywidgets notebook extension via command:
+
+.. code-block:: bash
+
+    jupyter nbextension enable --py widgetsnbextension
```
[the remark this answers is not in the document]
WardBrian: Are you using 'jupyter lab'? For me on windows and ubuntu it was enough to just use pip. I think this kind of advice goes stale very quickly, which is why we should just defer to their docs (which contain both that command and a lot of other specific advice depending on user's setups)

mitzimorris:
variational has problems.    I created a new test called "glitch" which calls the variational method on the bernoulli test model and data repeatedly.  not setting seed, mind you - third time through, variational fails to converge.
wtf?


```
    def test_glitch(self):
        stan = os.path.join(DATAFILES_PATH, 'matrix_var.stan')
        model = CmdStanModel(stan_file=stan)
        vb_fit = model.variational(show_console=True)
        self.assertTrue(isinstance(vb_fit.stan_variable('theta'), float))

        vb_fit = model.variational(show_console=True)
        z_as_ndarray = vb_fit.stan_variable(var="z")
        self.assertEqual(z_as_ndarray.shape, (4, 3))

        vb_fit = model.variational(show_console=True)
        vb_fit = model.variational()
        vb_fit = model.variational(show_console=True)
        for i in range(4):
            for j in range(3):
                self.assertEqual(int(z_as_ndarray[i, j]), i + 1)


```
failed on 3rd invocation of `model.variational()`

bob-carpenter:
> failed on 3rd invocation of model.variational()

Are you sure that's not just variational inference failing?  You're using three different seeds there.  I'd suggest setting the seeds all to the same value to see if it's just a problem on the outside.  Or setting them to three different values and then seeing if cmdstanr or CmdStan also have problems.

mitzimorris:
you're right.  same for optimize - will change offset in the RNG.

docsrc/hello_world.rst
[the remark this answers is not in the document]
mitzimorris: you've convinced me.  will do.

mitzimorris:
removed instructions from the doc.  I think this is ready to merge.

WardBrian:
Well nevermind then, I'll put it in the next PR

PR #464, opened 2021-10-04
Allow user to supply a c++ header file

### LINKED ISSUE #463: [FR] Support "USER_HEADER" CmdStan flag
#### Summary:
CmdStan [supports using external .hpp files](https://github.com/stan-dev/docs/blob/master/src/cmdstan-guide/external_code.Rmd) for function definitions. 

We should expose this to the model constructor and `compile` functions, and create an example notebook similar to [this one](https://cran.r-project.org/web/packages/rstan/vignettes/external.html) for RStan.

Rough design:
* New optional file argument `user_header` for `CmdStanModel()` and `CmdStanModel.compile`. 
* This should probably be part of the `CompilerOptions` class
* This should also enable --allow-undefined by default, IMO
* We probably should check the existence of the file before running the make command, because I'm not sure the error message produced otherwise will be helpful

WardBrian:
#### Submission Checklist

- [x] Run unit tests

#### Summary

Adds the ability to specify a .hpp file for external function definitions. Closes #463.

This requires storing some .hpp files in the test data, which overcomplicated the removal routine. I've switched to a global fixture which uses `git clean -Xf data/` to remove all files in the git ignore in the test/data folder. This is only run once per session at the moment, at the end of all testing. 

#### Copyright and Licensing

Simons Foundation

### DIFF
--- a/cmdstanpy/compiler_opts.py
+++ b/cmdstanpy/compiler_opts.py
@@ -48,6 +48,7 @@ class CompilerOptions:
     Attributes:
         stanc_options - stanc compiler flags, options
         cpp_options - makefile options (NAME=value)
+        user_header - path to a user .hpp file to include during compilation
     """
 
     def __init__(
@@ -55,11 +56,13 @@ class CompilerOptions:
         *,
         stanc_options: Optional[Dict[str, Any]] = None,
         cpp_options: Optional[Dict[str, Any]] = None,
+        user_header: Optional[str] = None,
         logger: Optional[logging.Logger] = None,
     ) -> None:
         """Initialize object."""
         self._stanc_options = stanc_options if stanc_options is not None else {}
         self._cpp_options = cpp_options if cpp_options is not None else {}
+        self._user_header = user_header if user_header is not None else ''
         if logger is not None:
             get_logger().warning(
                 "Parameter 'logger' is deprecated."
@@ -67,8 +70,8 @@ class CompilerOptions:
             )
 
     def __repr__(self) -> str:
-        return 'stanc_options={}, cpp_options={}'.format(
-            self._stanc_options, self._cpp_options
+        return 'stanc_options={}, cpp_options={}, user_header={}'.format(
+            self._stanc_options, self._cpp_options, self._user_header
         )
 
     @property
@@ -81,6 +84,11 @@ class CompilerOptions:
         """C++ compiler options."""
         return self._cpp_options
 
+    @property
+    def user_header(self) -> str:
+        """The user header file if it exists, otherwise empty"""
+        return self._user_header
+
     def validate(self) -> None:
         """
         Check compiler args.
@@ -88,6 +96,7 @@ class CompilerOptions:
         """
         self.validate_stanc_opts()
         self.validate_cpp_opts()
+        self.validate_user_header()
 
     def validate_stanc_opts(self) -> None:
         """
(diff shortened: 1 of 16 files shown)


WardBrian:
@mitzimorris this is now failing with 2.28 and wasn’t before, need to investigate tomorrow but very odd

cmdstanpy/compiler_opts.py
```diff
@@ -149,10 +156,37 @@ def validate_cpp_opts(self) -> None:
...
+            if self._user_header[-4:] != '.hpp':
+                raise ValueError(
+                    f"Header file must end in .hpp, got {self._user_header}"
+                )
+            if "allow_undefined" not in self._stanc_options:
+                self._stanc_options["allow_undefined"] = True
+            # set full path
+            self._user_header = os.path.abspath(self._user_header)
+
+            if ' ' in self._user_header:
```
mitzimorris, line 185: is it possible to use TemporaryCopiedFile here?

mitzimorris (commented):
looks good and appreciate the global cleanup.

question for @rok-cesnovar and @jgabry - add similar functionality to CmdStanR?

cmdstanpy/compiler_opts.py
WardBrian: I couldn't think of any way to do so inside CompilerOpts, because the context manager would get destroyed when the function returned. We could use one in the `compile` call, but you end up with a pretty ugly 'if we have a file, use `TemporaryCopiedFile` and do all this stuff, else, do all the exact same stuff' with no real way of de-duplicating it. It also starts to pollute the logic of `compile()` to do some argument validation again

Happy to take suggestions if you can see a way to do it. I figure this way of doing it is safe and simple and we can improve it in the future if need be

mitzimorris: this is an expert-level feature, correct?  in which case, this is OK.

mitzimorris (approved):

mitzimorris:
@WardBrian - should CmdStanPy do the same?  do we need to introduce add'l argument `user_header` ?

> > question for @rok-cesnovar and @jgabry - add similar functionality to CmdStanR?
> 
> CmdStanR supports this already via the cpp_options, stanc_options:

WardBrian:
I checked, and it does indeed work in the example I wrote up to replace

```python
model_external.compile(user_header='make_odds.hpp')
```

with

```python
model_external.compile(cpp_options={'USER_HEADER':'/home/brian/Dev/py/cmdstanpy/docsrc/examples/make_odds.hpp'}, stanc_options={'allow_undefined':True})
```.

While I do think it is a good idea to have this as a separate argument, if only so we can do some path resolution, I'm going to rewrite it slightly so it uses `self._cpp_options` under the hood.

File: docsrc/examples/Using External C++.ipynb [hunk 2]
```diff
@@ -0,0 +1,157 @@
...
+    "This will enables the `allow_undefined` flag automatically."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "model_external.compile(user_header='make_odds.hpp')"
```
mitzimorris, line 92: update this to `cpp_opts`

WardBrian: I'm not sure what you mean. The user interface is still to use `model_header`, this just gets turned into part of the CPP options during validation

mitzimorris: you're right - user error on my part - testing in wrong environment.

PR #468, opened 2021-10-06
Hide away progressbar details

WardBrian:
This is just summing up the conversations Mitzi and I have had about resolving the progress bar issues without making the code look too terrible. 

This is more for reference than anything. You know a PR has gotten out of hand when we need branches off of it.


mitzimorris:
done as part of #461

PR #472, opened 2021-10-09
logic fix for cmdstan_version_at, get_latest_cmdstan

mitzimorris:
#### Submission Checklist

- [x] Run unit tests

#### Summary
Cleanup on logic for managing CmdStan versions:
 - log warning when can't parse CMDSTAN_VERSION from its makefile
 - for release candidates, cleaner name munging

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting (this will be you or your assignee, such as a university or company): Columbia University

### DIFF
--- a/cmdstanpy/utils.py
+++ b/cmdstanpy/utils.py
@@ -84,38 +84,43 @@ def get_latest_cmdstan(cmdstan_dir: str) -> Optional[str]:
     """
     Given a valid directory path, find all installed CmdStan versions
     and return highest (i.e., latest) version number.
-    Assumes directory populated via `install_cmdstan`.
+
+    Assumes directory consists of CmdStan releases, created by
+    function `install_cmdstan`, and therefore dirnames have format
+    "cmdstan-<maj>.<min>.<patch>" or "cmdstan-<maj>.<min>.<patch>-rc<num>",
+    which is CmdStan release practice as of v 2.24.
     """
     versions = [
-        ''.join(name.split('-')[1:])  # name may contain '-rc'
+        name[8:]
         for name in os.listdir(cmdstan_dir)
         if os.path.isdir(os.path.join(cmdstan_dir, name))
         and name.startswith('cmdstan-')
         and name[8].isdigit()
+        and len(name[8:].split('.')) == 3
     ]
-    # munge rc for sort, e.g. 2.25.0-rc1 -> 2.25.0.-99
-    for i in range(len(versions)):  # # pylint: disable=C0200
-        tmp = versions[i].split('rc')
-        if len(tmp) == 1:
-            versions[i] = '.'.join([tmp[0], '0'])
-        else:
-            rc_sortable = str(int(tmp[1]) - 100)
-            versions[i] = '.'.join([tmp[0], rc_sortable])
-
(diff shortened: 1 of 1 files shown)


mitzimorris:
if this passes tests, ready for review.

cmdstanpy/utils.py
```diff
@@ -84,38 +84,43 @@ def get_latest_cmdstan(cmdstan_dir: str) -> Optional[str]:
     """
     Given a valid directory path, find all installed CmdStan versions
     and return highest (i.e., latest) version number.
-    Assumes directory populated via `install_cmdstan`.
+
+    Assumes directory consists of CmdStan releases, created by
+    function `install_cmdstan`, and therefore dirnames have format
+    "cmdstan-<maj>.<min>.<patch>" or "cmdstan-<maj>.<min>.<patch>-rc<num>",
+    which is CmdStan release practice as of v 2.24.
```
WardBrian, line 91: for the record, conda builds do not do this. They end up installed in a directory which has the top level name of 'cmdstan'. We can change this going forward, but I didn't realize we relied on the directory names so existing installations will not reflect this. I don't think this is a big deal because I'm pretty sure get_latest_cmdstan is never called when `$CMDSTAN` is set

cmdstanpy/utils.py
```diff
@@ -177,48 +182,75 @@ def cmdstan_path() -> str:
...
+            major,
+            minor,
+        )
+        return False
+
+    start_idx += len('CMDSTAN_VERSION := ')
+    end_idx = contents.find('\n', start_idx)
+
+    version = contents[start_idx:end_idx]
+    if version is None or len(version) < 3 or len(version.split('.')) < 2:
```
WardBrian, line 239: I don't believe `version` can ever be `None`, it would be `""` or the above line would error out.

test/test_utils.py
```diff
@@ -202,9 +202,30 @@ def test_munge_cmdstan_versions(self):
             os.makedirs(os.path.join(tdir, 'cmdstan-2.22.0'))
             self.assertEqual(get_latest_cmdstan(tdir), 'cmdstan-2.22.0')
 
-    def test_cmdstan_version_at(self):
+    def test_cmdstan_version_before(self):
```
WardBrian, line 205: This isn't really testing the utility, it's testing the failure conditions, so we might want a different name

cmdstanpy/utils.py
```diff
@@ -177,48 +182,75 @@ def cmdstan_path() -> str:
...
+            major,
+            minor,
+        )
+        return False
+
+    splits = version.split('.')
+    cur_major = int(splits[0])
+    cur_minor = int(splits[1])
+    if cur_major < major or (cur_major == major and cur_minor < minor):
+        return True
```
WardBrian, line 253: I think it would be useful (and easier to test) if we made this two functions. One called like `get_cmdstan_version` which does the above parsing and returns a tuple, and the other `cmdstan_version_before` which calls the first and basically consists of just these two lines

cmdstanpy/utils.py
mitzimorris: correct.

WardBrian: great, just wanted to check

WardBrian (approved):

PR #488, opened 2021-11-01
Serialize inf/nan to json

WardBrian:
#### Submission Checklist

- [x] Run unit tests

#### Summary

This changes `write_stan_json` to turn `Nan` into `"NaN"`, `inf` to `"+inf"`, and `-inf` to `"-inf"` as CmdStan expects. It only calls this as necessary, so it shouldn't be too much less efficient than the prior check which threw an error in these cases.

#### Copyright and Licensing

Simons Foundation

### DIFF
--- a/cmdstanpy/utils.py
+++ b/cmdstanpy/utils.py
@@ -412,6 +412,7 @@ def cxx_toolchain_path(
 def rewrite_inf_nan(
     data: Union[float, int, List[Any]]
 ) -> Union[str, int, float, List[Any]]:
+    """Replaces NaN and Infinity with string representations"""
     if isinstance(data, float):
         if math.isnan(data):
             return 'NaN'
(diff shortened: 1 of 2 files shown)


cmdstanpy/utils.py
```diff
@@ -409,6 +409,22 @@ def cxx_toolchain_path(
     return compiler_path, tool_path
 
 
+def rewrite_inf_nan(
```
ahartikainen, line 412: We should also test np.float32 versions if they work as they should.

WardBrian: By putting the call toward the end of `write_stan_json`, it is all Python objects at this point anyway, but I will add a test.

mitzimorris (approved):
lgtm!

ahartikainen (approved):
LGTM

PR #489, opened 2021-11-01
Give more informative error for bad draws (highlight disk space as possible culprit)

### LINKED ISSUE #485: check/warn about filling up _TMPDIR
#### Summary:

Problems reported via Discourse:  https://discourse.mc-stan.org/t/analyzing-the-posterior-prediction-samples/24956/20?u=mitzimorris

#### Description:
User ran model to fit larget dataset and then run posterior predictive checks in the generated quantities block, using all default sampler settings and input dataset of 10M items.

CmdStan run succeeded, CmdStanPy unable to check the outputs when output files written to _TMPDIR - 
fails with error:

```
  File "C:\Users\JORDAN.HOWELL.GITDIR\AppData\Local\Continuum\anaconda3\envs\cmdstan\lib\site-packages\cmdstanpy\utils.py", line 847, in scan_sampling_iters
    raise ValueError(

ValueError: line 51: bad draw, expecting 10799537 items, found 678606
```

When `output_dir` specified, call to sampler via CmdStanPy succeeds.

The result of running the sampler with defaults is 4 Stan CSV files of 1001 rows and 10M columns.
Is this enough data to fill up _TMPDIR?

WardBrian:
Closes #485

#### Submission Checklist

- [x] Run unit tests

#### Summary

Mentions disk space as a possible cause for the rror

#### Copyright and Licensing

Simons Foundation

### DIFF
--- a/cmdstanpy/utils.py
+++ b/cmdstanpy/utils.py
@@ -591,12 +591,15 @@ def scan_sampler_csv(path: str, is_fixed_param: bool = False) -> Dict[str, Any]:
     dict: Dict[str, Any] = {}
     lineno = 0
     with open(path, 'r') as fd:
-        lineno = scan_config(fd, dict, lineno)
-        lineno = scan_column_names(fd, dict, lineno)
-        if not is_fixed_param:
-            lineno = scan_warmup_iters(fd, dict, lineno)
-            lineno = scan_hmc_params(fd, dict, lineno)
-        lineno = scan_sampling_iters(fd, dict, lineno)
+        try:
+            lineno = scan_config(fd, dict, lineno)
+            lineno = scan_column_names(fd, dict, lineno)
+            if not is_fixed_param:
+                lineno = scan_warmup_iters(fd, dict, lineno)
+                lineno = scan_hmc_params(fd, dict, lineno)
+            lineno = scan_sampling_iters(fd, dict, lineno)
+        except ValueError as e:
+            raise ValueError("Error in reading csv file: " + path) from e
     return dict
 
 
@@ -897,7 +900,7 @@ def scan_sampling_iters(
                 'line {}: bad draw, expecting {} items, found {}\n'.format(
                     lineno, num_cols, len(line.split(','))
                 )
-                + 'This error can be caused by running out of disk space.\n'
+                + 'This error could be caused by running out of disk space.\n'
                 'Try clearing up TEMP or setting output_dir to a path'
                 ' on another drive.',
             )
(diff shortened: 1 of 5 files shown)


cmdstanpy/utils.py
```diff
@@ -894,9 +894,12 @@ def scan_sampling_iters(
         data = line.split(',')
         if len(data) != num_cols:
             raise ValueError(
-                'line {}: bad draw, expecting {} items, found {}'.format(
+                'line {}: bad draw, expecting {} items, found {}\n'.format(
```
mitzimorris, line 897: we should add logic to calling function `scan_sampler_csv` so that it catches raised ValueError and adds filepath to the error message.

cmdstanpy/utils.py
```diff
@@ -894,9 +894,12 @@ def scan_sampling_iters(
...
                     lineno, num_cols, len(line.split(','))
                 )
+                + 'This error can be caused by running out of disk space.\n'
```
mitzimorris, line 900: perhaps change 'can' to 'could' (subjunctive!)

mitzimorris (changes requested):

cmdstanpy/utils.py
WardBrian: Catching and raising from these errors causes all our tests for the exceptions to fail, as `assertRaisesRegex` only looks at the first exception, not any of the ones it is raised `from`. Still want me to do it?

mitzimorris: if there's a way to make the unit tests less brittle, that would be great, but yes, please make this change - I was tracking down this error with a user, and had the filepath name been available, that would have made things a whole lot easier.

test/test_utils.py
```diff
@@ -359,6 +360,18 @@ def test_write_stan_json_bad(self):
 
 
 class ReadStanCsvTest(unittest.TestCase):
+    # pylint: disable=invalid-name
+    @contextlib.contextmanager
+    def assertRaisesRegexNested(self, exc, msg):
```
mitzimorris, line 365: great - many thanks!

WardBrian: This is a little brittle because the traceback also contains the lines that triggered the error, which _in theory_ could match the regex. I think it's nearly impossible for this to happen in a well specified test though.

WardBrian:
Fixing tests now

test/test_utils.py
WardBrian: I've resolved the above isssue

mitzimorris (approved):
LGTM!

Issue #436, opened 2021-08-16
[FR] Use new num_chains arg in cmdstan


mitzimorris:
#### Summary:

Update both the CmdStanPy APIs and documentation to allow users to use CmdStan's new single-process multi-chain processing, introduced in 2.28.

As of 2.28, CmdStan provides two kinds of parallel processing:

- Running multiple sampler chains in a single CmdStan process (introduced in 2.28), 
- Within-chain parallelization using Stan functions `reduce_sum` and `map_rect` (since 2.23 and 2.18, respectively)

The CmdStan arguments `num_chains` and `num_threads` control the amount of parallelization.

- `num_chains` specifies the number of chains, equivalent to CmdStanX sampler arg `chains`
- `num_threads` specifies the total number of threads available to the TBB scheduler across all chains.

The CmdStanX interfaces use 3 arguments altogether: `chains`, and `parallel_chains`, `threads_per_chain`, which can be mapped into `num_chains` and `num_threads`.

However we need to decide what the default should be w/r/t when single-process multi-chain CmdStan is used/ not used and whether or not we need to add a new argument to the `compile` and `sample` commands.

#### Description:

All threading is handled by the TBB scheduler.  In order to take advantage of multi-threading, models must be compiled with C++ compiler option `STAN_THREADS=true`.

Since CmdStan 2.27, the compiled executable has option `info` which will report whether or not it was compiled with `STAN_THREADS=true`. See stan-dev/cmdstan#1010

The CmdStan Guide provides the following documentation for`num_chains` and `num_threads`

-  https://mc-stan.org/docs/2_28/cmdstan-guide/mcmc-intro.html#using-the-num_chains-argument-to-run-multiple-chains
- https://mc-stan.org/docs/2_28/cmdstan-guide/parallelization.html#multi-threading-with-tbb

Prior to 2.28, in order to run multiple chains (highly recommended) directly in CmdStan, users were directed to use bash shell `for` loops, or else use any of CmdStanR, CmdStanPy, etc. interfaces.

From the current CmdStanPy documentation:

> In order to evaluate the fit of the model to the data, it is necessary to run several Monte Carlo chains and compare the set of draws returned by each. By default, the sample command runs 4 sampler chains, i.e., CmdStanPy invokes CmdStan 4 times. CmdStanPy uses Python’s subprocess and multiprocessing libraries to run these chains in separate processes. This processing can be done in parallel, up to the number of processor cores available.

[...]

PR #492, opened 2021-11-05
Feature/436 num chains

mitzimorris:
#### Submission Checklist

- [x] Run unit tests

#### Summary
Added logic to expose CmdStan 2.28 cross-chain parallelism via argument `num_chains`.
When model has been compiled with `STAN_THREADS=true`, `sample` method will use `num_chains` arg to run multi-chain process by default.

- Refactored RunSet object so that is has attributes for both number of subprocesses and number of chains.  Removed redundant checks on input args in RunSet; documented RunSet class accordingly.

- Added tqdm hook to handle multi-chain processes output.

#### Copyright and Licensing


mitzimorris:
Setting up PR.  more unit tests would be good.

cmdstanpy/model.py
```diff
@@ -898,9 +906,28 @@ def sample(
...
-                args=args, chains=chains, chain_ids=chain_ids, time_fmt=time_fmt
-            )  # bookkeeping object for command, result, filepaths
+
+            # 2.28 and up default: if exe compiled w/ threads, run single proc
+            # CmdStanArgs ctor validates model exe file
+            num_threads = threads_per_chain
+            one_process_per_chain = True
+            if (
+                force_one_process_per_chain is None
+                or force_one_process_per_chain is False
```
WardBrian, line 916: This logic isn't what we discussed in #436 - if `force_one_process_per_chain` is false we should assume this works and not do the check

cmdstanpy/model.py
```diff
@@ -898,9 +906,28 @@ def sample(
...
+            ):
+                try:
+                    info = StringIO()
+                    do_command(cmd=[str(self.exe_file), 'info'], fd_out=info)
+                    lines = info.getvalue().split('\n')
+                    if lines[3] == 'STAN_THREADS=true':
```
WardBrian, line 922: I preferred the former logic which wasn't so brittle. Assuming this will be the 4th line of the output is bound to cause problems in the future

cmdstanpy/model.py
```diff
@@ -921,18 +948,24 @@ def sample(
...
+            runset = RunSet(
+                args=args,
+                chains=chains,
+                chain_ids=chain_ids,
+                time_fmt=time_fmt,
+                one_process_per_chain=one_process_per_chain,
+            )
             with ThreadPoolExecutor(max_workers=parallel_chains) as executor:
-                for i in range(chains):
+                for i in range(runset.num_procs):
```
WardBrian, line 959: I don't love having this be unconditionally inside a threadpoolexecutor - isn't python going to spawn a bunch of threads (due to `parallel_chains`)?

IMO if we're only spawning one process we should avoid doing this

cmdstanpy/model.py
```diff
@@ -1302,18 +1337,32 @@ def _run_cmdstan(
...
+        if show_progress and runset.one_process_per_chain:
+            progress_hook = self._wrap_sampler_per_chain_progress_hook(
+                chain_id=idx + 1, total=iter_total
+            )
+        elif show_progress:
+            progress_hook = self._wrap_sampler_multi_chain_progress_hook(
+                num_chains=runset.chains,
+                offset=runset.chain_ids[0],
+                total=iter_total,
+            )
```
WardBrian, line 1363: Is there a reason to prefer this over using one callback that handles both cases?

cmdstanpy/model.py
```diff
@@ -1343,17 +1393,18 @@ def _run_cmdstan(
...
             fd_out.close()
         except OSError as e:
             msg = 'Failed with error {}\n'.format(str(e))
             raise RuntimeError(msg) from e
         finally:
             fd_out.close()
 
-        if not show_progress:
-            get_logger().info('finish chain %d', idx + 1)
+        if show_console:
```
WardBrian, line 1404: I think we want to print the finished information even if show_console isn't true? Right now if you set both show_progress=False and show_console=False, do you not get any indication of the start/finish?

In general, most calls which use the logger should happen unconditionally. If a user doesn't want them, they can just assign `logger('cmdstan')` a `NullHandler`. The second call here should probably be debug, not info, though.

cmdstanpy/stanfit.py
```diff
@@ -229,10 +192,35 @@ def chain_ids(self) -> List[int]:
...
+            return self._args.compose_command(
+                idx,
+                csv_file=self.file_path('.csv'),
+                diagnostic_file=self.file_path(".csv", extra="-diagnostic")
+                if self._args.save_latent_dynamics
+                else None,
+                profile_file=self.file_path(".csv", extra="-profile")
+                if self._args.save_profile
+                else None,
+                num_chains=self._chains,
```
WardBrian, line 222: ah, I see why `num_chains` is part of the compose for the general args class and not SamplerArgs. I think it's fine for now, barring a bigger refactor of the args classes.

test/test_sample.py
```diff
@@ -321,25 +322,32 @@ def test_multi_proc(self):
...
-            ('cmdstanpy', 'INFO', 'finish chain 1'),
-            ('cmdstanpy', 'INFO', 'start chain 2'),
+            ('cmdstanpy', 'INFO', 'Chain [1] done processing'),
+            ('cmdstanpy', 'INFO', 'Chain [2] start processing'),
         )
+
+    # pylint: disable=no-self-use
+    def test_multi_proc_2(self):
+        logistic_stan = os.path.join(DATAFILES_PATH, 'logistic.stan')
+        logistic_model = CmdStanModel(stan_file=logistic_stan)
```
WardBrian, line 335: Are we going to add a test which uses the new logic? We don't even need to compile with `STAN_THREADS=True`, as it should just run the chains sequentially if we don't (though I see no issue with passing that cpp option in the tests)

test/test_sample.py
```diff
@@ -411,6 +418,18 @@ def test_multi_proc(self):
             logistic_model.sample(
                 data=logistic_data, chains=4, threads_per_chain=-4
             )
+        # this logic is wrong.  write test for 2.28
```
WardBrian, line 421: Plan for this?

WardBrian (changes requested):
Thanks again for doing this so quickly

cmdstanpy/model.py
mitzimorris: much easier to implement - you've fundamentally got two different data structures to manage - a single progress bar or an array of progress bars.  if you have a single method with two different data structures, it's really easy to mess up on the update statements.

I can take a closer look at the errors/exceptions thrown by the tqdm library and see if there's a more elegant solution.  In my option, it's whack-a-mole on understandability and maintainability here.

WardBrian: My partial solution was just to always use the “list of progress bars”, even in the one-chain-one-process case. You have to make the hook a parameter of the _run_cmdstan method, but since it’s only used by one of the methods that almost makes more sense IMO.

mitzimorris:
thanks for reviewing this so quickly.   great feedback - will address over the weekend.

cmdstanpy/model.py
mitzimorris: we are trying to protect the user from the unfriendliness of the CmdStan CLI.  if we assume this works and spawn a process with arg `num_chains` present, the error is not helpful at all.

cmdstanpy/cmdstan_args.py
```diff
@@ -897,7 +897,8 @@ def compose_command(
         csv_file: str,
         *,
         diagnostic_file: Optional[str] = None,
-        profile_file: Optional[str] = None
+        profile_file: Optional[str] = None,
+        num_chains: Optional[int] = None
```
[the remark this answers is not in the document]
mitzimorris: you mean num_chains? 
in theory, you might want to run multiple chains for optimization and advi - especially multi-path pathfinder.
so yes, belongs in CmdStan args.

test/test_sample.py
mitzimorris: yes

cmdstanpy/model.py
mitzimorris: does is spawn threads lazily?

mitzimorris:
regarding CmdStan args vs. method-specific args, the CmdStan CLI requires that args be grouped under headings "method", "data", "output".  Given this, you're correct that `num_chains` should be part of the sampler args.

cmdstanpy/model.py
WardBrian: The docs seem to imply it is fine as is. It think the code is misleading to read however, as the spawning of threads and the loop are unnecessary in the (hopefully increasingly common) use case going forward

As an aside, should we define a 'minimum cmdstan version' we still support? E.g., at some point/version in the future we could just assume you have 2.28 or newer

cmdstanpy/model.py
WardBrian: If you would like to not allow the user to force the behavior, there is no point having the default value be `None` - you can make the type non-optional and default to false. 

However, `num_chains` as an argument does have behavior that exists even if `STAN_THREADS` was not set. Without an override, there is no way to trigger this in cmdstanpy. Now, would anyone ever want it? Maybe in testing, but probably not in practice.

I think my preferred solution is to say `None` works as it currently does and does the check, `True` does the prior behavior where we manually spawn multiple processes, and `False` always uses 'num_threads', but either outputs a warning or verifies the _version_, but not `STAN_THREADS`. Does that make sense?

mitzimorris: yes, makes sense.  the version check is necessary.

WardBrian: Great. IMO, if `force_one_process_per_chain` is set to `False` by the user, and the version is  < 2.28, that should be an exception, not a 'silently do something else' situation. If they're specifically asking for the behavior, we should do that or throw an error saying we can't

cmdstanpy/model.py
mitzimorris: agreed, but how long should the window be?   I guess if we drop support for older versions of CmdStan in the future, then users can pin the CmdStanPy version to an earlier version of the package.

WardBrian: I think something like “the previous 3 versions from the time of CmdStanPy release”

so 1.0 for example would support 2.25+, and if we released a 1.1 following 2.29, it would only guarantee support for 2.26+ 

3 is of course an arbitrary number. Right now RStan is so far behind that it’s worth supporting as old as we can reasonably manage if people will be working in multiple ecosystems

cmdstanpy/model.py
mitzimorris: refactored into a single method.  the conditional logic is one thing, the bookkeeping is another - the combination is maddening.

cmdstanpy/model.py
mitzimorris: inasmuch as we have the "force_one_process_per_chain" option, we need this logic and it works for all cases.

cmdstanpy/model.py
mitzimorris: > Right now if you set both show_progress=False and show_console=False, do you not get any indication of the start/finish?

per-chain logging messages mess up the TQDM progress bars in a bad way, so we don't want to do this in _run_cmdstan.
instead, added global "start processing / done processing" messages in the `sample` method.

reasoning: if you don't want to see the progress bars, then all you care about is overall completion.

cmdstanpy/model.py
mitzimorris: added util method `model_info` which parses console outputs into dict of key value pairs.

cmdstanpy/model.py
```diff
@@ -898,9 +883,55 @@ def sample(
...
+                        info_dict.get('STAN_THREADS') == 'true'
+                        and major == 2
+                        and minor > 27
+                    ):
+                        one_process_per_chain = True
+                        num_threads = parallel_chains * num_threads
+                        parallel_procs = 1
+                elif force_one_process_per_chain is False:
+                    if major == 2 and minor < 28:
+                        get_logger().info(
```
WardBrian, line 932:
```suggestion
                        get_logger().warning(
```

cmdstanpy/model.py
```diff
@@ -1378,26 +1431,53 @@ def _run_cmdstan(
     @staticmethod
     @progbar.wrap_callback
     def _wrap_sampler_progress_hook(
-        chain_id: int, total: int
+        one_process_per_chain: bool,
```
WardBrian, line 1437: I think we should implement this such that there is only ever one callback, it may just be called on multiple threads. It avoids the need for the if/else on the core logic.

As a sketch:
```python
    @staticmethod
    @progbar.wrap_callback
    def _wrap_sampler_progress_hook(
        chain_ids: int,
        total: int,
    ) -> Optional[Callable[[str], None]]:
        """
        Sets up tqdm callback for CmdStan sampler console msgs.
        CmdStan progress messages start with "Iteration", for single chain
        process, "Chain [id] Iteration" for multi-chain processing.
        For the latter, manage array of pbars, update accordingly.
        """
        pat = re.compile(r'Chain \[(\d*)\] (Iteration.*)')

        pbars: Dict[int, tqdm] = 
        {chain_id: tqdm(
                    total=total,
                    bar_format="{desc} |{bar}| {elapsed} {postfix[0][value]}",
                    postfix=[dict(value="Status")],
                    desc=f'chain {chain_id}',
                    colour='yellow',
            )
            for chain_id in chain_ids
        }

        def progress_hook(line: str, id: Optional[int] = None) -> None:
            if line == "Done":
                for pbar in pbars.values():
                    pbar.postfix[0]["value"] = 'Sampling completed'
                    pbar.update(total - pbar[i].n)
                    pbar.close()
            if id is None:
                match = pat.match(line)
                if match:
                    idx = int(match.group(1))
                else:
                    return # no update to do
            else:
                idx = id
                
                
            if 'Sampling' in line:
                pbar[idx].colour = 'blue'
            pbar[idx].update(1)
            pbar[idx].postfix[0]["value"] = line

        return progress_hook
```

This then lets you call either with a known id (if we're in the one-one process-chain case) or just the line and it parses it. The only other change is that this is instantiated in `sample()` instead of `run_cmdstan`, but I actually like that better (if we ever wanted a progress hook for any other cmdstan command, we'd need to do this refactor anyway)

WardBrian: This is what I was working toward last week when I did some work on this PR, but I didn't get quiet to the use of the dictionary to handle chain IDs. The stuff I had with the dummy callback as the default argument should still work for `run_cmdstan` though, if we go this route

mitzimorris: nice - dict solves indexing problem!

mitzimorris: agreed this should be instantiated in sample method - at one point is was.  this code is still maddening and I'm really frustrated by the amount of logic in CmdStanPy necessary to deal with the deadly combo of CmdStan arg parser and Stan CSV files.

WardBrian: I'm very much so looking forward to using something like JSON+[pydantic](https://pydantic-docs.helpmanual.io/) to do the first part if CmdStan really does move that way. The CSV part will always be a bit annoying, simply because a row-based format will never be very good at representing n-dimensional arrays.

mitzimorris: sorry, cannot make above sketch work.  timing myself out.

test/test_sample.py
mitzimorris: updated tests.

test/test_sample.py
mitzimorris: added test for 2.28

cmdstanpy/model.py
mitzimorris: upgraded to warning, added more info

cmdstanpy/model.py
mitzimorris: maybe you can't update tqdm progress bars can be called from different threads?

WardBrian: My implementation from last week worked in my testing with the threadprocesspool

WardBrian: Want me to take a look? Waiting on some stan_math changes this week

mitzimorris: sure, go ahead.  it's also possible I messed up on the types or signatures.

mitzimorris:
if all unit tests pass, this is ready for re-review.

cmdstanpy/model.py
mitzimorris: actually, that was exactly the problem - overlooked sig in a type hint.  fixed.  will also add debugging hooks for tqdm.

WardBrian:
Great, I'll review after that change. 

One thing that immediately strikes out to me: I think the info_dict method may be useful for other stuff in the future and might belong as part of the `CmdStanModel` class. That would also allow us to move the `if exe_file is not None` logic into the method itself, so we could just call `model.info()` and get back either the dictionary or None

WardBrian:
Actually, to add on to the above comment, we should make info_dict return `{}` instead of `None`. That way we can (always) safely do something like `model.info_dict().get('STAN_THREADS')` without checking the return first.

cmdstanpy/model.py
mitzimorris: progress bar callback hook is now instantiated in sample method and passed into `_run_cmdstan`.
not quite as fancy as your sketch, but relatively clean.

mitzimorris:
the code is ready for review, but we still need to update the doc and and examples for 2.28.

the changes made on the CmdStan side w/r/t `fixed_params` should make people's lives easier (even if the code ties itself in knots trying to check the console outputs and CSV files).

WardBrian (changes requested):
Thanks for all the work on this. A few more nitpicky things and hopefully the last go-round on the progress bars

cmdstanpy/model.py
```diff
@@ -898,41 +914,100 @@ def sample(
...
+                    'Argument threads_per_chain must be a positive integer, '
+                    'found {}.'.format(threads_per_chain)
+                )
+
+            parallel_procs = parallel_chains
+            num_threads = threads_per_chain
+            one_process_per_chain = True
+            if (
+                force_one_process_per_chain is None
+                and not cmdstan_version_before(2, 28)
```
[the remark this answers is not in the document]
mitzimorris: OK.

cmdstanpy/model.py
```diff
@@ -898,41 +914,100 @@ def sample(
...
+                assert isinstance(self.exe_file, str)  # make typechecker happy
+                info_dict = self.exe_info()
+                if info_dict.get('STAN_THREADS') == 'true':
+                    one_process_per_chain = False
+                    num_threads = parallel_chains * num_threads
+                    parallel_procs = 1
+            elif (
+                force_one_process_per_chain is False
+                and cmdstan_version_before(2, 28)
+            ):
```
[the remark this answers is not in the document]
mitzimorris: these are all set correctly above - lines 941-943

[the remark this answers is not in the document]
WardBrian: If `force_one_process_per_chain` is false, they should be set to use the single-process multi-thread, e.g:

```python
                    one_process_per_chain = False
                    num_threads = parallel_chains * num_threads
                    parallel_procs = 1
```

Right now, force_one_process_per_chain=False does the same thing as force_one_process_per_chain=True, just with an extra version check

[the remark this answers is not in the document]
mitzimorris: you're right - the logic got jumbled during earlier changes.  added more logic.

cmdstanpy/utils.py
```diff
@@ -234,7 +236,18 @@ def cmdstan_version_before(major: int, minor: int) -> bool:
...
+        cur_version = cmdstan_version()
+    else:
+        if (
+            info['stan_version_major'] is not None
+            and info['stan_version_minor'] is not None
+        ):
+            cur_version = (
+                int(info['stan_version_major']),
+                int(info['stan_version_minor']),
+            )
```
WardBrian, line 250: This will fail to get any version if info_dict is `{}`, which is a possibility. I guess in that case we know the cmdstan version is less than we're asking of it?

cmdstanpy/model.py
```diff
@@ -1389,8 +1393,8 @@ def _run_cmdstan(
         logger_prefix = 'CmdStan'
         console_prefix = ''
         if runset.one_process_per_chain:
-            logger_prefix = 'Chain [{}]'.format(idx + 1)
-            console_prefix = 'Chain [{}] '.format(idx + 1)
+            logger_prefix = 'Chain [{}]'.format(idx + runset.chain_ids[0])
+            console_prefix = 'Chain [{}] '.format(idx + runset.chain_ids[0])
```
WardBrian, line 1409: In the one process per chain case there's no reason to assume the chain ids are contiguous in general. Can we just use `runset.chain_ids[idx]`?

cmdstanpy/model.py
```diff
@@ -941,19 +941,22 @@ def sample(
             parallel_procs = parallel_chains
             num_threads = threads_per_chain
             one_process_per_chain = True
+            assert isinstance(self.exe_file, str)  # make typechecker happy
+            info_dict = self.exe_info()
+            stan_threads = info_dict.get('STAN_THREADS')
+            if stan_threads is not None:
```
WardBrian, line 947: I prefer using the default value functionality of `get` rather than the extra if case. `stan_threads = info_dict.get('STAN_THREADS', 'false').lower()`

cmdstanpy/model.py
```diff
@@ -949,21 +949,33 @@ def sample(
...
+                        chains,
+                    )
+                elif stan_threads is None or stan_threads == 'false':
+                    get_logger().warning(
+                        'Stan program not compiled for threading, '
+                        'will run %d processes. '
+                        'Recompile model and specify argument '
+                        '"cpp_options={\'STAN_THREADS\':\'TRUE\'}.',
+                        chains,
+                    )
```
WardBrian, line 973: I don't think we should do this IMO. CmdStan provides the functionality to do multiple chains in one process without threading, we should let users have that if they explicitly ask for it. We can emit a messaging saying that is what is happening, but we shouldn't make that feature impossible to access

WardBrian (commented):
I think this is probably my last comments. This PR is looking great when you consider how inherently messy the cmdstan interface is

cmdstanpy/utils.py
WardBrian: The way I would change it is `info is None or 'stan_version_major' not in info`

WardBrian (approved):
LGTM - 1.0 here we come!

mitzimorris:
this totally sucks - discovered that model method `compile` with option `force=True` doesn't do the right thing and there are no unit tests for this either.  wtf?

WardBrian:
Unless you believe that bug was introduced by this PR I think it’s better to merge this and handle that separately. These big PRs get very hard to review

mitzimorris:
this bug has been there all along - that's why unit tests are removing exe files here and there.  I'm just surprised that this wasn't tested more back when we added this feature.

PR #496, opened 2021-11-15
Bugfix/force compile

mitzimorris:
#### Submission Checklist

- [x] Run unit tests

#### Summary

Cleanup of `compile` command logic with unit test to check that compile arg `force` does the right thing.

Also add logic to CmdStanModel ctor and `sample` method to handle CmdStan https://github.com/stan-dev/cmdstan/issues/1057

#### Copyright and Licensing


mitzimorris:
still needs more unit tests, including interaction w/ `fixed_param` and `num_chains`

test/test_model.py
```diff
@@ -153,22 +149,31 @@ def test_cpp_options(self):
...
+        # model.compile(force=True, cpp_options=override_opts)
+        # dict2 = model.exe_info
+        # get_logger().info('dict 2 %s', dict2)
+        # self.assertEqual(dict2['STAN_THREADS'].lower(), 'false')
+
+        model.compile(
+            force=True, cpp_options=override_opts, override_options=True
+        )
+        dict3 = model.exe_info
+        get_logger().info('dict 3 %s', dict3)
```
WardBrian, line 175: Why are we logging in our tests (as opposed to `print`)?

test/test_model.py
```diff
@@ -153,22 +149,31 @@ def test_cpp_options(self):
...
     def test_model_info(self):
-        # copy so that parallel compile tests don't mess up exe
-        b2_filename = os.path.join(DATAFILES_PATH, 'b2.stan')
-        b2_file = shutil.copyfile(BERN_STAN, b2_filename)
-        cpp_opts = {'STAN_THREADS': 'TRUE'}
-        model = CmdStanModel(stan_file=b2_file, cpp_options=cpp_opts)
-        if model.exe_file is not None and os.path.exists(model.exe_file):
-            os.remove(model.exe_file)
-        empty_dict = model.exe_info()
-        self.assertEqual(len(empty_dict), 0)
```
WardBrian, line 164: I think this is a reasonable test to still have

cmdstanpy/model.py
```diff
@@ -942,17 +942,19 @@ def sample(
...
             one_process_per_chain = True
             assert isinstance(self.exe_file, str)  # make typechecker happy
-            info_dict = self.exe_info()
+            info_dict = self.exe_info
             stan_threads = info_dict.get('STAN_THREADS', 'false').lower()
+            # run multi-chain sampler if possible, unless algo is fixed_param
             if (
                 force_one_process_per_chain is None
                 and not cmdstan_version_before(2, 28, info_dict)
+                and not fixed_param
```
WardBrian, line 951: Should probably just use `force_one_process_per_chain = force_one_process_per_chain or fixed_param`

WardBrian (commented):
Comments for pair programming

WardBrian:
I'll do some more Windows-specific bug hunting tomorrow

WardBrian:
Just a thought: We will now recompile if:

1. `force` is set to True
2. The compiler options are different.

Would it make sense to add "if the existing exe file was compiled with an older version of cmdstan than the current interpreter is using"? This would be useful but essentially impossible to test in our CI

WardBrian:
@mitzimorris We need some tests that instantiate a CmdStanModel class with the exe file only. Just caught a bug in testing where model._fixed_param was not initialized in this case, so .sample failed

WardBrian:
An extension I think we should allow: in the CmdStanModel constructor, we currently allow you to set `compile=True|False`. I should you should also be able to set it `compile="force"` which does the obvious thing.

If you think this is a good idea @mitzimorris I'll implement it tomorrow before doing the Windows debugging

mitzimorris:
> Would it make sense to add "if the existing exe file was compiled with an older version of cmdstan than the current interpreter is using"? This would be useful but essentially impossible to test in our CI

I'm not sure - why don't you ask on Discourse?  once a model is written and working, new language features don't matter, so the only benefit would be bugfixes/performance enhancements in Stan and/or mathlib.  which sometimes aren't as great as hoped.

mitzimorris:
> in general I think we should actually call sample etc on models constructed this way

not sure what you mean here.

the problem is that compilation takes many seconds if not a minute.  consider the teaching/demo use case:  there's a lecture's worth of model and data plus notebook.  new CmdStan release comes out, so user upgrades.  then goes to teach from the notebook - suddenly everything's recompiling - possibly unexpected - kinda ruins the packing of the show.

WardBrian:
Oh my last comment was related to the unit test you just added. I think it’s important we call .sample(), .variational() and .optimize() on CmdStanModel objects which are exe only in the unit tests, not just construct them

mitzimorris:
just pushed changes to fix logic around fixed_param and exe only.  will add more unit tests today.

WardBrian:
Could you explain the idea behind having the fixed_param flag be None by default? The logic doesn't seem to use it anywhere specifically

WardBrian:
Found issue: TemporaryCopiedFile returns a sanitized path even if `is_copied` is False. We should use this unconditionally as we have before. Doing that and some other clean up now, including renaming it to `SanitziedOrTmpFilePath`

mitzimorris:
w/r/t fixed_param = None,  will check the logic again - one of the added unit tests triggered it.

WardBrian:
As far as I can tell, the logic we have right now is completely ambivalent as to whether fixed_param is None or False, so we can just make it false. I'm in the middle of some logic refactors so I'd hold off for now to avoid merge issues

WardBrian:
Seems to now be passing on all platforms. Will look into `compile="force"` behavior and general cleanup.

WardBrian:
The solution I found in general required passing `shell=True` to `Popen` on Windows. This is generally not recommended, but without it the spec says Popen may completely ignore both `env` and `cwd` arguments:

>Resolving the path of executable (or the first item of args) is platform dependent. For POSIX, see os.execvpe(), and note that when resolving or searching for the executable path, cwd overrides the current working directory and env can override the PATH environment variable. For Windows, see the documentation of the lpApplicationName and lpCommandLine parameters of WinAPI CreateProcess, and note that when resolving or searching for the executable path with shell=False, cwd does not override the current working directory and env cannot override the PATH environment variable. Using a full path avoids all of these variations.

WardBrian:
If I come back from lunch and all tests pass then I think I'm happy with the state of this. That said, feel free to add more tests and I can them if they fail

WardBrian (commented):
I think this counts as my final review, for whatever that means in the pair-programming system we used for this PR

docsrc/workflow.rst
```diff
@@ -64,10 +67,6 @@ The constructor argument ``compile=False`` will override the default behavoir.
     my_model.exe_file
     my_model.code()
 
-The method :meth:`~CmdStanModel.compile` is used to compile the model as needed.
```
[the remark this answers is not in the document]
mitzimorris: workflow page is def a work in progress.  am trying to explain the overlap between constructor args and compile args.  I will continue this in a separate PR.

I can restore that sentence and then I think we should merge this PR.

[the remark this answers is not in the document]
WardBrian: I agree

WardBrian:
We seem to be getting that intermitten `Operation not permitted` error in variational tests - I don't recall what the solution was the last time we did this, but I'm also unsure why it would be happening now? Seems to be the exe_only test

mitzimorris:
> We seem to be getting that intermittent Operation not permitted error in variational tests - I don't recall what the solution was the last time we did this, but I'm also unsure why it would be happening now? 

will investigate - I think ADVI sometimes fails on that model unless you specify a specific seed, etc.  which is why all other calls in varatiational test have all args specified.  I forgot this detail when adding the new tests

WardBrian:
Great. I'm also having an issue with test_utils.test_cmdstan_version clobbering my CMDSTAN env variable on windows and then causing all subsequent tests to fail, so I'd like to fix that before merge as well

WardBrian:
I don't have all the dev tools set up properly on my windows machine, apologies for linting issues. If tests pass this is good to go IMO

Issue #506, opened 2021-12-04
Failed compilation leaves process cwd in CmdStan folder


maedoc:
#### Summary:
A failing compilation leaves the cwd in the CmdStan folder.

#### Description:
The [`do_command`](https://github.com/stan-dev/cmdstanpy/blob/0b50ae9a8bfdbe011cc9eb258a6f2477d7aefc90/cmdstanpy/utils.py#L1006) function uses a context manager which doesn't seem to restore the previous working directory, when a compilation failure occurs.

#### Additional Information:
The `pushd` context manager probably needs to catch the exception, restore the working directory and then re-raise the exception.

#### Current Version:
cmdstanpy-1.0.0

WardBrian:
Closed by #507

PR #507, opened 2021-12-05
Restore cwd after do_command error WIP

maedoc:
#### Submission Checklist

- [x] Run unit tests

#### Summary

Addresses #506.

#### Copyright and Licensing

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,5 +1,5 @@
 name: CmdStanPy
-    
+
 on:
   push:
     branches:
@@ -14,7 +14,7 @@ on:
         description: 'Version to test'
         required: false
         default: 'latest'
-      
+
 jobs:
   get-cmdstan-version:
     # get the latest cmdstan version to use as part of the cache key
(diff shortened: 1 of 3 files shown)


maedoc:
It would be better if `pushd` was a proper context manager which handled exceptions by restoring cwd but this is easier and fixes the issue at hand.  The only other use of `pushd` is in the install cmdstan/toolchain functions.

WardBrian:
You can use try/except blocks inside a [contextlib.contextmanager](https://docs.python.org/3/library/contextlib.html#contextlib.contextmanager) function like pushd. I think replacing the `yield` in pushd with

```python
try:
    yield
except Exception:
    raise
finally:
    os.chdir(previous_dir)
```

would resolve the issue and be much more general

We can also then test pushd directly:

```python
dir = os.getcwd()
with self.AssertRaises(RuntimeError):
    with pushd('data'):
        raise RuntimeError('testing')
self.assertPathsEqual(dir, os.getcwd())
```

ahartikainen:
Yes I think adding

```python
try:
    yield
finally:
    os.chdir(previous_dir)
```

is needed

https://github.com/stan-dev/cmdstanpy/blob/5b80eb84ad1bf8fa093013d3b06a1c3594093092/cmdstanpy/utils.py#L1317

WardBrian:
@maedoc Let me know if you'd prefer if myself or Ari take the PR from here in terms of changes. Thanks for finding this issue and jumping on it so quickly!

maedoc:
All the test failures are trying to compile the model with CmdStan on macOS.  The error is the same as when I try to use a CmdStan build for x86_64 on Rosetta on macOS from Python running on arm64, or vice versa.  Perhaps the CI script needs to be updated to check arch and rebuild CmdStan accordingly.

WardBrian:
I believe the error is because GitHub Actions updated the version of MacOs used but is caching old cmdstans. I’ll look into it

Thanks for the changes! I believe that with assertRaises there’s no need for a try/except in the test code. Other than that it looks great and I’ll merge once I fix the CI issues

maedoc:
> I believe that with assertRaises there’s no need for a try/except in the test code.

Some static analysers don't know that `self.assertRaises` catches exceptions and see the subsequent code as unreachable.  Up to you..

WardBrian:
We're not using try/except anywhere else in the tests where we use assertRaises, so for consistency I think don't here

WardBrian:
The one test failure is from a timeout (I think we tried to download cmdstan too many times in short succession) so this is good to go

WardBrian (approved):

WardBrian:
Thanks again @maedoc !

PR #510, opened 2021-12-15
assemble draws and generated quantities only once

qres:
#### Submission Checklist

- [ ] Run unit tests

#### Summary
Avoid loading the generated samples multiple times if multiple variables are accessed via `stan_variable('var_name')` which can take significant amount of time for large sample files. Unfortunately I had problems to run the unit test. Would somebody else run them, please?

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting (this will be you or your assignee, such as a university or company): University of Stuttgart

### DIFF
--- a/cmdstanpy/stanfit/mcmc.py
+++ b/cmdstanpy/stanfit/mcmc.py
@@ -231,7 +231,7 @@ class CmdStanMCMC:
         CmdStanMCMC.draws_xr
         CmdStanGQ.draws
         """
-        if self._draws.size == 0:
+        if self._draws.shape == (0,):
             self._assemble_draws()
 
         if inc_warmup and not self._save_warmup:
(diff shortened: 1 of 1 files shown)


WardBrian:
Reading the code diff I am not sure why this is any different from the current behavior?

qres:
The difference comes from lines as in [`stan_variable()`](https://github.com/stan-dev/cmdstanpy/blob/870cfe196bb9049fe29fe5dbe6793ca549ff92ef/cmdstanpy/stanfit/mcmc.py#L664) with a direct call to `_assemble_generated_quantities()` without a surrounding check whether the quantities have been assembled before.

WardBrian:
Ah, which explains why it doesn't show up in the diff. Thanks for catching that. I think I'd prefer we add the if statement there rather than have the _assemble_draws function sometimes not assemble draws

ahartikainen:
Hi, could someone please explain the logic?

I'm having a hard time understand the current changes. (Not saying they are wrong, I just wonder what is the correct way to not call that function more than once)

WardBrian:
Thanks for this

PR #512, opened 2022-01-11
speed up _assemble_generated_quantities

qres:
#### Submission Checklist

- [ ] Run unit tests

#### Summary
```python
import numpy as np
import pandas as pd

def new(f):
    return pd.read_csv(
        f,
        sep=',',
        comment='#',
        header=0,
    ).values

# current implementation
def current(f):
    with open(f, 'r') as fd:
        lines = (line for line in fd if not line.startswith('#'))
        return  np.loadtxt(
            lines, dtype=np.ndarray, ndmin=2, skiprows=1, delimiter=','
        )
```

Runtime comparison for different file sizes on my laptop measured by ipython's `%timeit`:

|        | 49KB    | 2.2MB   | 66MB    | 132MB   |
|--------|---------|---------|---------|---------|
|new | 1.56 ms | 26.7 ms | 632 ms  | 1250 ms |
|current | 5.03 ms | 204 ms  | 2030 ms | 4190 ms |

The 49KB file is the one generated by the bernoulli experiment in the cmdstan repository.

Adding the parameters `quoting=3` or `low_memory=False` to `read_csv` did not improve the performance of `new`.

One difference between the two versions is, that for a file with zero samples `new` returns an array of shape `(0,num_columns)` while `current` returns an array of shape `(0,1)` and prints a `UserWarning: loadtxt: Empty input file`.

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting (this will be you or your assignee, such as a university or company): University of Stuttgart

### DIFF
--- a/cmdstanpy/stanfit/mcmc.py
+++ b/cmdstanpy/stanfit/mcmc.py
@@ -1222,7 +1222,7 @@ class CmdStanGQ:
         return result
 
     def _assemble_generated_quantities(self) -> None:
-        # use numpy loadtxt
+        # use pandas read_csv
         warmup = self.mcmc_sample.metadata.cmdstan_config['save_warmup']
         num_draws = self.mcmc_sample.draws(inc_warmup=warmup).shape[0]
         gq_sample: np.ndarray = np.empty(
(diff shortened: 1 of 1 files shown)


ahartikainen:
What are the runtimes when you have 100k columns and 100 draws?

ahartikainen:
dtype should be forced to `float`

Is that timing table correct?

qres:
The last two files are 30000 rows or 60000 rows x 236 columns.

Changing to `dtype=np.float64` did not change a lot:
| |49KB | 2.2MB | 66MB |
|--|-- | -- | -- |
|current with float64 |5.79 ms | 243 ms | 2.53 s |

> Is that timing table correct?

No. I swapped the row names. They are correct now.

> 100k columns and 100 draws

Good point. Seems like pandas does not like many columns even though the file size is still 87M. It took about 1min which is much more than the current numpy version wich takes 4.1s .

Therefore, I think it is best to keep everything as it is.

ahartikainen:
Yeah, it would be really cool to find out why pandas is so slow for multiple columns.

I used initially pandas in arviz (for cmdstan) but in the end, for huge tables (huge # of columns) pandas was just too slow compared to basic python + numpy solution.

I think currently CmdStanPy has ok (ish) solution for reading large csv files. For small the time difference is not too big.

In future, a dask solution would be something that might be best for us.

ahartikainen:
Oh, btw for a future reference, that forced float is due to stuck chains (e.g. stuck at init which has been set to integer value) -> merging int and float might cause problems. (of course I don't remember how it behaves when we put that data into a numpy array directly)

PR #515, opened 2022-01-18
Add write_stan_file function (fixes #513).

### LINKED ISSUE #513: Feature: write stan file from python.
Would there be interest in adding a convenience function to write a temporary stan file from python (happy to create a PR if so)? I've found this functionality useful for model iteration without modifying external files. In particular, I've been using the following snippet which

- computes the hash of the code
- writes the code to a hash-based filename in a given directory (defaults to the `models` subdirectory of the `cmdstanpy` directory)
- returns the path

```python
import cmdstanpy
import hashlib
import os


def write_stanfile(code: str, directory: str = None) -> str:
    """
    Write the model code to a file and return the path based on the code hash.

    Args:
        code: Code to write to a file.
        directory: Directory to write the code to (defaults to the cmdstanpy models directory).

    Returns:
        path: Path to the file containing the code.
    """
    # Generate the path based on the code and check whether it already exists.
    directory = directory or os.path.join(cmdstanpy.cmdstan_path(), 'models')
    digest = hashlib.sha256(code.encode()).hexdigest()
    stan_file = os.path.join(directory, f'{digest}.stan')
    if os.path.isfile(stan_file):
        return stan_file

    # Write the file to disk.
    os.makedirs(directory, exist_ok=True)
    with open(stan_file, 'w') as fp:
        fp.write(code)
    return stan_file
```

The function can be used like so

```python
model = cmdstanpy.CmdStanModel(stan_file=write_stanfile('[MODEL CODE HERE]'))
```

tillahoffmann:
#### Submission Checklist

- [x] Run unit tests

#### Summary

Adds a `write_stan_file` function as discussed in #513 together with unit tests.

#### Copyright and Licensing

### DIFF
--- a/cmdstanpy/utils.py
+++ b/cmdstanpy/utils.py
@@ -1445,11 +1445,9 @@ def write_stan_file(code: str, directory: Union[str, None] = None) -> str:
     :return: Path to the file containing the code..
     """
     # Generate the path based on the code and check whether it already exists.
-    directory = os.path.abspath(
-        directory or os.path.join(tempfile.gettempdir(), 'cmdstanpy_models')
-    )
+    directory = os.path.abspath(directory or _TMPDIR)
     digest = hashlib.sha1(code.encode()).hexdigest()
-    stan_file = os.path.join(directory, f'{digest}.stan')
+    stan_file = os.path.join(directory, f'model_{digest}.stan')
     if os.path.isfile(stan_file):
         return stan_file
 
--- a/cmdstanpy/utils.py
+++ b/cmdstanpy/utils.py
@@ -3,6 +3,7 @@ Utility functions
 """
 import contextlib
 import functools
+import hashlib
 import logging
 import math
 import os
@@ -1431,3 +1432,27 @@ class SanitizedOrTmpFilePath:
     def __exit__(self, exc_type, exc_val, exc_tb) -> None:  # type: ignore
         if self._tmpdir:
             shutil.rmtree(self._tmpdir, ignore_errors=True)
+
+
+def write_stan_file(code: str, directory: Union[str, None] = None) -> str:
+    """
+    Write the model code to a file and return the path based on the code hash.
+
+    :param code: Code to write to a file.
+    :param directory: Directory to write the code to (defaults to the cmdstanpy
+        models directory).
+
+    :return: Path to the file containing the code..
+    """
+    # Generate the path based on the code and check whether it already exists.
+    directory = directory or os.path.join(cmdstan_path(), 'models')
+    digest = hashlib.sha1(code.encode()).hexdigest()
+    stan_file = os.path.join(directory, f'{digest}.stan')
+    if os.path.isfile(stan_file):
+        return stan_file
+
+    # Write the file to disk.
+    os.makedirs(directory, exist_ok=True)
+    with open(stan_file, 'w') as stream:
+        stream.write(code)
+    return stan_file
--- a/cmdstanpy/utils.py
+++ b/cmdstanpy/utils.py
@@ -1445,7 +1445,9 @@ def write_stan_file(code: str, directory: Union[str, None] = None) -> str:
     :return: Path to the file containing the code..
     """
     # Generate the path based on the code and check whether it already exists.
-    directory = directory or os.path.join(cmdstan_path(), 'models')
+    directory = os.path.abspath(
+        directory or os.path.join(tempfile.gettempdir(), 'cmdstanpy_models')
+    )
     digest = hashlib.sha1(code.encode()).hexdigest()
     stan_file = os.path.join(directory, f'{digest}.stan')
     if os.path.isfile(stan_file):
--- a/cmdstanpy/__init__.py
+++ b/cmdstanpy/__init__.py
@@ -55,6 +55,7 @@ from .utils import (
     set_make_env,
     show_versions,
     write_stan_json,
+    write_stan_file,
 )
 
 __all__ = [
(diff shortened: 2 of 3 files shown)


cmdstanpy/utils.py
```diff
@@ -1431,3 +1432,27 @@ def __enter__(self) -> Tuple[str, bool]:
...
+    Write the model code to a file and return the path based on the code hash.
+
+    :param code: Code to write to a file.
+    :param directory: Directory to write the code to (defaults to the cmdstanpy
+        models directory).
+
+    :return: Path to the file containing the code..
+    """
+    # Generate the path based on the code and check whether it already exists.
+    directory = directory or os.path.join(cmdstan_path(), 'models')
```
WardBrian, line 1448: I don't know if I like the idea of a cmdstan models directory. Some immediate questions/concerns come to mind:

Does it get deleted if cmdstan is rebuilt? When installing a new version of cmdstan, all older models are left behind. I as the user may not actually know where my cmdstan folder is, and unless I'm printing the return of this function won't necessarily know that they're stored here.

IMO, we should either require the directory parameter or initialize it to `"."` (the current working directory). Alternatively, we could write all these files to the tempdir, if we were okay with re-compiling each time

cmdstanpy/utils.py
```diff
@@ -1431,3 +1432,27 @@ def __enter__(self) -> Tuple[str, bool]:
...
+    digest = hashlib.sha1(code.encode()).hexdigest()
+    stan_file = os.path.join(directory, f'{digest}.stan')
+    if os.path.isfile(stan_file):
+        return stan_file
+
+    # Write the file to disk.
+    os.makedirs(directory, exist_ok=True)
```
WardBrian, line 1455: I'm concerned by this note in the doc for `makedirs`:
>Note `makedirs()` will become confused if the path elements to create include pardir (eg. “..” on UNIX systems). 

I could see someone wanting to store their stan files in `../models/` for example

WardBrian (commented):
A couple comments on the code. Thanks for putting this together

In general, I'm not sure if we reached a consensus in #513 about whether or not this is a feature we desire in CmdStanPy. @mitzimorris - what do you think?

cmdstanpy/utils.py
tillahoffmann: Makes sense. Another option could be to use `.cmdstanpy` as other packages do, e.g. `.coverage`, `.mypy_cache`, or `.pytest_cache`.

WardBrian: My philosophy is we should stray away from hiding these files in a dot folder. If users want to use it to be quick, they can, but we shouldn't hide the fact that they're doing it

mitzimorris:
I think putting stuff in tmpdir is best.
if users want to be quick and dirty, then we'll clean up after them.

cmdstanpy/utils.py
```diff
@@ -1431,3 +1432,29 @@ def __enter__(self) -> Tuple[str, bool]:
...
+
+    :param code: Code to write to a file.
+    :param directory: Directory to write the code to (defaults to the cmdstanpy
+        models directory).
+
+    :return: Path to the file containing the code..
+    """
+    # Generate the path based on the code and check whether it already exists.
+    directory = os.path.abspath(
+        directory or os.path.join(tempfile.gettempdir(), 'cmdstanpy_models')
```
WardBrian, line 1449: Cmdstanpy makes its own tempdir we use for everything - https://github.com/stan-dev/cmdstanpy/blob/683a135088eb48a8372fea9837ef96805aeaca10/cmdstanpy/__init__.py#L21
we should use it here too

tillahoffmann: Maybe slightly premature commit on my part. We can either prefix the filename (e.g. to get filenames like `model_{hash}.stan`) or have a subdirectory (as in the current commit). Any preferences?

WardBrian: I think I prefer `model_`

tillahoffmann:
Do you have any additional thoughts regarding merging/closing this PR. All previous comments should be addressed.

WardBrian:
Hi @tillahoffmann - we're currently planning a small patch release (1.0.1) to resolve some issues that have come up since the last release. We can take this back up after that, but in the mean time the code looks good to me

WardBrian:
Hi @tillahoffmann - Sorry to wait so long on this. 

I am personally still torn on adding this to the API. My sense is that more people will mis-use this than use it in the way we want, and we will end up with code like


```python
m = CmdStanModel(stan_file=write_stan_file("""
data {
  int<lower=0> N;
  int<lower=0,upper=1> y[N];
}
parameters {
  real<lower=0,upper=1> theta;
}
model {
  theta ~ beta(1,1);
  y ~ bernoulli(theta);
}
""")
```

Of course, there's nothing stopping people from doing it on their own now, but it still feels like this is encouraging putting code inside strings

tillahoffmann:
Yes, it's a tricky one. I do find it useful for prototyping because the feedback cycle is shorter, but, as you say, there is the potential for more mangled code. Having said that, there is also [some resistance to Jupyter notebooks](https://towardsdatascience.com/5-reasons-why-jupyter-notebooks-suck-4dc201e27086) because they tend to encourage bad practice (from a software development perspective), but I still find them very useful in the right context.

Let me know what you decide and I'll resolve the merge conflicts if you want to move ahead.

mitzimorris:
although I see the use case,  please let's not do this.  @tillahoffmann - this is perhaps best left as a branch or gist that people could use.

WardBrian:
Thanks @tillahoffmann for the work you put in on this, but I think I agree this kind of function may be better off living in a gist or similar, not in the API

PR #525, opened 2022-01-26
Avoid changing root logger level

WardBrian:
#### Submission Checklist

- [x] Run unit tests
- [x] Declare copyright holder and open-source license: see below

#### Summary

Closes #522. Avoids the use of `basicConfig`, which changes the root logger level. Behavior for cmdstanpy on its own should be unchanged.

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting (this will be you or your assignee, such as a university or company):

Simons Foundation

By submitting this pull request, the copyright holder is agreeing to license the submitted work under the following licenses:
- Code: BSD 3-clause (https://opensource.org/licenses/BSD-3-Clause)

PR #526, opened 2022-01-27
Revert "Avoid changing root logger level"

mitzimorris:
Reverts stan-dev/cmdstanpy#525

this seems to break things -

```
--- Logging error ---
Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/logging/__init__.py", line 1084, in emit
    stream.write(msg + self.terminator)
ValueError: I/O operation on closed file.
Call stack:
  File "/Users/mitzi/github/stan-dev/cmdstanpy/cmdstanpy/__init__.py", line 31, in _cleanup_tmpdir
    logging.getLogger('cmdstanpy').info('deleting tmpfiles dir: %s', _TMPDIR)
Message: 'deleting tmpfiles dir: %s'
Arguments: ('/var/folders/db/4jnggnf549s42z50bd61jskm0000gq/T/tmpqgyh_gr5',)
--- Logging error ---
Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/logging/__init__.py", line 1084, in emit
    stream.write(msg + self.terminator)
ValueError: I/O operation on closed file.
Call stack:
  File "/Users/mitzi/github/stan-dev/cmdstanpy/cmdstanpy/__init__.py", line 33, in _cleanup_tmpdir
    logging.getLogger('cmdstanpy').info('done')
Message: 'done'
Arguments: ()

```

### DIFF
--- a/cmdstanpy/utils.py
+++ b/cmdstanpy/utils.py
@@ -51,13 +51,7 @@ def get_logger() -> logging.Logger:
     """cmdstanpy logger"""
     logger = logging.getLogger('cmdstanpy')
     if len(logger.handlers) == 0:
-        # send all messages to handlers
-        logger.setLevel(logging.DEBUG)
-        # add a default handler to the logger to INFO and higher
-        handler = logging.StreamHandler()
-        handler.setLevel(logging.INFO)
-        handler.setFormatter(logging.Formatter(logging.BASIC_FORMAT))
-        logger.addHandler(handler)
+        logging.basicConfig(level=logging.INFO)
     return logger
 
 


mitzimorris:
we don't revert - we move forward. closing

omulayim:
> we don't revert - we move forward. closing

Hi @mitzimorris, 
Have you found a solution to this issue and that's why you gave up reverting? 
After an upgrade from CmdStanPy `v0.9.76` to `v1.0.1`, while testing our Stan models with `pytest`, we get the same pair of errors you've reported in #526.

WardBrian:
Hi @omulayim - we fixed it in #530 by using a pytest fixture to de-register the cleanup function. You can do the same thing in theory, or (if you don’t need it) remove the log handler from the cmdstanpy logger

omulayim:
Thanks @WardBrian, I believe the #530 fix applies only to CmdStanPy's own tests. However, we were getting these errors when testing our code which uses cmdstanpy. That said, by adding the following lines in our test code (as you did in [`test/conftest.py`](https://github.com/stan-dev/cmdstanpy/blob/0a2539fdc454f3c97b83fc0fc27bba5999033a57/test/conftest.py)), we were able to evade the error: 
```
import atexit
atexit.unregister(cmdstanpy._cleanup_tmpdir)
```
In any case, possibly a more rigorous way of fixing the root of the error in `cmdstanpy.__init__.py` may be found so that the above patch becomes unnecessary in third-party tests.

PR #532, opened 2022-02-03
Added passing of new stanc O options as well as validation

grburgess:
#### Submission Checklist

- [ ] Run unit tests

#### Copyright and Licensing

J. Michael Burgess

### DIFF
--- a/cmdstanpy/compiler_opts.py
+++ b/cmdstanpy/compiler_opts.py
@@ -46,6 +46,11 @@ STANC_IGNORE_OPTS = [
 ]
 
 
+VALID_OPTIM_OPTS = (0,
+                    1,
+                    "experimental")
+
+
 class CompilerOptions:
     """
     User-specified flags for stanc and C++ compiler.
@@ -165,6 +170,10 @@ class CompilerOptions:
                     self._cpp_options = {'STAN_OPENCL': 'TRUE'}
                 else:
                     self._cpp_options['STAN_OPENCL'] = 'TRUE'
+            elif key == 'O':
+                if val not in VALID_OPTIM_OPTS:
+                    raise ValueError(f'unknown optimization options {key}')
+
 
         for opt in ignore:
             del self._stanc_options[opt]
@@ -275,6 +284,9 @@ class CompilerOptions:
                     )
                 elif key == 'name':
                     opts.append(f'--name={val}')
+                elif key == 'O':
+                    opts.append(f'--O={val}')
+
                 else:
                     opts.append(f'--{key}')
         return opts
--- a/cmdstanpy/compiler_opts.py
+++ b/cmdstanpy/compiler_opts.py
@@ -172,7 +172,7 @@ class CompilerOptions:
                     self._cpp_options['STAN_OPENCL'] = 'TRUE'
             elif key.startswith('O'):
                 if has_O_flag:
-                    raise UserWarning('More than one of (O, O1, O2, Oexperimental)'
+                    get_logger().warning('More than one of (O, O1, O2, Oexperimental)'
                                       'optimizations passed. Only the last one will'
                                       'be used')
                 else:
--- a/cmdstanpy/compiler_opts.py
+++ b/cmdstanpy/compiler_opts.py
@@ -10,11 +10,9 @@ from typing import Any, Dict, List, Optional, Union
 from cmdstanpy.utils import get_logger
 
 STANC_OPTS = [
-    'O'
-    'O1',
+    'O' 'O1',
     'O2',
-    'Oexperimental'
-    'allow-undefined',
+    'Oexperimental' 'allow-undefined',
     'use-opencl',
     'warn-uninitialized',
     'include-paths',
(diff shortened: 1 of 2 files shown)


cmdstanpy/compiler_opts.py
```diff
@@ -165,6 +170,10 @@ def validate_stanc_opts(self) -> None:
                     self._cpp_options = {'STAN_OPENCL': 'TRUE'}
                 else:
                     self._cpp_options['STAN_OPENCL'] = 'TRUE'
+            elif key == 'O':
```
WardBrian, line 173: I don't think we want to pass these as key=val. `--O` is still valid all by itself for backwards compatibility (synonymous with `--Oexperimental`)

I think that the best way to implement this is just to add `O0`, `O1`, and `Oexperimental` to the list of allowed keys. If we wanted to get really fancy we could try to enforce that at most one of these is used, but even that isn't strictly necessary (the compiler will just use the final one as the definitive argument)

WardBrian (changes requested):
Thanks for catching this!

We'll probably want to add a test to the compiler options test using these.

cmdstanpy/compiler_opts.py
grburgess: Ah, rather having (key, val ) pairs. Yes, that would make more sense for a user. I can change that.

grburgess: that should fix it

cmdstanpy/compiler_opts.py
```diff
@@ -165,6 +170,13 @@ def validate_stanc_opts(self) -> None:
                     self._cpp_options = {'STAN_OPENCL': 'TRUE'}
                 else:
                     self._cpp_options['STAN_OPENCL'] = 'TRUE'
+            elif key.startswith('O'):
+                if has_O_flag:
+                    raise UserWarning('More than one of (O, O1, O2, Oexperimental)'
+                                      'optimizations passed. Only the last one will'
+                                      'be used')
```
WardBrian, line 177: We should use `get_logger().warning(...)` here to match the rest of what we do.

Besides that, I think this just needs a test or two and then it should be good. Thanks @grburgess !

grburgess: Cool. I will switch that and add the tests within an hour or so.

grburgess: Should be ready now. I tests adding more than one --O so that the logger gets covered as well.

test/test_compiler_opts.py
```diff
@@ -72,6 +72,29 @@ def test_opts_stanc(self):
...
+        opts = CompilerOptions(stanc_options=stanc_opts)
+        opts.validate()
+        self.assertEqual(
+            opts.compose(),
+            ['STANCFLAGS+=--warn-uninitialized',
+             'STANCFLAGS+=--name=foo',
+             'STANCFLAGS+=--O1'],
+             )
+
+        # should add to logger
```
WardBrian, line 85: We sometimes use LogCapture in the tests to make sure that warnings are emitted. Here's an example:
https://github.com/stan-dev/cmdstanpy/blob/2b14038979da9b6490d215221575b3592e7dae29/test/test_cmdstan_args.py#L580-L596

grburgess: everything passes locally
```base
test/test_compiler_opts.py::CompilerOptsTest::test_opts_add PASSED                       [  8%]
test/test_compiler_opts.py::CompilerOptsTest::test_opts_add_include_paths PASSED         [ 16%]
test/test_compiler_opts.py::CompilerOptsTest::test_opts_cpp PASSED                       [ 25%]
test/test_compiler_opts.py::CompilerOptsTest::test_opts_cpp_opencl PASSED                [ 33%]
test/test_compiler_opts.py::CompilerOptsTest::test_opts_empty PASSED                     [ 41%]
test/test_compiler_opts.py::CompilerOptsTest::test_opts_empty_eq PASSED                  [ 50%]
test/test_compiler_opts.py::CompilerOptsTest::test_opts_stanc PASSED                     [ 58%]
test/test_compiler_opts.py::CompilerOptsTest::test_opts_stanc_deprecated PASSED          [ 66%]
test/test_compiler_opts.py::CompilerOptsTest::test_opts_stanc_ignore PASSED              [ 75%]
test/test_compiler_opts.py::CompilerOptsTest::test_opts_stanc_includes PASSED            [ 83%]
test/test_compiler_opts.py::CompilerOptsTest::test_opts_stanc_opencl PASSED              [ 91%]
test/test_compiler_opts.py::CompilerOptsTest::test_user_header PASSED                    [100%]


```

grburgess:
ha. I religiously blacken... but didnt want to force it on others. Should be done now.

cmdstanpy/compiler_opts.py
```diff
@@ -10,11 +10,9 @@
 from cmdstanpy.utils import get_logger
 
 STANC_OPTS = [
-    'O'
-    'O1',
+    'O' 'O1',
```
WardBrian, line 13: Missing comma here it seems

cmdstanpy/compiler_opts.py
```diff
@@ -10,11 +10,9 @@
...
     'O2',
-    'Oexperimental'
-    'allow-undefined',
+    'Oexperimental' 'allow-undefined',
```
WardBrian, line 15: Here as well

WardBrian (approved):
Thanks for this!

PR #539, opened 2022-02-15
Model formatting

WardBrian:
#### Submission Checklist

- [x] Run unit tests

#### Summary

Add `model.format_model`. This calls `stanc --auto-format`, with optional arguments to 'canonicalize' the program. By default the formatted model is printed, if `save=True` is passed it is written back out to the file on disk, after first creating a copy (named `filename.stan.bak`)

#### Copyright and Licensing

Simons Foundation


cmdstanpy/model.py
```diff
@@ -297,6 +297,67 @@ def src_info(self) -> Dict[str, Any]:
...
+        try:
+            cmd = (
+                [os.path.join(cmdstan_path(), 'bin', 'stanc' + EXTENSION)]
+                # handle include-paths, allow-undefined etc
+                + self._compiler_options.compose_stanc()
+                + [self.stan_file]
+            )
+
+            if canonicalize:
+                if isinstance(canonicalize, list):
```
ahartikainen, line 333: Should we also check for tuple or collections.abc.Sequence?

cmdstanpy/model.py
```diff
@@ -297,6 +297,67 @@ def src_info(self) -> Dict[str, Any]:
...
+            result = out.stdout
+            if save:
+                if result:
+                    if not unsafe:
+                        shutil.copyfile(self.stan_file, self.stan_file + '.bak')
+                    with (open(self.stan_file, 'w')) as file:
+                        file.write(result)
+            else:
+                print(result)
+
```
ahartikainen, line 357: Printing is not very pythonic. Should it just return the result?

test/test_model.py
```diff
@@ -374,6 +377,34 @@ def test_model_includes_implicit(self):
         model2 = CmdStanModel(stan_file=stan)
         self.assertPathsEqual(model2.exe_file, exe)
 
+    def test_model_format(self):
```
ahartikainen, line 380: Pytest should skip this if not suitable cmdstan found

ahartikainen (commented):
Couple of comments

cmdstanpy/model.py
```diff
@@ -297,6 +297,67 @@ def src_info(self) -> Dict[str, Any]:
...
+            strings, those options are passed to stanc (new in Stan 2.29)
+        :param unsafe: If True, do not create stanfile.bak backups before
+            writing to the file. Only do this if you're sure you have other
+            copies of the file or are using a version control system like Git.
+        """
+        if self.stan_file is None or not os.path.isfile(self.stan_file):
+            raise ValueError("No Stan file found for this module")
+        try:
+            cmd = (
+                [os.path.join(cmdstan_path(), 'bin', 'stanc' + EXTENSION)]
```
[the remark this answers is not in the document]
WardBrian: Most of the codebase is still using os.path. I'd be nice to swap a lot of it over at some point, but for now I went with consistency

cmdstanpy/model.py
WardBrian: I think we can actually expand it all the way to `collections.abc.Iterable`

ahartikainen: I think Sequence is more suitable?

Or what where the "rules"?

Either is fine by me

WardBrian:
I want to write a few more test cases but need to update my local cmdstan to 2.29 to do so first. I'll try to do this by the end of the day

rok-cesnovar:
Sorry for jumping late, I am about to add the same thing for cmdstanr and was just wondering about naming. I know naming is hard and I know I am jumping in the last minutes, just bouncing a few ideas that come to mind before we release this feature.

- model.format_model: the second `model` seems redundant. What about just `model.format()`?
- save arg: how about `overwrite_file`? @spinkney suggested that. I don't have a preference, though overwrite_file does seem more self-explanatory
- unsafe: maybe rather than `unsafe` that defaults to False, `backup` that defaults to True?

WardBrian:
I agree on all three counts, `model.format()`, `overwrite_file` and `backup`

rok-cesnovar:
One more suggestion: exposing max_line_length?

WardBrian:
Will do that as well, hopefully tomorrow

cmdstanpy/model.py
```diff
@@ -330,25 +334,48 @@ def format_model(
...
             )
             if out.stderr:
                 get_logger().warning(out.stderr)
             result = out.stdout
-            if save:
+            if overwrite_file:
                 if result:
-                    if not unsafe:
+                    if backup:
                         shutil.copyfile(self.stan_file, self.stan_file + '.bak')
```
ahartikainen, line 379: I'm splitting hairs here, but multiple call will overwrite the older backup file?

WardBrian: Yes. We could add a random component to the string or a timestamp if we wanted

ahartikainen: Maybe timestamp?

WardBrian: Done

mitzimorris: I see the wisdom of doing this.

the potential downside is lots of backup files.  since Stan program files aren't that big, this is fine, other than visual clutter when listing a directory.   were someone to build some kind of workflow tool around CmdStanPy, this behavoir would be most useful.

WardBrian: The docstring/doc hints at the better way of doing this, which is storing your files in something like Git and then using `backup=False` to avoid the clutter. Either way, I imagine most people will probably only run this one or two times per model, probably when they see the deprecation warning, and then never again

PR #543, opened 2022-03-03
Support `.` syntax for Stan variable access

WardBrian:
#### Submission Checklist

- [x] Run unit tests

#### Summary

For all the fit objects, `fit.foo` is now a synonym for `fit.stan_variable("foo")`. This follows the default Python behavior where built-in attributes and functions are resolved before this, so if you name a variable `draws`, then `foo.draws` will still bind the function, forcing you to use `foo.stan_variable("draws")`

#### Copyright and Licensing

Simons Foundation

### DIFF
--- a/cmdstanpy/stanfit/mcmc.py
+++ b/cmdstanpy/stanfit/mcmc.py
@@ -619,7 +619,7 @@ class CmdStanMCMC:
 
     def stan_variable(
         self,
-        var: Optional[str] = None,
+        var: str,
         inc_warmup: bool = False,
     ) -> np.ndarray:
         """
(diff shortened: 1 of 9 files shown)


mitzimorris:
question:   do people also want to access method vars as well?   `draws.lp__` ?

WardBrian:
My guess is that is a far less common use case, so the extra friction of calling a function isn't as noticeable

mitzimorris:
meta-question:  is these any way to refactor the classes so that there's only one `stan_variable` method somewhere? 
can we abstract out the "drawset" object?  two flavors:  sample, point estimate?

WardBrian:
Eh, I think you quickly get into the ugly side of OOP doing that, since currently the different versions all have slightly different signatures:

```python
stan_variable(var=None, inc_warmup=False) #MCMC
stan_variable(var=None, inc_warmup=False) #GQ
stan_variable(var=None, *, inc_iterations=False, warn=True) #MLE
stan_variable(var=None) #VB
```
And even when the signature is the same (MCMC and GQ), the internals are very different due to the fact the GQ objects let you get variables either from themselves or from the MCMC fit used to create it.

mitzimorris:
thanks for reminding me why the implementation is the way it is.

mitzimorris (approved):
lgtm!

PR #556, opened 2022-05-06
macOS force arch

maedoc:
#### Submission Checklist

- [ ] Run unit tests

#### Summary

This is a WIP to force the architecture for invoking the Make toolchain on macOS, to avoid annoying errors about incompatible PCH files, when running (a) make from arm64 on a x86_64 compiled cmdstan and (b) vice versa.

This is WIP w/o tests yet to ask for initial feedback: (a) is this useful/desirable and (b) does it want full tests (i.e. building two arch copies) or can exist on a best effort basis?

#### Copyright and Licensing

### DIFF
--- a/cmdstanpy/model.py
+++ b/cmdstanpy/model.py
@@ -43,6 +43,7 @@ from cmdstanpy.utils import (
     do_command,
     get_logger,
     returncode_msg,
+    macos_make_arch_args,
 )
 
 from . import progress as progbar
@@ -413,7 +414,8 @@ class CmdStanModel:
                 'MAKE',
                 'make' if platform.system() != 'Windows' else 'mingw32-make',
             )
-            cmd = [make]
+            maybe_force_arch = macos_make_arch_args()
+            cmd = maybe_force_arch + [make]
             if self._compiler_options is not None:
                 cmd.extend(self._compiler_options.compose())
             cmd.append(Path(exe_file).as_posix())
--- a/cmdstanpy/utils.py
+++ b/cmdstanpy/utils.py
@@ -1478,3 +1478,24 @@ class SanitizedOrTmpFilePath:
     def __exit__(self, exc_type, exc_val, exc_tb) -> None:  # type: ignore
         if self._tmpdir:
             shutil.rmtree(self._tmpdir, ignore_errors=True)
+
+
+def macos_make_arch_args():
+    """Determine what, if any, `arch -arm64` prefix for the `make` invocation."""
+    if sys.platform != 'darwin':
+        return []
+    pch_name = 'stan/src/stan/model/model_header.hpp.gch'
+    pch_path = f'{cmdstan_path()}/{pch_name}'
+    with tempfile.TemporaryDirectory() as td:
+        with open(f'{td}/foo.cpp', 'w') as fd:
+            fd.write('int foo() { }')
+        out = subprocess.Popen(
+            ['g++', '-include-pch',  pch_path, '-c', 'foo.cpp'],
+            cwd=td, stderr=subprocess.PIPE
+        )
+        err = out.stderr.read().decode('ascii')
+    match = re.search(r"was compiled for the target '(.*?)-", err, re.MULTILINE)
+    if match:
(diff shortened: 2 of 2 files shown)


WardBrian:
I think a best-effort basis is what we've done for some similar hackery on Windows. It's almost impossible to test this in CI.

Is there no way to do this other than trying to run `g++` on something (also, shouldn't this be clang, or better, `os.environ.get('CXX', 'clang')`?

maedoc:
I thought pch files were Clang and gch files were GCC, and I found a gch file, but yes I can test.  I couldn’t file a cli tool to dump the target of a pch file, so I resorted to invoking it directly.  I’ll look at checking bin/stansummary or similar since it’s also arch dependent (while stanc is x86 in any case I think).

I’ll get it looking a bit nicer and tested locally at least and then undraft the PR.

bob-carpenter:
Hi, @maedoc.  I just wanted to say thanks again, given that we started from your cmdstan wrapper.

Of course, we'll be eager to merge anything that reduces the number of PCH errors we get.

maedoc:
I've simplified the arch test by just checking what `file bin/stansummary` says (either `.Mach-O 64-bit executable x86_64` or `Mach-O 64-bit executable arm64`), and here's it working in my case, calling from an arm64 kernel to a x86_64 compiled CmdStan, 
image

WardBrian:
Yeah, unsure how to handle this. I think a simple check that says if the arch of the compiler doesn't match the arch of stansummary, fail with a nice error message? This should only trigger if `CXX` is set, I think

maedoc:
Actually, if `make` is the only command being invoked directly, then setting `CXX` to the wrong arch won't be a problem: `make` itself is a so-called universal binary with both architectures (as is the platform `c++`), and then subprocess calls from make handle the invocation across architectures w/o problem. To wit, a test makefile like this
```make
all:
	file $(shell which $(CXX))
	$(CXX) -v
```
with invocations like
- `make`: defaults to arm64
- `arch -x86_64 make`: make runs as x86_64, c++ is invoked as x86_64
- `arch -arm64 make`: make runs as arm64, c++ is invoked as arm64
- `arch -x86_64 make CXX=/opt/homebrew/bin/g++-11`: make runs as x86_64, invokes an arm64 version of g++
- `arch -arm64 make CXX=/usr/local/bin/g++-11`: make runs as arm64, invokes an x86_64 version of g++
all produce the right output.

In summary, I think the implemented approach is sound and macOS' behavior is kind-of sane, and so we avoid just the issue of a mismatch between the compiled CmdStan arch and the implicit arch when invoking make from Python.

WardBrian:
I guess the concern is if they installed cmdstan with something set for CXX that is arm64-based, and then later are trying to use a CXX which is x86, won't that lead to the PCH error this PR is trying to avoid?

WardBrian:
It seems like an easy thing to detect and (at least) issue a warning about, though. Would you mind adding that?

maedoc:
Sure, can you link to an example? I couldn't find anything in the codebase which shows how an CXX is specified.  I thought it'd be `cpp_options` but it's typed as `Dict[str, Union[bool, int]]`.

WardBrian:
I'm seeing the definition of `cpp_options` as `Dict[str,Any]` - where are you seeing that other type? It is wrong if it is in the code somewhere.

I think we need to check both `cpp_options` and `os.environ('CXX')`, in cases where the user did something like 
`CXX=something python` or is using a conda environment which can set `CXX`

maedoc:
> where are you seeing that other type?

[Here](https://github.com/stan-dev/cmdstanpy/blob/develop/cmdstanpy/compiler_opts.py#L105) for instance.  Is that a typo or it's something else?  I can add the check for env var as well yes.

WardBrian:
Yeah that is incorrect. I think the semantics of mypy’s typing mean that it doesn’t notice that because elsewhere it is declared to be Any

WardBrian:
I’m not sure if we have a cmdstanpy dev with an M1/2 machine, so happy to wait. We can push this out of the 1.0.2 milestone

PR #571, opened 2022-06-15
missing import in Quickstart to installation

mjhajharia:
os was not imported that code snippet won't work without it


WardBrian (approved):
Thanks for this!

Issue #574, opened 2022-06-16
couple of errors in mcmc notebook


mjhajharia:
Here's an example of a cell which isn't executed correctly https://cmdstanpy.readthedocs.io/en/v1.0.1/examples/MCMC%20Sampling.html#Running-a-data-generating-model CELL 12. making a PR to add fix soon!

Cause: range is an iterator, you need to do list(range) and futhermore in this case (I think, correct me otherwise please) range isn't needed at all? like for the bins we can just give datagen['N']=300 right?

if my assumption is right `fit_sim.draws_pd()['theta_rep'].plot.hist(bins=datagen_data['N'])` should suffice for the entire cell?

mitzimorris:
what we're trying to plot is the cumulative number of successes in 10 trials across the 300 draws from the posterior.

Issue #555, opened 2022-05-04
add ability to report max treedepth and divergences


mitzimorris:
#### Summary:
instead of running `diagnose` command to get treedepth warnings and divergence counts, these can just be summed out of column of sample; cf `cmdstanr` functionality

#### Description:
Describe the issue as clearly as possible.

#### Additional Information:
Provide any additional information here.

mitzimorris:
CmdStanr currently reports divergences, max_treedepth, and ebfmi (estimate of Bayesian Fraction of Missing Information)
https://mc-stan.org/cmdstanr/reference/fit-method-diagnostic_summary.html

these 3 numbers are easily derived from the HMC diagnostics reported by the sampler.

column `divergent__` flags iterations where the sampler failed to find a new point to jump to - this is properly an int value, either 0 or 1.

column `treedepth__` is the number of doublings of the number of leapfrog steps - this is properly an int value.

column `energy__` is explained here:  https://discourse.mc-stan.org/t/definition-of-the-energy-term/20019/3
> the energy__ variable is precisely the value of the Hamiltonian function at the state sampled for the given iteration. It is meant to probe how well the sampler is exploring between Hamiltonian level sets 

calculating `divergences` and `max_treedepth` is pretty easy:

```
vars = fit.method_variables()
max(vars['treedepth__'].max(axis=0))
sum(vars['divergent__'].sum(axis=0))
```

`eBFMI` is explained here: https://mc-stan.org/docs/cmdstan-guide/diagnose.html#low-e-bfmi-values---sampler-transitions-hmc-potential-energy.

diagnosing low eBFMI requires a comparison between the standard deviation of `energy__` across all chains to  `sqrt(D * 0.5)` where D is the number of *unconstrained* parameters.  At the point we have the sample, we can get D from the metric information.  therefore, we also need to store `sqrt(0.5 * D)` - and find a good name for it - "energy_threshold" or similar?
then property `low_eBFMI` would report `eBFMI > energy_threshold`.

it makes sense to compute `divergences`, `max_treedepth` and `eBFMI` when reading in the sample and they should be exposed as properties on the `CmdStanMCMC` object.

WardBrian:
I assume cmdstanr issues some warnings based of these numbers if they are too high or too low?

mitzimorris:
https://mc-stan.org/cmdstanr/reference/model-method-sample.html

[...]

PR #277, opened 2020-08-12
Feature/263 cmdstanr harmonize

mitzimorris:
#### Submission Checklist

- [x] Run unit tests
- [x] Declare copyright holder and open-source license: see below

#### Summary
Deprecate CmdStanMCMC object properties `warmup` and `sample` and introduce
method `draws(inc_warmup: bool = False)`.

Update docs.

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting (this will be you or your assignee, such as a university or company):  Columbia University

By submitting this pull request, the copyright holder is agreeing to license the submitted work under the following licenses:
- Code: BSD 3-clause (https://opensource.org/licenses/BSD-3-Clause)

PR #136, opened 2019-10-11
Docs/updates

mitzimorris:
#### Submission Checklist

- [x] Run unit tests
- [x] Declare copyright holder and open-source license: see below

#### Summary

Updating docs, bugfix and name tweaks

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting (this will be you or your assignee, such as a university or company):  Columbia University

By submitting this pull request, the copyright holder is agreeing to license the submitted work under the following licenses:
- Code: BSD 3-clause (https://opensource.org/licenses/BSD-3-Clause)

PR #575, opened 2022-06-16
Update MCMC Sampling.ipynb

mjhajharia:
See #574 I've only changed the last cell and re-ran the notebook

### DIFF
--- a/docsrc/examples/MCMC Sampling.ipynb
+++ b/docsrc/examples/MCMC Sampling.ipynb
@@ -39,9 +39,25 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
-   "metadata": {},
-   "outputs": [],
+   "execution_count": 1,
+   "metadata": {
+    "execution": {
+     "iopub.execute_input": "2022-02-14T15:59:14.811548Z",
+     "iopub.status.busy": "2022-02-14T15:59:14.810746Z",
+     "iopub.status.idle": "2022-02-14T15:59:15.026743Z",
+     "shell.execute_reply": "2022-02-14T15:59:15.025694Z"
+    }
+   },
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Enabling notebook extension jupyter-js-widgets/extension...\n",
+      "      - Validating: \u001b[32mOK\u001b[0m\n"
+     ]
+    }
+   ],
    "source": [
     "!jupyter nbextension enable --py widgetsnbextension"
    ]
@@ -74,9 +90,24 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
-   "metadata": {},
-   "outputs": [],
+   "execution_count": 2,
+   "metadata": {
+    "execution": {
+     "iopub.execute_input": "2022-02-14T15:59:15.035939Z",
+     "iopub.status.busy": "2022-02-14T15:59:15.034474Z",
+     "iopub.status.idle": "2022-02-14T15:59:15.391468Z",
+     "shell.execute_reply": "2022-02-14T15:59:15.390919Z"
+    }
+   },
+   "outputs": [
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "INFO:cmdstanpy:found newer exe file, not recompiling\n"
+     ]
+    }
+   ],
    "source": [
     "import os\n",
     "from cmdstanpy import CmdStanModel, cmdstan_path\n",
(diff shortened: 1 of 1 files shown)


WardBrian:
The change is great, do you mind clearing the outputs of the notebook? The versions we check in are always blank, they get run during the docs build

mjhajharia:
> The change is great, do you mind clearing the outputs of the notebook? The versions we check in are always blank, they get run during the docs build

you got it!

mjhajharia:
> The result is different from the intended image (for some reason, the code works totally fine for me locally despite not building in the docs)
> 
> Current output: image
> 
> This PR:
> 
> image
> 
> I think we should use either `list(range(...` or `np.arange` to get the intended binning

which python version are you on?
`fit_sim.draws_pd()['theta_rep'].plot.hist(list(range(1,datagen_Data['N']+1)))` throws ValueError: Grouper for '0' not 1-dimensional for me

mjhajharia:
and `y_sums = y_sims.sum(axis=1)` is doing nothing, its just converting the dataframe to a series, was DataFrame.cumsum meant to use here?

WardBrian:
I believe you need to name the argument with `bins=0`, the first positional argument of `.plot.hist` is used for something different

WardBrian:
Yeah, the original code is kind of messy. I believe that `sum(axis=1)` is functioning like a `squeeze()` - before hand the object has shape `(1000, 1)`, afterwards the shape is just `(1000, )`

mjhajharia:
> I believe you need to name the argument with `bins=`, the first positional argument of `.plot.hist` is used for something different

that would mean 300 bins? i think @mitzimorris said we wanted 10?

mjhajharia:
> Yeah, the original code is kind of messy. I believe that `sum(axis=1)` is functioning like a `squeeze()` - before hand the object has shape `(1000, 1)`, afterwards the shape is just `(1000, )`

yeah no i think it might be old code, currently the .plot.hist is i think a pandas wrapper for matplotlib so it can directly access dataframes. and we could just use by="theta_rep" instead of removing the columns.

WardBrian:
We also have the ability to request only specific variables from `draws_pd`, so in the interest of both the cleanest code and displaying the greatest number of features I would have the cell be 

```python
fit_sim.draws_pd('theta_rep').plot.hist()
```

mjhajharia:
> We also have the ability to request only specific variables from `draws_pd`, so in the interest of both the cleanest code and displaying the greatest number of features I would have the cell be
> 
> ```python
> fit_sim.draws_pd('theta_rep').plot.hist()
> ```

agreed, also i get where the code is coming from now -> https://cmdstanpy.readthedocs.io/en/stable-0.9.65/sample.html#example-generate-data-fixed-param-true check the end here,  this was pre- `draw_pd()` so they had to convert it to a dataframe. still makes no sense because axis=1 for an array????

mjhajharia:
> Besides the code change, there's a bunch of extra metadata being added by this PR. Are you running these in Jupyter or something like VSCode/Pycharm? Do you have something like `nbstripout` installed to remove it? It's also possible to open in just a normal text editor and edit as json
> 
> I don't think there is necessarily any harm in it, but it would be nicer if it wasn't in the repo

yeah i'll do that no worries

mitzimorris:
ignore my previous comment.  all I can say is whenI wrote this, I wasn't particularly fluent in Python and more stuff was added to API later which made this a whole lot easier.  whatever you do to clean this up would be great.

mjhajharia:
> ignore my previous comment. all I can say is whenI wrote this, I wasn't particularly fluent in Python and more stuff was added to API later which made this a whole lot easier. whatever you do to clean this up would be great.

no worries, i think the issue is that the code in the .stan file has changed from the original code in 0.9, and the code in the notebook is the same, I'll try to get it working tonight

WardBrian:
I’m still not sure what row wise sums mean in this context, since theta_rep in the model is a scalar

mjhajharia:
mitzi is right, as i said, in the original stan file(check release 0.9’s
file instead of theta ‘y’ was used which was an N dimensional vector so
you’d have theta_rep[0] - theta_rep[10]

mitzimorris:
@mjhajharia -  you mean 'y_rep[1] ... y_rep[10]', not 'theta_rep[0] ... theta_rep[10]' ?

Stan indexes from 1 (like a mathematician). whereas Python indexes from 0 (like a computer scientist).

'y_rep'  is the posterior predictive check - we use the estimate of 'theta' to generate new observations of y.

WardBrian:
Ah, so at some point the model used here was changed to output only one value, which is already the sum of the the n trials. That makes sense as to why the post processing code looks like it did, but also why we can now do this in basically one line without any extra sums or anything. 

It looks like this file also tries to link to the Stan code used but the link fails. We should probably just do what we do in other notebooks and add a cell which is print(model.code()) to show it rather than a link which might break

mitzimorris:
what model are we talking about - the model should be this one:  https://github.com/stan-dev/cmdstanpy/blob/develop/docsrc/examples/bernoulli_ppc.stan

mjhajharia:
> what model are we talking about - the model should be this one: https://github.com/stan-dev/cmdstanpy/blob/develop/docsrc/examples/bernoulli_ppc.stan

this is the model the notebook takes as input, i think that's what causing issues?

mitzimorris:
that datagen model makes no sense.

mjhajharia:
> that datagen model makes no sense.

yep

WardBrian:
`bernoulli_ppc` is used as the example in the "Generated Quantities" notebook, but as far as I can tell the MCMC notebook has always used `bernoulli_datagen` since it was added in https://github.com/stan-dev/cmdstanpy/pull/136

mjhajharia:
> `bernoulli_ppc` is used as the example in the "Generated Quantities" notebook, but as far as I can tell the MCMC notebook has always used `bernoulli_datagen` since it was added in #136

yeah the original version of the datagen one was similar to the ppc one and then at some point some editing happened, either way i don't understand the stan model or purpose but if @mitzimorris can fix that i can definitely clean up the python part in the notebook

WardBrian:
Yeah, looks like it changed in https://github.com/stan-dev/cmdstanpy/pull/277 

Is having a data generating model even a necessity for this notebook now that the standalone GQ notebook exists?

mitzimorris:
the docs have expanded organically.  there's a tension between lightweight user's guide and worked examples and case studies, etc.   this was an issue in the beginning because we had to have "fixed_param" set.  now that's not necessary,
so the best thing to do is get rid of the datagen model here.

a datagen model belongs in its own section on prior predictive checks, SBC, etc.
I did a meetup a long time about about poisson vs neg binomial - that would be a reasonable example.

mitzimorris:
@mjhajharia and @WardBrian - this notebook really needs cleaning up.
I started adding more to this as part of #555 - if it's OK with you, I'll continue doing so on that branch (https://github.com/stan-dev/cmdstanpy/tree/feature/555-compute-diagnostics)

mitzimorris:
show off diagnostics.

PR #577, opened 2022-06-17
Feature/555 compute diagnostics

mitzimorris:
#### Submission Checklist

- [x] Run unit tests

#### Summary

Adds logic to check for divergences and iterations which hit max treedepth and automatically prints warning when any are found.   Adds unit tests, updates documentation.

Checking is done during CSV file validation; given that we're already doing a serious ad-hoc parse of the output files, adding more logic to check 2 columns is just adding a few more straws to that camel's back.

In documenting this, I cleaned up the example notebook for MCMC sampling.  My initial idea was to add this to the "Hello World" doc page, and therefore did a little cleanup on that doc page.  In the end, I decided that explaining divergences and treedepth is beyond the scope of "Hello World" and instead put it into the MCMC sampling example notebook.

#### Copyright and Licensing


cmdstanpy/stanfit/mcmc.py
```diff
@@ -78,31 +77,51 @@ def __init__(
...
         self._sig_figs = runset._args.sig_figs
+
         # info from CSV values, instantiated lazily
+        self._draws: np.ndarray = np.array(())
+        # only valid when not is_fixed_param
         self._metric: np.ndarray = np.array(())
         self._step_size: np.ndarray = np.array(())
-        self._draws: np.ndarray = np.array(())
+        self._divergences: np.ndarray = np.zeros(self.runset.chains, int)
+        self._max_treedepths: np.ndarray = np.zeros(self.runset.chains, int)
```
WardBrian, line 99: For the dtype argument to something like np.array or np.zeros my personal preference is to always name it like `dtype=int`

cmdstanpy/utils.py
```diff
@@ -957,9 +961,21 @@ def scan_sampling_iters(
...
+    idx_treedepth = None
+    max_treedepth = None
+    ct_divergences = 0
+    ct_max_treedepth = 0
+    try:
+        idx_divergent = config_dict['column_names'].index('divergent__')
+        idx_treedepth = config_dict['column_names'].index('treedepth__')
+        max_treedepth = config_dict['max_depth']
+    except ValueError:
+        pass
```
WardBrian, line 978: I'm curious about this logic - under what condition would this exception be thrown?

Is there any way to avoid then checking if `max_treedepth` is None for every line as we iterate through?

cmdstanpy/stanfit/mcmc.py
```diff
@@ -78,31 +77,51 @@ def __init__(
...
-            self._thin = thin
+        self._iter_sampling:int = _CMDSTAN_SAMPLING
+        if sampler_args.iter_sampling is not None:
+            self._iter_sampling = sampler_args.iter_sampling
+        self._iter_warmup:int = _CMDSTAN_WARMUP
+        if sampler_args.iter_warmup is not None:
+            self._iter_warmup = sampler_args.iter_warmup
+        self._thin:int = _CMDSTAN_THIN
+        if sampler_args.thin is not None:
+            self._thin = sampler_args.thin
```
WardBrian, line 88: [Comment] this is great

cmdstanpy/stanfit/mcmc.py
```diff
@@ -78,31 +77,51 @@ def __init__(
...
         self._metric: np.ndarray = np.array(())
         self._step_size: np.ndarray = np.array(())
-        self._draws: np.ndarray = np.array(())
+        self._divergences: np.ndarray = np.zeros(self.runset.chains, int)
+        self._max_treedepths: np.ndarray = np.zeros(self.runset.chains, int)
+
         # info from CSV initial comments and header
         config = self._validate_csv_files()
         self._metadata: InferenceMetadata = InferenceMetadata(config)
+        # prelim diagnostics
```
WardBrian, line 104: How would you feel about putting this in a function? something like `self._report_diagonostics()`

test/test_sample.py
```diff
@@ -1795,6 +1795,71 @@ def test_attrs(self):
...
+                show_progress=False,
+                show_console=False,
+            )
+
+            self.assertTrue(
+                all([a == b for a, b in zip(fit.max_treedepths, [0, 0, 0, 6])])
+            )
+            self.assertTrue(
+                all([a == b for a, b in zip(fit.divergences, [10, 143, 5, 4])])
+            )
```
WardBrian, line 1818: can we use `self.assertListEqual(fit.max_treedepths.as_list(), [0,0,0,6])` or something similar?

I also worry that this test will be a bit brittle in terms of cmdstan changes. Maybe we should instead assert that `fit.max_treedepths.sum()` is greater than some small number?

test/test_sample.py
```diff
@@ -1795,6 +1795,71 @@ def test_attrs(self):
...
+
+            log.check_present(
+                (
+                    'cmdstanpy',
+                    'WARNING',
+                    'Some chains may have failed to converge.\n'
```
WardBrian, line 1824: Similarly, maybe we should just test that this is output, and match the rest of the message with a wildcard.

WardBrian (changes requested):
This looks good. Some preliminary comments before the long weekend, I'll look more at the doc changes next week

cmdstanpy/utils.py
mitzimorris: this code is generalized for samples which could be generated by "fixed_param" (program has no model block) or standalone generated quantities (model block isn't executed).

we could make 2 methods - one which expects sample diagnostics and one which doesn't - would that be cleaner?

test/test_sample.py
mitzimorris: indeed it was brittle.  fixed.

test/test_sample.py
mitzimorris: you'd think wildcard was possible, but that's not how `log.check` and `log.check_present` work.
found a workaround.

WardBrian: In other places we’ve used the StringComparison class from the same test utilities, I think that would be less brittle if another warning ever pops up

cmdstanpy/utils.py
WardBrian: I believe the function which calls this takes in an argument indicating that it is fixed_param - it would at least be cleaner if this function did too, even if that doesn’t fix check-per-loop issue. Splitting it into two functions wouldn’t be too bad either

test/test_sample.py
mitzimorris: thanks - much better!

cmdstanpy/utils.py
mitzimorris: refactored.

mitzimorris:
ready for re-review.

WardBrian (commented):
A few minor comments/questions. I also reviewed the docs, but they both look great, so no comments there.

WardBrian (approved):

PR #583, opened 2022-06-29
Ensure stanfit objects can be deepcopied and pickled

WardBrian:
#### Submission Checklist

- [x] Run unit tests

#### Summary

The change in #543 broke the ability to pickle models. Because pickling does not call `__init__` on object, its internal state may not be 100% correct. 

This led to an infinite recursion in our `__getattr__` logic: `__getattr__` would call `stan_variable`, which would check `self._metadata`, which would call `__getattr__`. Luckily it is easy to fix this, since Stan variable names are forbidden from starting with an _

This is important for downstream packages like prophet, and sktime: https://github.com/alan-turing-institute/sktime/issues/2898

As a result I would like to release 1.0.3 as soon as this is merged.

#### Copyright and Licensing


test/test_optimize.py
```diff
@@ -627,6 +627,28 @@ def test_attrs(self):
...
+        """Ensure fit objects are pickle-able and copy-able"""
+
+        import pickle
+
+        csvfiles_path = os.path.join(
+            DATAFILES_PATH, 'optimize', 'rosenbrock_mle.csv'
+        )
+        fit = from_csv(path=csvfiles_path)
+        pickled = pickle.dumps(fit)
+        unpickled = pickle.loads(pickled)
```
ahartikainen, line 640: should we add fit delete step here (if the unpickled would reference something from fit?)

ahartikainen (approved):
I added only a small comment, but otherwise looks good

aiwalter:
just linking our issue :)
https://github.com/alan-turing-institute/sktime/issues/2898

WardBrian:
Seems like GHA is down at the moment (https://www.githubstatus.com/) so CI isn't running on this. Hopefully it's up soon so we can merge and release a patch.

I moved these tests to their own folder. It seems pickling is weird and can be highly dependent on module import order, so I'm hoping putting the tests in their own file resolves the error the 3.7 runners were raising before.

PR #537, opened 2022-02-10
Properly handle complex outputs

WardBrian:
#### Submission Checklist

- [x] Run unit tests
- [x] Declare copyright holder and open-source license: see below

#### Summary

Closes #536 

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting (this will be you or your assignee, such as a university or company):
Simons Foundation, Columbia University

By submitting this pull request, the copyright holder is agreeing to license the submitted work under the following licenses:
- Code: BSD 3-clause (https://opensource.org/licenses/BSD-3-Clause)

PR #612, opened 2022-08-24
Fix/complex output ordering

WardBrian:
#### Submission Checklist

- [x] Run unit tests

#### Summary

First reported by @tillahoffmann on the [forums here](https://discourse.mc-stan.org/t/difference-in-fft-implementation-between-stan-and-numpy-scipy/28642), CmdStanPy was mangling complex-valued outputs.

The original PR #537 did not include a test asserting that the actual values were recovered in the correct order, which this also fixes. The current code that complex numbers in the output are stored like an extra final dimension of size 2 (this is, after all, how they are treated in the program _inputs_). But they’re not really, they’re stored strided alongside the other dimensions.

To make this concrete, a vector of 3 elements was being read in as if it was stored as `x.1.real, x.2.real, x.3.real, x.1.imag. x.2.imag, x.3.imag`, but really it is actually written to disk as `x.1.real, x.1.imag, x.2.real, x.2.imag, ...`

I think once this is merged we should release a patch version (1.0.7) almost immediately. 

#### Copyright and Licensing

Simons Foundation

### DIFF
--- a/cmdstanpy/stanfit/gq.py
+++ b/cmdstanpy/stanfit/gq.py
@@ -523,10 +523,15 @@ class CmdStanGQ:
             col_idxs = self._metadata.stan_vars_cols[var]
             if len(col_idxs) > 0:
                 dims.extend(self._metadata.stan_vars_dims[var])
-            # pylint: disable=redundant-keyword-arg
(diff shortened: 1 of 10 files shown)


mitzimorris (approved):
great!  thanks for fixing!

PR #616, opened 2022-09-14
Download correct CmdStan tarball on non-x86 linux

### LINKED ISSUE #615: Problem when compiling on ARM
#### Summary:

Compilation on ARM do not work properly: the version option is not enough to download the correct cmdstan release file.

#### Description:

In our deployment process we install facebook "prophet" (1.1) which depends on cmdstanpy 1.0.7.
The deployment is done on an ARM architecture.

The process stops with the error: 

```
--- Translating Stan model to C++ code ---
        bin/stanc  --o=examples/bernoulli/bernoulli.hpp examples/bernoulli/bernoulli.stan
        bash: line 1: bin/stanc: cannot execute binary file: Exec format error
```

which indicate an architecture problem.

After some investigation it seems that the url to download the cmdstan release is not correct for ARM, here: https://github.com/stan-dev/cmdstanpy/blob/e885cb0edf2e611802cb36f491fa05192d4190b3/cmdstanpy/install_cmdstan.py#L461

```
url = (
        'https://github.com/stan-dev/cmdstan/releases/download/'
        'v{0}/cmdstan-{0}.tar.gz'.format(version)
    )
```
This url construct do not take into account the architecture of the target which is: 

```
https://github.com/stan-dev/cmdstan/releases/download/v2.26.1/cmdstan-2.26.1-linux-arm64.tar.gz
```

The Ruby wrapper for cmdstan (https://github.com/ankane/cmdstan-ruby) fixed this problem in that commit: https://github.com/ankane/cmdstan-ruby/commit/7a6a38437c7664c8021190bf4e4e052d6d120035

Would it be possible to use the same fix with cmdstanpy ?
Thanks !

#### Additional Information:

Everything works fine when installing on amd64 architecture.

#### Current Version:

- cmdstanpy version: 1.0.7
- cmdstan version: 2.26.1

WardBrian:
#### Submission Checklist

- [x] Run unit tests

#### Summary

Closes #615. This PR changes the URL used in `install_cmdstan` to try to use the specific non-x86 linux binaries when applicable. It does its best to determine this automatically based on the value of `platform.machine()`. 

Users can specify an environment variable, `CMDSTAN_ARCH`, to override the default choice

#### Copyright and Licensing

Simons Foundation

### DIFF
--- a/cmdstanpy/utils/cmdstan.py
+++ b/cmdstanpy/utils/cmdstan.py
@@ -48,7 +48,7 @@ def get_download_url(version: str) -> str:
     if not arch and platform.system() == "Linux":
         arch = determine_linux_arch()
 
-    if arch:
+    if arch and arch.lower() != "false":
         url_end = f'v{version}/cmdstan-{version}-linux-{arch}.tar.gz'
     else:
         url_end = f'v{version}/cmdstan-{version}.tar.gz'
--- a/cmdstanpy/utils/cmdstan.py
+++ b/cmdstanpy/utils/cmdstan.py
@@ -3,6 +3,7 @@ Utilities for finding and installing CmdStan
 """
 import os
 import platform
+import subprocess
 import sys
 from collections import OrderedDict
 from typing import Callable, Dict, Optional, Tuple, Union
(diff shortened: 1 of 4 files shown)


cmdstanpy/utils/cmdstan.py
```diff
@@ -17,6 +18,44 @@
...
+    machine = platform.machine()
+    arch = ""
+    if machine == "aarch64":
+        arch = "arm64"
+    elif machine == "armv7l":
+        # Telling armel and armhf apart is nontrivial
+        # c.f. https://forums.raspberrypi.com/viewtopic.php?t=20873
+        if subprocess.run(
+            ["readelf -A /proc/self/exe | grep Tag_ABI_VFP_args"],
+            shell=True,
```
ahartikainen, line 31: Why is the shell=True needed here?

cmdstanpy/utils/cmdstan.py
```diff
@@ -17,6 +18,44 @@
...
+            check=False,
```
ahartikainen, line 32: Does this command also fail sometimes?

ahartikainen (approved):
Couple comments but otherwise LGTM

cmdstanpy/utils/cmdstan.py
WardBrian: `grep` sets the return code to 1 when it doesn’t find anything (which is why we’re branching on it), so using check would be a problem 

I don’t believe it would fail, since readelf and grep are both standard Unix tools which should be available

cmdstanpy/utils/cmdstan.py
WardBrian: It wasn't, refactored to avoid

jbjuin:
@WardBrian wouah ! That was fast :) Perfect for me ! Thanks a lot.

WardBrian:
Are you able to confirm that this resolved the issue by installing from GitHub? After that it should be reasonably quick to release I think

jbjuin:
Hi @WardBrian, sorry for the late reply.
We had other problems with ortools installation on this architecture and asked our partner to switch to amd64 machines.
I didn't check your modifications in that context. I may have a raspberrypi at hand to check this. I'm planning one hour within a month to try it. Thanks a lot for your fast solution and all the work on cmdstan.

PR #621, opened 2022-09-21
Add timeout parameter.

tillahoffmann:
#### Submission Checklist

- [x] Run unit tests (on macOS; will have to wait for Windows and Linux in CI)

#### Summary

This PR adds a `timeout` parameter to `optimize`, `sample`, `generate_quantities`, and `variational` including unit tests. I was running hyperparameter sweeps for a model and some configurations took much longer than others. Setting a timeout means that "bad" configurations can be aborted if they take too long.

I used a `threading.Timer` to terminate the cmdstan subprocess. Checking the elapsed time in `_run_cmdstan` didn't work because the `readline` call blocks until the next line of `stdout` becomes available (which can take a while if the model is sampling inefficiently).

#### Copyright and Licensing

Harvard University

### DIFF
--- a/cmdstanpy/model.py
+++ b/cmdstanpy/model.py
@@ -733,6 +733,7 @@ class CmdStanModel:
                 show_console=show_console,
                 timeout=timeout,
             )
+        runset.raise_for_timeouts()
 
         if not runset._check_retcodes():
             msg = 'Error during optimization: {}'.format(runset.get_err_msgs())
@@ -1167,6 +1168,8 @@ class CmdStanModel:
                 sys.stdout.write('\n')
                 get_logger().info('CmdStan done processing.')
 
+            runset.raise_for_timeouts()
+
             get_logger().debug('runset\n%s', repr(runset))
 
             # hack needed to parse CSV files if model has no params
@@ -1348,6 +1351,7 @@ class CmdStanModel:
                         timeout=timeout,
                     )
 
+            runset.raise_for_timeouts()
             errors = runset.get_err_msgs()
             if errors:
                 msg = (
@@ -1518,6 +1522,7 @@ class CmdStanModel:
                 show_console=show_console,
                 timeout=timeout,
             )
+        runset.raise_for_timeouts()
 
         # treat failure to converge as failure
         transcript_file = runset.stdout_files[dummy_chain_id]
@@ -1606,7 +1611,15 @@ class CmdStanModel:
                 universal_newlines=True,
             )
             if timeout:
-                timer = threading.Timer(timeout, proc.terminate)
+
+                def _timer_target():
+                    # Abort if the process has already terminated.
+                    if proc.poll() is not None:
+                        return
+                    proc.terminate()
+                    runset._set_timeout_flag(idx, True)
+
+                timer = threading.Timer(timeout, _timer_target)
                 timer.setDaemon(True)
                 timer.start()
             else:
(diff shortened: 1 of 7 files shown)


test/test_variational.py
```diff
@@ -291,6 +291,14 @@ def test_attrs(self):
         with self.assertRaisesRegex(AttributeError, 'Unknown variable name:'):
             dummy = fit.c
 
+    def test_timeout(self):
+        stan = os.path.join(DATAFILES_PATH, 'timeout.stan')
+        timeout_model = CmdStanModel(stan_file=stan)
+        self.assertRaisesRegex(
```
WardBrian, line 297: Would you mind reorganizing these tests so they use the `with self.assertRaises...` style?

cmdstanpy/model.py
```diff
@@ -1579,6 +1596,12 @@ def _run_cmdstan(
                 env=os.environ,
                 universal_newlines=True,
             )
+            if timeout:
+                timer = threading.Timer(timeout, proc.terminate)
```
WardBrian, line 1600: Should we later be calling `timer.cancel()` if the task finishes before this timeout hits?

WardBrian (changes requested):
This is a nice idea, thanks for looking into it. 

Do you have the `black` formatter installed? Some of the changes look like they need to be formatted

cmdstanpy/model.py
tillahoffmann: Ah neat, didn't know about that function.

tillahoffmann:
I just raised `RuntimeError` for the time being. Thoughts on raising `TimeoutError` instead so it's easier to catch the exception?

WardBrian:
That would certainly be clearer, but I assume it has some difficulties with our `check_returncode`/`get_err_msg` functionality?

tillahoffmann:
I'm currently manually patching the return code to 60 as `os.strerror(60) == 'Operation timed out'` if the timer terminates the subprocess. So we could raise a `TimeoutError` based on the return code?

WardBrian:
Sounds good to me

WardBrian:
I suppose the other thing you could do besides patch the return code is have the Timer function both terminate the proc _and_ indicate it did this by setting an attribute in runset. I'm mostly indifferent, especially if we think that `-15` as a return code would never show up normally

tillahoffmann:
I've tried it with storing state in `runset` which might also fix the failing Windows tests.

tillahoffmann:
Alright, these seem to be passing now.

As a timing-related thing, what are your thoughts on adding a `time.sleep(1e-3)` in the `while proc.poll() is None` loop [here](https://github.com/stan-dev/cmdstanpy/pull/621/files#diff-af39d3679809ef0504a5e945eb5588b237ae0c42c92631c0c2f626b14edb414dL1582-L1591)? That would reduce the CPU overhead from the busy waiting a bit. The `subprocess.Popen.wait` method [uses a sleep of 1 ms](https://github.com/python/cpython/blob/8a221a853787c18d5acaf46f5c449d28339cde21/Lib/subprocess.py#L1961), for example.

WardBrian:
Seems reasonable, but looking at that code it seems like it will do the wrong thing if more than one line of output is buffered at a time (which is bad)? So we would also definitely want to do `readlines` instead of `readline`, and process any potential output in a loop, I think.

WardBrian (approved):
This looks good to me. Do you want to look at the proc.poll issue in a different PR? If so I will merge this

tillahoffmann:
Makes sense; maybe I'll have a look at that in a separate PR.

tillahoffmann:
As a follow up, the issue of busy-waiting is not an issue: `readline` blocks when there's no content so the loop is only busy when there's actually stuff to do.

PR #623, opened 2022-09-23
Fix json encoding of NaN/infinity

WardBrian:
#### Submission Checklist

- [x] Run unit tests

#### Summary

We implemented our json encoding based on the spec given in the Cmdstan Users Guide, which was incorrect. In particular, the value `"inf"` does not parse. The default values Python serializes these to _does_, so we should just leave it as it is.

I also removed ujson as part of this fix. ujson implemented differing behavior from Python for infinity which meant round-tripping values was not possible. This was first noted in April of 2013, and did not get fixed until ujson 5.5.0 released last week. So, the options were either pin >=5.5.0, or just remove it. 

#### Copyright and Licensing

Simons Foundation

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -25,7 +25,7 @@ repos:
       - id: mypy
         # Copied from setup.cfg
         exclude: ^test/
-        additional_dependencies: [ numpy >= 1.22, types-ujson ]
+        additional_dependencies: [ numpy >= 1.22]
   # local uses the user-installed pylint, this allows dependency checking
   - repo: local
     hooks:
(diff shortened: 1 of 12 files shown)


ahartikainen:
Do we read or write any large json files in the package?

WardBrian:
We will write large files if the user has large data they're creating in Python

ahartikainen:
Then maybe we should at least have some library that can write those in reasonable time?

There are also other possibilities (not ujson)

WardBrian:
I would be interested in doing some benchmarking, but my guess is that the preprocessing we need to do (such as turning all numpy/pandas objects into lists) is probably nearly as expensive as the serialization. 

Its probably worth looking into options other than `ujson`. I'd like at least:
1. Minimal/no dependencies
2. Matches python on these edge cases
3. Supports custom encoding like the built-in json does. If the user had data which had type `complex` anywhere in it, we were previously calling `ujson`, then, if that failed, calling standard `json` with our custom `complex` encoding.

But I propose that search happens separately

ahartikainen (approved):

Issue #614, opened 2022-09-06
`compile=True` behavior with includes


tillahoffmann:
When the `compile` argument of `CmdStanModel` is `True`, the following code checks whether a newer executable exists such that the compile step can be skipped.

https://github.com/stan-dev/cmdstanpy/blob/e885cb0edf2e611802cb36f491fa05192d4190b3/cmdstanpy/model.py#L481-L488

If the `.stan` file has an `#include` directive, compilation will be skipped if changes were made to the included file but not the `stan_file` argument passed to `CmdStanModel`. Would it be feasible to check included files as well or is that overkill, and I should just use `compile="force"`?

WardBrian:
I think solving this in the most general cases is really tricky. In `stanc_info()` we have a list of included files, but most of them will be relative paths and you need to know the value of `--include-paths` to find them. 

I think this is the kind of thing we'd be happy to accept a PR for, but `compile='force'` will be the recommended thing for now

PR #627, opened 2022-10-12
Recursive up-to-date checks (ref #614).

tillahoffmann:
#### Submission Checklist

- [x] Run unit tests

#### Summary

This PR addresses #614 by running up-to-date checks recursively on included files.

#### Copyright and Licensing

### DIFF
--- a/cmdstanpy/model.py
+++ b/cmdstanpy/model.py
@@ -418,6 +418,35 @@ class CmdStanModel:
             )
         return code
 
+    @staticmethod
+    def _is_out_of_date(stan_file: str, exe_time: float,
+                        include_paths: Iterable[str]) -> bool:
+        """
+        Determine whether the Stan program file or any of its includes are
+        newer than the executable.
+        """
+        if os.path.getmtime(stan_file) > exe_time:
+            return True
+
+        # Recursively identify whether any included files changed.
+        pattern = re.compile(r"^\s*#include\s+(.*?)(?:$|\/\/)", re.MULTILINE)
+        with open(stan_file) as handle:
+            text = handle.read()
+        for child_file in pattern.findall(text):
+            discovered = False
+            for include_path in include_paths:
+                candidate = os.path.join(include_path, child_file)
+                if not os.path.isfile(candidate):
+                    continue
+                discovered = True
+                if CmdStanModel._is_out_of_date(candidate, exe_time,
+                                                include_paths):
+                    return True
+            if not discovered:
+                raise FileNotFoundError(f"{child_file} included by {stan_file} "
+                                        "does not exist")
+        return False
+
     def compile(
         self,
         force: bool = False,
@@ -474,9 +503,13 @@ class CmdStanModel:
                     self._compiler_options.add(compiler_options)
         exe_target = os.path.splitext(self._stan_file)[0] + EXTENSION
         if os.path.exists(exe_target):
-            src_time = os.path.getmtime(self._stan_file)
             exe_time = os.path.getmtime(exe_target)
-            if exe_time > src_time and not force:
+            include_paths = [os.path.dirname(self._stan_file)]
+            include_paths.extend(
+                self._compiler_options._stanc_options.get("include_paths", []))
+            out_of_date = self._is_out_of_date(self._stan_file, exe_time,
+                                               include_paths)
+            if not out_of_date and not force:
                 get_logger().debug('found newer exe file, not recompiling')
                 if self._exe_file is None:  # called from constructor
                     self._exe_file = exe_target
--- a/cmdstanpy/model.py
+++ b/cmdstanpy/model.py
@@ -418,35 +418,6 @@ class CmdStanModel:
             )
         return code
 
-    @staticmethod
-    def _is_out_of_date(stan_file: str, exe_time: float,
-                        include_paths: Iterable[str]) -> bool:
-        """
-        Determine whether the Stan program file or any of its includes are
-        newer than the executable.
-        """
-        if os.path.getmtime(stan_file) > exe_time:
-            return True
-
-        # Recursively identify whether any included files changed.
-        pattern = re.compile(r"^\s*#include\s+(.*?)(?:$|\/\/)", re.MULTILINE)
-        with open(stan_file) as handle:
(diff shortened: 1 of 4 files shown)


cmdstanpy/model.py
```diff
@@ -418,6 +418,35 @@ def code(self) -> Optional[str]:
...
+                        include_paths: Iterable[str]) -> bool:
+        """
+        Determine whether the Stan program file or any of its includes are
+        newer than the executable.
+        """
+        if os.path.getmtime(stan_file) > exe_time:
+            return True
+
+        # Recursively identify whether any included files changed.
+        pattern = re.compile(r"^\s*#include\s+(.*?)(?:$|\/\/)", re.MULTILINE)
```
WardBrian, line 432: Not sure if you're looking for feedback at this stage, but regex is notoriously bad at finding `#include`s and this has lead to a lot of bugs in RStan where they use them.

For example,
```stan
functions {
  /*
  #include to load in common functions
  */
  #include funs.stan
}
```

Is something someone might reasonably write but will be almost impossible to disentangle with regex.

There is another option, though, which is use `self.src_info()`. The dictionary returned by that will have a full list of includes, including recursive includes, precisely as the compiler itself interprets them

tillahoffmann: Much nicer! Updated.

cmdstanpy/model.py
```diff
@@ -474,9 +474,14 @@ def compile(
...
         exe_target = os.path.splitext(self._stan_file)[0] + EXTENSION
         if os.path.exists(exe_target):
-            src_time = os.path.getmtime(self._stan_file)
             exe_time = os.path.getmtime(exe_target)
-            if exe_time > src_time and not force:
+            included_files = [self._stan_file]
+            included_files.extend(self.src_info().get('included_files', []))
+            out_of_date = any(
+                os.path.getmtime(included_file) > exe_time
+                for included_file in included_files
```
WardBrian, line 482: ~Even after using `src_info`, this still needs to take `include_paths` into account, no?~

Oh, actually, I just checked and `src_info` returns the actual included file paths, not just the relative paths. That's much better.

Regardless, I think it is still worth adding a test for an include which uses `include_paths` set to something non-default.

cmdstanpy/model.py
```diff
@@ -474,9 +474,14 @@ def compile(
                     self._compiler_options.add(compiler_options)
         exe_target = os.path.splitext(self._stan_file)[0] + EXTENSION
         if os.path.exists(exe_target):
-            src_time = os.path.getmtime(self._stan_file)
             exe_time = os.path.getmtime(exe_target)
-            if exe_time > src_time and not force:
+            included_files = [self._stan_file]
+            included_files.extend(self.src_info().get('included_files', []))
+            out_of_date = any(
+                os.path.getmtime(included_file) > exe_time
```
WardBrian, line 481: How does `os.path.getmtime` work if a file doesn't exist? It is unlikely, but it's possible that a file gets deleted between line 479 and 481 here, so it would be good to check if it exists if `getmtime` fails badly.

WardBrian (changes requested):
Thanks for putting this together! 

A couple comments and a request for an extra test, but otherwise this looks great!

cmdstanpy/model.py
tillahoffmann: The function raises `FileNotFoundError` if the file doesn't exist. I imagine that's a reasonable error to raise here without wrapping in a custom exception?

WardBrian: Fair enough!

test/test_model.py
```diff
@@ -207,6 +207,52 @@ def test_model_info(self):
         self.assertIn('theta', model_info_include['parameters'])
         self.assertIn('included_files', model_info_include)
 
+    def test_compile_with_includes(self):
+        getmtime = os.path.getmtime
+        configs = [
+            ('add_one_model.stan', ['include-path']),
+            ('bernoulli_include.stan', []),
+        ]
+        for stan_file, include_paths in configs:
```
WardBrian, line 216: Sorry to be nit-picky, can you turn this into a pytest fixture?

e.g.
```python
    @pytest.mark.parametrize("stan_file,include_paths", [('add_one_model.stan', ['include-path']),('bernoulli_include.stan', [])])
    def test_compile_with_includes(self, stan_file, include_paths):
        ...

File: test/test_model.py [hunk 5]
```diff
@@ -207,6 +207,52 @@ def test_model_info(self):
...
+            with LogCapture(level=logging.DEBUG) as log:
+                model.compile()
+            log.check_present(
+                ('cmdstanpy', 'DEBUG', StringComparison('found newer exe file'))
+            )
+
+            # Compile after modifying included file, ensuring cache is not used.
+            def _patched_getmtime(filename: str) -> float:
+                includes = ['divide_real_by_two.stan', 'add_one_function.stan']
+                if any(filename.endswith(include) for include in includes):
```
WardBrian, line 244: Could this be more general but just saying that the filename does _not_ end with the `stan_file` parameter?

WardBrian (approved):
This looks great to me! 

Two comments in the test, but otherwise I think this is ready to merge. Thanks again!

WardBrian:
That’s no worse that the behavior before this PR, right?

tillahoffmann:
True, maybe something to fix another time.

test/test_model.py
tillahoffmann: > Oh, bummer. More of a reason to ditch TestCase one day.

It looks like moving the tests fully to the `pytest` framework wouldn't be too much work. Interest in a PR?

WardBrian: Good question. I’m not sure why we have our tests in the unittest style, and I would be okay with starting to ditch that, but we should probably open an issue to discuss more. 

I have a few gripes with the testing we do, we have a lot of older tests which rely on internal behavior of the classes and make refactors (even backwards compatible ones) very annoying

tillahoffmann: I played around with doing the translation while my brain was stuck on something else. I'll open a draft PR and we can discuss from there?

WardBrian: Sounds good!

PR #632, opened 2022-11-04
Add serialization of samples.

tillahoffmann:
#### Submission Checklist

- [x] Run unit tests

#### Summary

- add a call to `_assemble_draws` to the `__getstate__` magic function such that samples are included when a fit object is serialized with `pickle` or `dill`.
- remove checks before calling `_assemble_draws` because the function already performs this check internally.

#### Copyright and Licensing

### DIFF
--- a/cmdstanpy/stanfit/mcmc.py
+++ b/cmdstanpy/stanfit/mcmc.py
@@ -129,6 +129,10 @@ class CmdStanMCMC:
             # pylint: disable=raise-missing-from
             raise AttributeError(*e.args)
 
+    def __getstate__(self) -> dict:
+        self._assemble_draws()
+        return self.__dict__
+
     @property
     def chains(self) -> int:
         """Number of chains."""
@@ -259,8 +263,7 @@ class CmdStanMCMC:
         CmdStanMCMC.draws_xr
         CmdStanGQ.draws
         """
-        if self._draws.shape == (0,):
-            self._assemble_draws()
+        self._assemble_draws()
 
         if inc_warmup and not self._save_warmup:
             get_logger().warning(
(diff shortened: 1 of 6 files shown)


cmdstanpy/stanfit/mcmc.py
```diff
@@ -129,6 +129,10 @@ def __getattr__(self, attr: str) -> np.ndarray:
             # pylint: disable=raise-missing-from
             raise AttributeError(*e.args)
 
+    def __getstate__(self) -> dict:
```
WardBrian, line 132: Would you mind adding a docstring or comment just explaining what this is used for/why it is here?

WardBrian (commented):
This is very cool. Would you mind adding it to the non-MCMC objects?

tillahoffmann:
So this requires a slightly hacky fix for the tests to work. I have renamed `test_mcmc_serialization` such that it is executed before `test_no_xarray`. The latter uses the `without_import` context manager which in turn uses `reload`. Using `reload` causes an inconsistent state because there are now two types with the same name (but they are two different types). See below for an example.

```python
def test_state(self):
    with self.without_import('xarray', cmdstanpy.stanfit.mcmc):
        pass
    # This will print `False`. Removing the context manager above ensures
    # the line prints `True`.
    print(CmdStanMCMC is cmdstanpy.stanfit.mcmc.CmdStanMCMC)
```

tillahoffmann:
> This is very cool. Would you mind adding it to the non-MCMC objects?

I didn't check carefully, but they seemed to already work out of the box. Will double check though and add some tests.

WardBrian:
We use pytest order (https://pypi.org/project/pytest-order/) to force test ordering for situations like that

PR #639, opened 2022-11-15
[WIP] Change tests to pytest format.

tillahoffmann:
#### Submission Checklist

- [x] Run unit tests

#### Summary

This PR changes the test suite from `unittest.TestCase` format to `pytest`. This should, for example, allow the use of fixtures and `@parametrize` decorators. To minimize the risk of this change breaking anything, it only touches files in the `test` folder. There are *a lot* of changes, but they are mostly trivial.

Some specific changes:

- Used `assert` statements to replace `self.assert*` from the `unittest.TestCase`.
- Used `np.testing.*` wherever the previous code used `self.assertTrue(np.*)`.
- Replaced nested calls in tests (e.g., to run with different `stan_file`s) to use `@parameterize` decorators.
- Removed custom environment variable modifier and instead used `unittest.mock.patch.dict`.
- Replaced `testfixtures.LogCapture` with `pytest`'s `caplog` fixture.

It looks like other tests could benefit from using `@parameterize` decorators, but I've left them unchanged here to avoid changing both framework and logic. Changes and suggestions welcome on this work in progress.

#### Copyright and Licensing


test/__init__.py
```diff
@@ -1,81 +1,66 @@
...
-        finally:
-            env.update(update_after)
-            for k in remove_after:
-                env.pop(k)
 
-    # pylint: disable=invalid-name
-    def assertPathsEqual(self, path1, path2):
-        """Assert paths are equal after normalization"""
-        self.assertTrue(os.path.samefile(path1, path2))
+def check_present(
```
tillahoffmann, line 45: `pytest`'s `caplog` doesn't have a nice `check_present` feature. But I thought that keeping the framework consistent, dropping the `testfixtures` dependency, and adding this function was preferable to mixing frameworks. Bit of a toss up.

test/conftest.py
```diff
@@ -8,12 +8,9 @@
 DATAFILES_PATH = os.path.join(HERE, 'data')
 
 
-# after we have run all tests, use git to delete the built files in data/
-
-
-@pytest.fixture(scope='session', autouse=True)
+@pytest.fixture(autouse=True)
```
WardBrian, line 11: Not making this session level will slow down the tests considerably due to compilation times. Was this change motivated by something being incorrect in the other tests?

test/__init__.py
WardBrian: I agree it's probably better to implement ourselves.

test/__init__.py
```diff
@@ -1,81 +1,66 @@
...
+    caplog: pytest.LogCaptureFixture,
+    *conditions: List[tuple],
+    clear: bool = True,
+) -> tuple:
```
WardBrian, line 49: Lowercase `t` tuple as a type is Python 3.9+ only I believe

test/conftest.py
tillahoffmann: Yes, it looks like the tests have become pretty slow. I changed it to function-level because I had some non-deterministic tests depending on the order in which I ran them. Will try reverting this change and get the tests to pass.

tillahoffmann:
> Did you do this in any sort of automated way?

Mostly a bit of regex matching and replacement with some manual fixes where it failed.

WardBrian:
I noticed tests passed, do you think this is ready for a more thorough review?

tillahoffmann:
Yes, I think this is ready for a more thorough review. Thank you for looking into it.

WardBrian (approved):
Thanks a ton for cleaning all this up!

PR #652, opened 2023-02-13
Variational samples

### LINKED ISSUE #636: Parity for `stan_variable` of `CmdStanVB`.
Calling `stan_variable(name)` on a `CmdStanVB` object currently returns a point estimate whereas calling `stan_variable(name)` on a `CmdStanMCMC` object returns posterior samples. Would there be interest to update `CmdStanVB.stan_variable` to return samples from the variational posterior approximation such that variational and MCMC sampling can be swapped in/out easily? This would be a breaking change.

tillahoffmann:
#### Submission Checklist

- [x] Run unit tests

#### Summary

This PR adds a method to get variational samples akin to `CmdStanMCMC.stan_variable` with shape `(num_draws, *variable_shape)`. It fixes #636 once we figure out a good name for it/are up for making breaking changes to the interface.

It also fixes a small documentation issue in `CmdStanMCMC.stan_variable`.

#### Copyright and Licensing


test/test_cmdstan_args.py
```diff
@@ -103,88 +103,31 @@ def test_good() -> None:
...
-    args = SamplerArgs(adapt_metric_window=-2)
-    with pytest.raises(ValueError):
-        args.validate(chains=2)
-
-    args = SamplerArgs(adapt_step_size=-3)
-    with pytest.raises(ValueError):
-        args.validate(chains=2)
-
-    args = SamplerArgs(adapt_delta=0.88, fixed_param=True)
+@pytest.mark.parametrize("kwargs", [
```
tillahoffmann, line 106: Sorry, just noticed that this change snuck in from #639. I can rewrite the history to get rid of this change if preferable.

tillahoffmann:
@WardBrian, do you have thoughts on this PR, especially naming as discussed in #636?

WardBrian:
Let me give it a more thorough look tomorrow, but the code looks like I’d expect. Naming is the hard thing, especially with the behavior of the . access tied to the “stan_variable” name

WardBrian:
Sorry for the delay @tillahoffmann

The code here looks good. The remaining question is just what we want to do about a name. To me I think making this a breaking change (e.g. naming it `stan_variable`, replacing the existing function) is the thing that makes the most sense.

I'm not sure how @mitzimorris @ahartikainen feel. I think if we were thinking of a major version bump there are probably another small handful of breaking changes which would be nice to do. 

One I just ran into this morning is the fact that functions like `optimize(...)` accept all their arguments positionally, so adding a new argument to the middle of the list is technically a breaking change, which is a bummer. We could make everything after `data` a keyword-only argument which would make this easier to adapt going forward, but that would also be a breaking change to the API (though I imagine almost all existing code uses the keyword calling style for these functions anyway)

tillahoffmann:
> To me I think making this a breaking change (e.g. naming it `stan_variable`, replacing the existing function) is the thing that makes the most sense.

That sounds good; I'll defer to you, @mitzimorris, and @ahartikainen.

ahartikainen:
Yes, let's try to keep everything simple for user, and change stan_variable function.

test/test_variational.py
```diff
@@ -84,7 +84,8 @@ def test_variables() -> None:
     # pylint: disable=C0103
     stan = os.path.join(DATAFILES_PATH, 'variational', 'eta_should_be_big.stan')
     model = CmdStanModel(stan_file=stan)
-    variational = model.variational(algorithm='meanfield', seed=999999)
+    variational = model.variational(algorithm='meanfield', seed=999999,
+                                    output_samples=999)
```
tillahoffmann, line 88: I've added `output_samples=999` for explicitness and in case the default number of variational samples changes in the future.

mitzimorris:
> making this a breaking change (e.g. naming it stan_variable, replacing the existing function) is the thing that makes the most sense.

agreed

tillahoffmann:
Should we close this one in favor of #681?

WardBrian:
I just pushed up the doc changes you made here I missed in that other PR, so at this point I think it makes sense to close this. Thanks again for pointing out this inconsistency originally!

PR #671, opened 2023-06-12
Handle tilde `~` in model paths.

tillahoffmann:
Apple uses tildes in some paths for files that are synced through iCloud (see [here](https://apple.stackexchange.com/q/280338/102603), for example). It's of course a somewhat questionable choice. This PR allows models in these iCloud folders to be compiled and used.

#### Submission Checklist

- [x] Run unit tests

#### Copyright and Licensing

### DIFF
--- a/cmdstanpy/utils/filesystem.py
+++ b/cmdstanpy/utils/filesystem.py
@@ -4,6 +4,7 @@ Utilities for interacting with the filesystem on multiple platforms
 import contextlib
 import os
 import platform
+import re
 import shutil
 import tempfile
 from typing import Any, Iterator, List, Mapping, Tuple, Union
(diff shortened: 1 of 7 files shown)


cmdstanpy/utils/filesystem.py
```diff
@@ -169,11 +170,15 @@ def __exit__(self, exc_type, exc_val, exc_tb) -> None:  # type: ignore
 
 
 class SanitizedOrTmpFilePath:
-    """Context manager for tmpfiles, handles spaces in filepath."""
+    """
+    Context manager for tmpfiles, handles special characters in filepath.
+    """
+    PATTERN = re.compile(r"[\s~]")
```
WardBrian, line 176: This doesn't need to be a member variable, could just be outside the class and avoid the `self.`s later

cmdstanpy/utils/filesystem.py
```diff
@@ -183,12 +188,13 @@ def __init__(self, file_path: str):
...
-        if ' ' in os.path.abspath(file_path):
+        if self.PATTERN.search(os.path.abspath(file_path)):
             tmpdir = tempfile.mkdtemp()
-            if ' ' in tmpdir:
+            if self.PATTERN.search(tmpdir):
                 raise RuntimeError(
-                    'Unable to generate temporary path without spaces! \n'
-                    + 'Please move your stan file to location without spaces.'
+                    'Unable to generate temporary path without special '
+                    'characters! \n Please move your stan file to location '
```
WardBrian, line 196: I think we should say 'special characters or spaces' since the space is by far the most common issue but might not immediately be recognizable as special to users

WardBrian (changes requested):
A couple small comments, but more generally the Windows failures are obviously the bigger issue here. `~` is used on Windows for "short paths": https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file?redirectedfrom=MSDN#short-vs-long-names

cmdstanpy/utils/filesystem.py
tillahoffmann: Makes sense. My motivation for the static variable was to keep everything in one place. Happy to change it though.

ahartikainen:
Maybe pytest skipif for Windows is enough for the tests.

WardBrian (approved):

PR #681, opened 2023-07-25
Stan 2.33: Move IO munging to external package, refactors

WardBrian:
#### Submission Checklist

- [x] Run unit tests

#### Summary

IO changes for 2.33 (tuple support).

- use https://github.com/WardBrian/stanio (tbd: will this move under stan-dev?)
  - support writing tuples to json
  - support reading/extracting tuples from draws array 
- breaking changes:
  - internals of `InferenceMetadata` changed (not technically part of our public API, but it was mentioned some places in our docs) 
  - VB's `.stan_variable` function now returns the varational draws, not the mean approximation (same as #652)
  - MLE and VB always return a np.ndarray from `.stan_variable`, never just a Python float

#### Copyright and Licensing

Simons Foundation

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -15,6 +15,11 @@ on:
         required: false
         default: ''
 
+# only run one copy per PR
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
 jobs:
   get-cmdstan-version:
     # get the latest cmdstan version to use as part of the cache key
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -40,7 +40,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        python-version: ["3.7.1 - 3.7.16", "3.8", "3.9", "3.10", "3.11"]
+        python-version: ["3.8", "3.9", "3.10", "3.11"]
     env:
       GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
     steps:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -32,8 +32,7 @@ jobs:
             if [[ "${{ github.event.inputs.cmdstan-version }}" != "" ]]; then
               echo "version=${{ github.event.inputs.cmdstan-version }}" >> $GITHUB_OUTPUT
             else
-              echo "version=git:develop" >> $GITHUB_OUTPUT
-              # python -c 'import requests;print("version="+requests.get("https://api.github.com/repos/stan-dev/cmdstan/releases/latest").json()["tag_name"][1:])' >> $GITHUB_OUTPUT
+              python -c 'import requests;print("version="+requests.get("https://api.github.com/repos/stan-dev/cmdstan/releases/latest").json()["tag_name"][1:])' >> $GITHUB_OUTPUT
             fi
     outputs:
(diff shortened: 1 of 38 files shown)


cmdstanpy/model.py
mitzimorris, line 1: looks great - this cleanup much needed!

cmdstanpy/compiler_opts.py
```diff
@@ -295,11 +299,11 @@ def compose_stanc(self) -> List[str]:
                     opts.append(f'--{key}')
         return opts
 
-    def compose(self) -> List[str]:
+    def compose(self, filename_in_msg: Optional[str] = None) -> List[str]:
         """Format makefile options as list of strings."""
```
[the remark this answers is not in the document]
WardBrian: Yes, for stanc3. This really only comes into play when we’re using a temporary file (like if there was a space in the path) and we want the original filename to be used in error messages

mitzimorris:
I took a look at the FBProphet code, and it doesn't use `stan_variable` - although it runs optimization, it's pulling things back as a np array.
I'm willing to go with a breaking change.

WardBrian:
I don't think arviz touches anything other than CmdStanMCMC

.github/workflows/main.yml
```diff
@@ -27,7 +32,8 @@ jobs:
             if [[ "${{ github.event.inputs.cmdstan-version }}" != "" ]]; then
               echo "version=${{ github.event.inputs.cmdstan-version }}" >> $GITHUB_OUTPUT
             else
-                python -c 'import requests;print("version="+requests.get("https://api.github.com/repos/stan-dev/cmdstan/releases/latest").json()["tag_name"][1:])' >> $GITHUB_OUTPUT
+              echo "version=git:develop" >> $GITHUB_OUTPUT
+              # python -c 'import requests;print("version="+requests.get("https://api.github.com/repos/stan-dev/cmdstan/releases/latest").json()["tag_name"][1:])' >> $GITHUB_OUTPUT
```
WardBrian, line 36: This needs to be reverted before merging

mitzimorris:
regarding moving stanio under stan-dev - should we create repo stanio with subdirs according to language - 
`python`, `r`, `c++`?

WardBrian:
I don't know enough about the R ecosystem to know if they need this - I think rvars/`posterior` cover a lot of the existing use-cases, but of course tuples change things.

C++ I think it would be very difficult to write something like this, since the return type of the extracting functions depends on the data

mitzimorris:
I think it's OK to make this a repo under standev just for the python io used by CmdStanPy and BridgeStan and future Python packages.

tillahoffmann:
Would it make sense to add a `dtype` argument somewhere in the data loading, e.g., as an extra argument to `CmdStanModel.{sample,optimize,...}` to specify the `dtype` returned by `stan_variable`? This could be useful for reducing the memory footprint of larger models. For example, to evaluate WAIC for the `election88` dataset of CBS News polls, the `log_lik` variable has `num_draws_sampling * chains * 11565` elements. Using cmdstanpy defaults that amounts to ~ 740 MB or ~ 1.5 GB if we include the logits for the binary outcome. Not enormous, but my laptop wasn't too happy when I compared several models.

This may not be the right thread, but thought I'd mention it in case it affects the data munging.

WardBrian:
@tillahoffmann I think that's worth its own issue. It is sort of orthogonal to this issue, since the transformations we apply here are agnostic to what the dtype of the draws are. If we changed what we were reading it from the files, these should propagate nicely

PR #689, opened 2023-08-31
Fix handling of `include-paths` in compiler options.

tillahoffmann:
#### Submission Checklist

- [x] Run unit tests

#### Summary

Including `include-paths` in the `stanc_options` passed to compile currently breaks compilation if `include-paths` is a list of paths rather than a string with comma-separated paths. This PR adds conditional handling based on the type.

Follow-up question: Calling `compile` extends `self._compiler_options` such that the call changes the state of `CmdStanModel` in addition to compiling the model. Is that intended? My gut feeling would've been that passing in `stanc_options` would only affect this one call but not future ones.

#### Copyright and Licensing

### DIFF
--- a/cmdstanpy/compiler_opts.py
+++ b/cmdstanpy/compiler_opts.py
@@ -5,7 +5,7 @@ Makefile options for stanc and C++ compilers
 import os
 from copy import copy
 from pathlib import Path
-from typing import Any, Dict, List, Optional, Union
+from typing import Any, Dict, Iterable, List, Optional, Union
 
 from cmdstanpy.utils import get_logger
 
@@ -99,7 +99,7 @@ class CompilerOptions:
         )
 
     @property
-    def stanc_options(self) -> Dict[str, Union[bool, int, str]]:
+    def stanc_options(self) -> Dict[str, Union[bool, int, str, Iterable[str]]]:
         """Stanc compiler options."""
         return self._stanc_options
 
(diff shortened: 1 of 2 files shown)


WardBrian:
> Follow-up question: Calling `compile` extends `self._compiler_options` such that the call changes the state of `CmdStanModel` in addition to compiling the model. Is that intended? My gut feeling would've been that passing in `stanc_options` would only affect this one call but not future ones.

This is interesting, I can see it going either way (after all, if you pass these arguments to the constructor they do affect more than just the first compilation).

I have recently been wondering what is the use-case for having an _uncompiled_ `CmdStanMCMC` object. It's essentially a broken object at that point, since methods like `sample()` are unavailable. 

I think it's worth considering whether we really want to support the ability to call `compile` on the object after the fact, or if we should move toward an API where the object is constructed either from an executable or a stan file which is immediately compiled. Besides in our tests, I have a hard time seeing use cases for a standalone `compile` method on the object

tillahoffmann:
> I have recently been wondering what is the use-case for having an uncompiled CmdStanMCMC object. It's essentially a broken object at that point, since methods like sample() are unavailable.

That makes sense; I always forget about the `compile = "force"` option in the constructor (which is the only reason I use `compile(force=True)`.

tillahoffmann:
Hm, I can't quite figure out why this is failing in CI on macOS (my local environment is also macOS with python 3.10). Any ideas?

WardBrian:
I think it's because CmdStan 2.33 just released. If you merge in `develop` it should pass I believe

WardBrian (approved):

WardBrian:
Thanks!

PR #729, opened 2024-01-22
Add release notes for 1.2.1

WardBrian:
#### Submission Checklist

- [ ] Run unit tests

#### Copyright and Licensing

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -103,16 +103,14 @@ jobs:
         run: |
           install_cmdstan -h
           install_cxx_toolchain -h
-          python -c "import cmdstanpy; cmdstanpy.install_cmdstan(version='${{ needs.get-cmdstan-version.outputs.version }}', cores=2)"
+          python -c "import cmdstanpy; cmdstanpy.install_cmdstan(version='${{ needs.get-cmdstan-version.outputs.version }}', cores=4)"
 
       - name: Install CmdStan (Windows)
         if: matrix.os == 'windows-latest'
         run: |
           install_cmdstan -h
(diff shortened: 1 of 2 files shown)


cmdstanpy/model.py
```diff
@@ -84,7 +84,8 @@ class CmdStanModel:
     unless the argument ``compile=False`` is specified.
     The set of constructor arguments are:
 
-    :param model_name: Model name, used for output file names.
+    :param model_name: Deprecated: Model name will be taken from executable.
```
bob-carpenter, line 87: Does "Model name will be taken from executable" apply to the current situation or is it what will happen after deprecation?  I'd be inclined to put the actual doc first, then follow with deprecation warning, i.e.,

<tt>
:param model_name: Model name, used for output file names. **Warning**: This argument is deprecated and will be removed in the next release (Version ???). As of Version ???, the model_name will always be taken to be the name of the executable file minus the ``.exe`` suffix if it has one.]
</tt>

bob-carpenter (changes requested):
I was a bit unclear on the language in the first comment.  Otherwise, looks good.

cmdstanpy/model.py
WardBrian: It describes the default behavior now(if the argument is unspecified) and the only behavior in the next version

bob-carpenter (approved):
That's much clearer. Thanks!

WardBrian:
@mitzimorris I think it would be good to release this week if you have a chance to look over the release notes. Thank you!

mitzimorris (commented):
LGreatTM!

PR #734, opened 2024-02-09
Add `diagnose` method to `CmdStanModel`.

tillahoffmann:
#### Submission Checklist

- [x] Run unit tests

#### Summary

This PR adds the `diagnose` method to `CmdStanModel` which can be used to verify gradients, e.g., explicit gradients for custom functions implemented in C++. Documentation is based on the [equivalent `cmdstanr` function](https://mc-stan.org/cmdstanr/reference/model-method-diagnose.html) and other methods, such as `CmdStanModel.variational`.

#### Copyright and Licensing

### DIFF
--- a/cmdstanpy/model.py
+++ b/cmdstanpy/model.py
@@ -2254,6 +2254,11 @@ class CmdStanModel:
                 cmd, capture_output=True, check=False, text=True
             )
             if proc.returncode:
+                get_logger().error(
+                    "'diagnose' command failed!\nstdout:%s\nstderr:%s",
+                    proc.stdout,
+                    proc.stderr,
+                )
                 if require_gradients_ok:
                     raise RuntimeError(
                         "The difference between autodiff and finite difference "
@@ -2268,8 +2273,13 @@ class CmdStanModel:
                 )
 
             # Read the text and get the last chunk separated by a single # char.
-            with open(output) as handle:
-                text = handle.read()
+            try:
+                with open(output) as handle:
+                    text = handle.read()
+            except FileNotFoundError as exc:
+                raise RuntimeError(
+                    "Output of 'diagnose' command does not exist."
+                ) from exc
             *_, table = re.split(r"#\s*\n", text)
             table = (
                 re.sub(r"^#\s*", "", table, flags=re.M)
--- a/cmdstanpy/model.py
+++ b/cmdstanpy/model.py
@@ -2221,6 +2221,8 @@ class CmdStanModel:
             * "model": Gradients evaluated using autodiff.
             * "finite_diff": Gradients evaluated using finite differences.
             * "error": Delta between autodiff and finite difference gradients.
+
+            Gradients are evaluated in the unconstrained space.
         """
 
         with temp_single_json(data) as _data, \
@@ -2237,7 +2239,7 @@ class CmdStanModel:
             if _data is not None:
                 cmd += ["data", f"file={_data}"]
             if _inits is not None:
-                cmd.append(f"inits={_inits}")
+                cmd.append(f"init={_inits}")
 
             output_dir = tempfile.mkdtemp(prefix=self.name, dir=_TMPDIR)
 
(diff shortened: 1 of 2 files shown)


WardBrian:
How would you feel about modeling this more after the `log_prob` method on `CmdStanModel`? Namely, not needing its own Args class or a holder for the return type, but just giving you back the table as a pandas dataframe and that's it.

tillahoffmann:
Good point, updated. Only caveat is that if `required_gradients_ok` is `False`, there is no way to check if the gradients passed the checks. But this is likely to be used only interactively anyway so probably not a problem.

test/test_model.py
```diff
@@ -593,3 +594,30 @@ def test_format_old_version() -> None:
         model.format(max_line_length=88)
 
     model.format(canonicalize=True)
+
+
+def test_diagnose():
```
WardBrian, line 599: Would you mind adding a test of the `inits` argument? Should be easy enough to test against `log_prob`, actually.

cmdstanpy/model.py
```diff
@@ -2203,3 +2204,106 @@ def progress_hook(line: str, idx: int) -> None:
...
+            cmd += ["output", f"file={output}"]
+            if sig_figs is not None:
+                cmd.append(f"sig_figs={sig_figs}")
+
+            get_logger().debug("Cmd: %s", str(cmd))
+
+            proc = subprocess.run(
+                cmd, capture_output=True, check=False, text=True
+            )
+            if proc.returncode:
```
WardBrian, line 2286: The command could have failed for other reasons, like leaving out data for a model that requires it. 

I think to be safe, if there was a return code, to always display the value of `stderr`. 

We should also check that `output` exists before trying to read it, for this reason.

cmdstanpy/model.py
```diff
@@ -2203,3 +2204,106 @@ def progress_hook(line: str, idx: int) -> None:
...
+                        "would like to inspect the output, re-call with "
+                        "`require_gradients_ok=False`."
+                    )
+                get_logger().warning(
+                    "The difference between autodiff and finite difference "
+                    "gradients may exceed the error threshold. Proceeding "
+                    "because `require_gradients_ok` is set to `False`."
+                )
+
+            # Read the text and get the last chunk separated by a single # char.
```
WardBrian, line 2300: [Comment] I'd never looked at the output of `diagnose` before, and wow is it not what I would have expected.

WardBrian (changes requested):

tillahoffmann:
Thanks for the comments, I'll have a look. As an aside, do we mind how close the interfaces of `cmdstanr` and `cmdstanpy` are? E.g., `cmdstanr` uses the object-oriented approach for `diagnose` (see [here](https://mc-stan.org/cmdstanr/reference/CmdStanDiagnose.html)). `cmdstanr` doesn't seem to implement `log_prob` (but I haven't dug too deep into the docs).

WardBrian:
I believe they have support for log_prob through some optional RCPP stuff, which is pretty far outside anything we’d want to do in Python.

We try to be consistent-ish for big things like the names and default values of the arguments to `sample`, but I don’t think it makes sense to aim for complete API matching, especially for smaller things like this

WardBrian (approved):
LGTM!

I'm going to hold off on merging this to see if anything comes in which would justify a 1.2.2 patch, but this will be in the next major version.

bob-carpenter:
@tillahoffmann:  Is this ready to merge?  

@WardBrian:  If it's ready to merge, would you mind taking another final look?

tillahoffmann:
Yes, I think this is ready to be merged.

WardBrian (approved):
Apologies for letting this slip -- my intention was that this would be merged before cmdstanpy 2.0.0, as we've stated that the next non-bugfix release would be 2.0.0 and contain the removal of our stated deprecations.

But, there is a lot of other work that needs to be done (primarily by me) to get to a 2.0.0, that I have been avoiding. 

I think this is fine to go in -- if we need another patch release before 2.0.0, we can either cherry-pick around this or just include it (even if it makes the semantic versioning gods angry).

Thanks again for the feature!

PR #751, opened 2024-05-17
Fix profile file output when running multiple chains in one process

### LINKED ISSUE #750: fit.runset references non-existent *profile*.csv (with STAN_THREADS enabled)
Hi all

I ran a simple 2 chain sampler to test out the usage of profile. According the the cmdstanpy docs it says that the profile csv gets the a similar name as that of the output csv but with "profile-<chain_ID>" appended (or something like this). I have implemented this and the the actual output gets merged into a single .csv file without the <chain_ID> appended when running with `STAN_THREADS=True` (I have not tested this with `STAN_THREADS=False`). When running `fit.runset` however, the profile csv referenced has chain_ID appended, which is a non-existent file. Just wanted to bring this to your attention.

Also, I wanted to know if there maybe is a method to access the output from the profile.csv file? Currently I am using just using runset to locate the folder in which the results are stored and then reading to a dataframe (will probably change `output_dir` to something other then the default when running my actual inference).

## Versions
cmdstan: 2.34.0
cmdstanpy: 1.2.1

WardBrian:
#### Submission Checklist

- [x] Run unit tests

#### Summary

Closes #750 - profile_files is now like stdout_files where there is only one of them when there is only one process being run.

Note: I am not convinced cmdstanpy's API here makes any sense. The profile file is _not_ like the diagnostic file where it is only saved if requested, but rather it is saved _if the model in question uses the profiling feature at all_. The default value is `'profile.csv'`, not `''`

#### Copyright and Licensing

### DIFF
--- a/cmdstanpy/install_cmdstan.py
+++ b/cmdstanpy/install_cmdstan.py
@@ -523,6 +523,8 @@ def retrieve_version(version: str, progress: bool = True) -> None:
         first = tar.next()
         if first is not None:
             top_dir = first.name
+        else:
+            top_dir = ''
         cmdstan_dir = f'cmdstan-{version}'
         if top_dir != cmdstan_dir:
             raise CmdStanInstallError(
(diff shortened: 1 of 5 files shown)


mitzimorris (approved):
LGTM - very clean - many thanks!

PR #779, opened 2025-02-17
add timing information for sampler outputs

mitzimorris:
#### Submission Checklist

- [x] Run unit tests

#### Summary

Added routine to `utils/stancsv.py' to scrape timing information from sampler outputs; it captures the timing information from each run as a dict with keys 'warmup', 'sampling', and 'total'.

Capturing total time will allow us to calculate the ESS/sec.   One could argue that this should be part of CmdStan's 'stansummary' - it was dropped to match current outputs of `posterior` and `cmdstanr`.  Note that `cmdstanr` provides method 'time' to do this:  https://mc-stan.org/cmdstanr/reference/fit-method-time.html

#### Copyright and Licensing


cmdstanpy/stanfit/mcmc.py
```diff
@@ -97,8 +97,9 @@ def __init__(
         self._max_treedepths: np.ndarray = np.zeros(
             self.runset.chains, dtype=int
         )
+        self._chain_timing: List[Optional[Dict[str, float]]] = []
```
WardBrian: The `Optional` here means some elements of this list can be `None`, but if I'm reading the parsing code correctly that case is impossible. Either we should let the parser fail without raising an error or change the type here and on the method

cmdstanpy/stanfit/mcmc.py
```diff
@@ -240,6 +241,14 @@ def max_treedepths(self) -> Optional[np.ndarray]:
         """
         return self._max_treedepths if not self._is_fixed_param else None
 
+    @property
+    def timing(self) -> List[Optional[Dict[str, float]]]:
```
WardBrian: Should this be `time()` to match cmdstanr's naming? (if we want to go our own way, I prefer the plural `timings`, but that might not make as much sense if added for the other algos)

cmdstanpy/utils/stancsv.py
```diff
@@ -380,6 +381,62 @@ def scan_sampling_iters(
...
+        else:
+            if 'sampling' in tokens[2]:
+                key = 'sampling'
+            elif 'total' in tokens[2]:
+                key = 'total'
+            try:
+                t = float(tokens[0])
+            except ValueError:
+                raise ValueError(f"Invalid timing value at line {lineno}: {content}")
+        timing[key] = t
```
WardBrian: I think this could be a bit clearer, something like:

```python
        if 'Warm-up' in content:
            key = 'warmup'
            time_str = tokens[2]
        elif 'Sampling' in content:
            key = 'sampling'
            time_str = tokens[0]
        elif  'Total' in content:
            key = 'total'
            time_str = tokens[0]
            
       
        try:
            timing[key] = float(time_str)
        except ValueError as e:
            raise ValueError(f"Invalid timing value at line {lineno}: {content}") from e
```

This is independent of the question of if we want to actually raise the value error here or
instead make this Optional like the type in `mcmc.py` suggests

WardBrian (changes requested):
Three minor points but this looks good, thanks for tackling!

cmdstanpy/stanfit/mcmc.py
mitzimorris: good call - removed Optional

cmdstanpy/utils/stancsv.py
mitzimorris: we should be strict - otherwise, consumers of 'time' will have to deal with missing info.

implemented suggestion, added a few more checks.

cmdstanpy/stanfit/mcmc.py
mitzimorris: changed to match cmdstanr.

cmdstanpy/utils/stancsv.py
```diff
@@ -381,6 +382,67 @@ def scan_sampling_iters(
...
+            time_str = tokens[0]
+        elif 'Total' in content:
+            key = 'total'
+            time_str = tokens[0]
+        else:
+            raise ValueError(f"Invalid time at line {lineno}: {content}")
+        try:
+            t = float(time_str)
+        except ValueError:
+            # pylint: disable=raise-missing-from
```
WardBrian: I think we actually should raise-from here

cmdstanpy/stanfit/mcmc.py
```diff
@@ -301,6 +311,8 @@ def _validate_csv_files(self) -> Dict[str, Any]:
                     save_warmup=self._save_warmup,
                     thin=self._thin,
                 )
+                time_info = cast(Dict[str, float], dzero.get("time") or {})
```
WardBrian: Is this just to make mypy happy? I generally prefer `# type: ignore` comments over cast statements, as casts actually do have a runtime cost

WardBrian (approved):
Looks great -- thanks!

PR #788, opened 2025-05-04
Fix different behavior for jupyter vs. terminal

qiangxinglin:
#### Submission Checklist

- [x] Run unit tests

#### Summary

When run code in non-English system, jupyter would implicitly set encoding to 'utf-8', explicitly set the encoding here would make the behavior consistent.  
Fix downstream bug https://github.com/facebook/prophet/issues/2462

#### Copyright and Licensing

myself

### DIFF
--- a/cmdstanpy/utils/command.py
+++ b/cmdstanpy/utils/command.py
@@ -48,7 +48,7 @@ def do_command(
                 stderr=subprocess.STDOUT,  # avoid buffer overflow
                 env=os.environ,
                 universal_newlines=True,
-                encoding=locale.getencoding(),
+                encoding=locale.getdefaultlocale()[1],
             )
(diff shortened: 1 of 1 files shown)


cmdstanpy/utils/command.py
```diff
@@ -47,6 +48,7 @@ def do_command(
                 stderr=subprocess.STDOUT,  # avoid buffer overflow
                 env=os.environ,
                 universal_newlines=True,
+                encoding=locale.getdefaultlocale()[1],
```
WardBrian, line 51: It's too bad `encoding='locale'` and `locale.getencoding()` are only available on newer versions. I think `locale.getpreferredencoding(False)` is the best option for older versions

Issue #745, opened 2024-04-22
add "create_inits" to fits from sampler, optimization, advi


mitzimorris:
#### Summary:

Add method `create_inits` to  classes CmdStanMCMC, CmdStanMLE, CmdStanLaplace

#### Description:

Facilitate better initialization using outputs of all inference methods.

amas0:
@mitzimorris Is this still a desired feature? I'd be happy to work on it if so.

mitzimorris:
yes!  that would be great.

amas0:
This can now be closed @mitzimorris

mitzimorris:
many thanks @amas0 !

https://github.com/stan-dev/cmdstanpy/pull/791

PR #791, opened 2025-05-08
Add new `create_inits()` methods to other stanfit classes

amas0:
#### Submission Checklist

- [ x ] Run unit tests
- [ x ] Declare copyright holder and open-source license: see below

#### Summary

This PR addresses #745. Implementation largely mirrors that for `CmdStanPathfinder`. In particular, this adds the following methods: `CmdStanVB.create_inits()`, `CmdStanLaplace.create_inits()`, `CmdStanMLE.create_intis()`, `CmdStanMCMC.create_inits()`. With the exception of MLE, inits are selected by sampling without replacement from draws. The MLE implementation will initialize all requested chains at the optimized parameter values.

Unit tests for all these methods are added as well.

I also updated the User's Guide example on creating inits from VI outputs to be more general and cover these new methods.

I assume this is already well-known but it seems like there is a cleaner library structure to have each of these different stanfit objects be implementations of the same parent (or abstract) class to unify these implementations (and those of other common methods). I haven't worked through the details, but it seemed possible given how similar the internal structures tend to be.

#### Copyright and Licensing

### DIFF
--- a/cmdstanpy/stanfit/laplace.py
+++ b/cmdstanpy/stanfit/laplace.py
@@ -1,5 +1,5 @@
 """
-    Container for the result of running a laplace approximation.
+Container for the result of running a laplace approximation.
 """
 
 from typing import (
@@ -51,6 +51,39 @@ class CmdStanLaplace:
 
         config = scan_generic_csv(runset.csv_files[0])
         self._metadata = InferenceMetadata(config)
+
+    def create_inits(
+        self, seed: Optional[int] = None, chains: int = 4
+    ) -> Union[List[Dict[str, np.ndarray]], Dict[str, np.ndarray]]:
+        """
+        Create initial values for the parameters of the model
+        by randomly selecting draws from the Laplace approximation.
+
+        :param seed: Used for random selection, defaults to None
+        :param chains: Number of initial values to return, defaults to 4
+        :return: The initial values for the parameters of the model.
+
+        If ``chains`` is 1, a dictionary is returned, otherwise a list
+        of dictionaries is returned, in the format expected for the
+        ``inits`` argument. of :meth:`CmdStanModel.sample`.
+        """
+        self._assemble_draws()
+        rng = np.random.default_rng(seed)
+        idxs = rng.choice(self._draws.shape[0], size=chains, replace=False)
+        if chains == 1:
+            draw = self._draws[idxs[0]]
+            return {
+                name: var.extract_reshape(draw)
+                for name, var in self._metadata.stan_vars.items()
+            }
+        else:
+            return [
+                {
+                    name: var.extract_reshape(self._draws[idx])
+                    for name, var in self._metadata.stan_vars.items()
+                }
+                for idx in idxs
+            ]
 
     def _assemble_draws(self) -> None:
         if self._draws.shape != (0,):
--- a/cmdstanpy/stanfit/laplace.py
+++ b/cmdstanpy/stanfit/laplace.py
@@ -65,7 +65,7 @@ class CmdStanLaplace:
 
         If ``chains`` is 1, a dictionary is returned, otherwise a list
         of dictionaries is returned, in the format expected for the
-        ``inits`` argument. of :meth:`CmdStanModel.sample`.
+        ``inits`` argument of :meth:`CmdStanModel.sample`.
         """
         self._assemble_draws()
         rng = np.random.default_rng(seed)
--- a/cmdstanpy/stanfit/mle.py
+++ b/cmdstanpy/stanfit/mle.py
@@ -1,7 +1,7 @@
 """Container for the result of running optimization"""
 
 from collections import OrderedDict
-from typing import Dict, List, Optional, Tuple, Union
+from typing import Dict, Optional, Tuple, Union
 
 import numpy as np
 import pandas as pd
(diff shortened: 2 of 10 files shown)


cmdstanpy/stanfit/laplace.py
```diff
@@ -52,6 +52,39 @@ def __init__(self, runset: RunSet, mode: CmdStanMLE) -> None:
...
+        Create initial values for the parameters of the model
+        by randomly selecting draws from the Laplace approximation.
+
+        :param seed: Used for random selection, defaults to None
+        :param chains: Number of initial values to return, defaults to 4
+        :return: The initial values for the parameters of the model.
+
+        If ``chains`` is 1, a dictionary is returned, otherwise a list
+        of dictionaries is returned, in the format expected for the
+        ``inits`` argument. of :meth:`CmdStanModel.sample`.
```
WardBrian: I think this typo is also in Pathfinder's doc, but the period after "argument" should be removed

cmdstanpy/stanfit/mle.py
```diff
@@ -36,6 +36,30 @@ def __init__(self, runset: RunSet) -> None:
...
+        """
+        mle_inits = {
+            name: var.extract_reshape(self.optimized_params_np)
+            for name, var in self._metadata.stan_vars.items()
+        }
+
+        if chains == 1:
+            return mle_inits
+        else:
+            return [mle_inits for _ in range(chains)]
```
WardBrian: Because the default behavior is if you pass a dictionary to sample, it uses it for every chain, I don't know if this is super necessary. 

I do think this function should probably accept both a seed and chains argument, even if they're both unused, to let people write more generic code

WardBrian: (I also think this function might be exactly equivalent to `self.stan_variables()` after those changes?

docsrc/users-guide/examples/VI as Sampler Inits.ipynb
WardBrian, line 1: I think we don't generally check in the output of the notebooks

WardBrian (commented):
Thanks! Overall this looks great, I just have a few minor comments.

Re:
> I assume this is already well-known but it seems like there is a cleaner library structure to have each of these different stanfit objects be implementations of the same parent (or abstract) class to unify these implementations (and those of other common methods).

Yes, we definitely could. So far most of what we've done in this style is using helper methods we throw in the utils subpackage for code re-use, but inheritance would also get us there, and maybe a bit more cleanly. In another project that deals with Stan outputs, I don't even differentiate between the different algorithms, I just have one 'StanOutput' class that can hold any of them.

cmdstanpy/stanfit/laplace.py
amas0: The perils of copy/paste -- good catch.

cmdstanpy/stanfit/mle.py
amas0: Updated the signature for compatibility. I also use `stan_variables()` in the function now, but because `stan_variables()` will currently return a `float` if the parameter is scalar, I just map everything to an `np.array` for type compatibility. It seems this is intended to be the future behavior of `stan_variables()` in 2.0, at which point it will be equivalent.

docsrc/users-guide/examples/VI as Sampler Inits.ipynb
amas0: Good catch, fixed. It might be worth adding a pre-commit hook for this using `nbstripout`?

WardBrian: I think there is one that takes too long to run so we *do* check it in, otherwise we probably could do with a pre-commit

amas0:
Thanks for the review -- responded to your comments and made changes where appropriate.

Not sure what's going on with that failed pytest check. Some of the cancelled jobs seem to have gotten through the testing stage without issue.

Going through the library, there are a number of references to a 2.0 version -- that might be a good opportunity to clean up some of these stanfit objects and such. Not sure if there are explicit plans for what that will contain (or timeline), just a thought. Happy to help out if there's a direction there.

WardBrian:
Because ADVI can be unstable it might be worth setting a known-good seed in the new test that runs it

amas0:
Okay, pinned some seeds that seemed fine and checks look good. Serendipitous that I went back and did that, because I found a bug in one of the tests I wrote, which I now have fixed.

I think this should be set?

bob-carpenter:
Thanks.  @mitzimorris may be able to review and approve sooner than @WardBrian for this one.

mitzimorris:
the code looks great.

the documentation section on creating inits should also mention creating inits from previous MCMC runs.
I would be happy merging this now and doing a follow-on PR to improve the docs.

docsrc/users-guide/examples/VI as Sampler Inits.ipynb
```diff
@@ -4,13 +4,16 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "## Using Variational Estimates to Initialize the NUTS-HMC Sampler\n",
+    "## Using Estimates from Variational, Laplace, or Optimization Methods to Initialize the NUTS-HMC Sampler\n",
```
mitzimorris: we should rename this page altogether - "Initializing the NUTS-HMC sampler"

docsrc/users-guide/examples/VI as Sampler Inits.ipynb
```diff
@@ -4,13 +4,16 @@
...
     "\n",
-    "In this example we show how to use the parameter estimates return by Stan's variational inference algorithms\n",
-    "[pathfinder ](https://mc-stan.org/docs/cmdstan-guide/pathfinder-config.html)  and \n",
-    "[ADVI ](https://mc-stan.org/docs/cmdstan-guide/variational_config.html) \n",
-    "as the initial parameter values for Stan's NUTS-HMC sampler.\n",
-    "By default, the sampler algorithm randomly initializes all model parameters in the range uniform\\[-2, 2\\].  When the true parameter value is outside of this range, starting from the estimates from Pathfinder and ADVI will speed up and improve adaptation.\n",
+    "In this example, we show how to use parameter estimates returned by Stan's various posterior approximation or optimization algorithms as initial values for Stan's NUTS-HMC sampler. These include:\n",
```
mitzimorris: In this example we show how to use the returned estimates from a previous run of any of the inference algorithms can be used as initial values for  Stan's NUTS-HMC sampler.

mitzimorris (approved):
This could be done as a follow-on PR - the code looks good.

amas0:
I'm happy to quickly update the docs as part of this PR. I do mention that previous MCMC estimates can be used as inits, but it's more of an aside at the end. I'll make it clearer that it's included.

docsrc/users-guide/examples/VI as Sampler Inits.ipynb
amas0: Updated.

docsrc/users-guide/examples/VI as Sampler Inits.ipynb
amas0: Updated.

amas0:
Updated the docs to clarify that one can use the output of one MCMC run as inits into another.

One relatively minor point. With this update, the file name of the example notebook `VI as Sampler Inits.ipynb` is somewhat outdated. I think it would make sense to update this, but doing so would break any existing links to that example as the URL is generated from the file name. I imagine a redirect could be set up with Sphinx somehow.

mitzimorris:
you're right - we're stuck with the filename.

mitzimorris (approved):
Thanks - this all looks great!

Issue #795, opened 2025-06-12
Feature idea: module-level helper functions for controlling/disabling logging


amas0:
Currently when using `cmdstanpy`, a decent bit of logging is produced by default that is handled by the built-in `logging` library. This isn't always desired and the documentation details how it can be disabled:

```python
import logging
import cmdstanpy

cmdstanpy_logger = logging.getLogger("cmdstanpy")
cmdstanpy_logger.disabled = True
```
or to set debug logs:

```python
import logging
import cmdstanpy

cmdstanpy_logger = logging.getLogger("cmdstanpy")
cmdstanpy_logger.setLevel(logging.DEBUG)
```
while functional, it's somewhat clunky and requires users add a separate import and remember the details of how to work with the Python `logging` library. What I am envisioning is some helper functions at the module level that can act as shortcuts to these behaviors. So, something like:

```python
import cmdstanpy

cmdstanpy.disable_logging()
```

would be shorthand for the first example. A corresponding `cmdstanpy.enable_logging()` would go along with it. Log levels could be similar:

```python
import cmdstanpy

cmdstanpy.set_logging_level("DEBUG")
```

The lower-level interface of using the `logging` library directly could still be used, but this would simplify the process for end users.

Curious to hear what the maintainers think (this has been a minor annoyance as a user of the library), and if interesting, I'd be happy to contribute a PR.

bob-carpenter:
Thanks for writing a very clear issue.   I wrote my own `disable_logging()` function for the same reason.

Given that setting logging levels has global side effects, I'm not sure if this is something that should go into CmdStanPy's API.  

@mitzimorris and @WardBrian should be able to answer more definitively.

amas0:
Are there instances where `cmdstanpy`'s named logger conflicts with loggers in other libraries? My understanding of how named loggers work in Python is that the logging level or enabling/disabling can be done independent from other logging behavior. I'd be curious what global side effects would result.

WardBrian:
> Given that setting logging levels has global side effects, I'm not sure if this is something that should go into CmdStanPy's API.

CmdStanPy already [fiddles with your global state](https://github.com/stan-dev/cmdstanpy/blob/develop/cmdstanpy/utils/logging.py#L12-L24) to enable logging, so this seems fairly benign to add to me.

That being said, I've never loved that we enable logging by default, so I'd also consider an alternative to this proposal where logging is off by default and we provide a helper to turn it _on_ in the 2.0 API

[...]

PR #796, opened 2025-06-18
Enable/disable logging helpers

amas0:
#### Submission Checklist

- [x] Run unit tests

#### Summary

This is an example implementation to meet #795. It also includes a bug fix pointed out by @skelliest. The primary contribution is the inclusion of the `enable_logging/disable_logging` "functions" at the module level.

Here's an example demonstration of usage:

```python
import cmdstanpy

cmdstanpy.disable_logging()

model = cmdstanpy.CmdStanModel(...)
model.sample(chains=1, show_progress=False)

cmdstanpy.enable_logging()
```

I also tried to be clever and make this same interface work as a context manager, so you can do:

```python
import cmdstanpy

with cmdstanpy.disable_logging():
    model = cmdstanpy.CmdStanModel(...)
    model.sample(chains=1, show_progress=False)
```

I like the idea of this functionality, but the implementation may be stretching the bounds of what is reasonable. For one, the `disable_logging`/`enable_logging` are actually classes so the function call in `cmdstanpy.disable_logging()` actually returns an instance of the `cmdstanpy.utils.logging.disable_logging` class, but this is the only way I could conceive of getting the same interface to function as both a "function" and as a context manager.

The most noticeable result of this is that if a user were to call `cmdstanpy.disable_logging()` in a standalone cell in a Jupyter notebook; they'd see the default `repr` in the cell output. You can get around this by putting a semicolon at the end of the line, which is what I did in the `outputs.rst` docs, but it's more of a band-aid. This bothers me, but I don't see a way around it and also maintain the dual-interface.

Some feedback/thoughts I'd like:

1. Is the juice not worth the squeeze for this dual "function"/context manager?
2. If the dual interface is fine, should the `__repr__`s for these objects be updated so if called in an empty Juypter cell, it's a little cleaner?
3. If not, what would the preferred solution be? Some variations off the top of my head:

* Remove the context manager and implement as a simple function. Probably the most straightforward approach, but will require users to manually re-enable logging after disabling.
* Create separate interfaces for the function and the context manager, something like:
```python
import cmdstanpy

with cmdstanpy.context(logging=False):
  model.sample()
```  
in addition to `cmdstanpy.disable_logging()`

* Remove the _function_ and only have a context manager option. This would encourage only local disabling of logging and the like, while of course leaving open the option to disable logging globally the old fashioned way. 

Probably overthinking what is a relatively minor change, happy to hear some thoughts.

#### Copyright and Licensing

### DIFF
--- a/cmdstanpy/utils/logging.py
+++ b/cmdstanpy/utils/logging.py
@@ -3,6 +3,7 @@ CmdStanPy logging
 """
 import functools
 import logging
+from contextlib import AbstractContextManager
 
 
 @functools.lru_cache(maxsize=None)
@@ -23,3 +24,33 @@ def get_logger() -> logging.Logger:
         )
         logger.addHandler(handler)
     return logger
+
+
+class enable_logging(AbstractContextManager):
+    """Enable cmdstanpy logging. Can be used as a context manager"""
+
+    def __init__(self) -> None:
+        self.logger = get_logger()
+        self.prev_state = self.logger.disabled
+        self.logger.disabled = False
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_value, traceback) -> None:
+        self.logger.disabled = self.prev_state
+
+
+class disable_logging(AbstractContextManager):
+    """Disable cmdstanpy logging. Can be used as a context manager"""
+
+    def __init__(self) -> None:
+        self.logger = get_logger()
+        self.prev_state = self.logger.disabled
+        self.logger.disabled = True
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_value, traceback) -> None:
+        self.logger.disabled = self.prev_state
--- a/cmdstanpy/utils/logging.py
+++ b/cmdstanpy/utils/logging.py
@@ -1,15 +1,18 @@
 """
 CmdStanPy logging
 """
+
 import functools
 import logging
+import types
 from contextlib import AbstractContextManager
+from typing import Optional, Type
 
 
 @functools.lru_cache(maxsize=None)
 def get_logger() -> logging.Logger:
     """cmdstanpy logger"""
-    logger = logging.getLogger('cmdstanpy')
+    logger = logging.getLogger("cmdstanpy")
     if not logger.hasHandlers():
         # send all messages to handlers
         logger.setLevel(logging.DEBUG)
@@ -18,7 +21,7 @@ def get_logger() -> logging.Logger:
         handler.setLevel(logging.INFO)
         handler.setFormatter(
             logging.Formatter(
-                '%(asctime)s - %(name)s - %(levelname)s - %(message)s',
+                "%(asctime)s - %(name)s - %(levelname)s - %(message)s",
                 "%H:%M:%S",
             )
         )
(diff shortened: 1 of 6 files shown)


WardBrian:
I think that the function-doubling-as-context-manager is fine and matches up well with some other patterns like how the `open` "function" works. pytorch's `no_grad` context manager also doubles as a decorator, which might be a step too far...

My other thoughts:
- It might be worth making [re-entrant](https://docs.python.org/3/library/contextlib.html#reentrant-cms) by having __entry__ set the `logger.disabled` flag again? I doubt this would come up in real use cases, but it doesn't hurt
- I would prefer a repr that returns `''`
- These two classes are almost identical, so I think it might be worth having one class that takes in a flag true/false, and then two functions that only create-and-return that class with a different argument.

bob-carpenter:
Thanks, @amas0.  I'm looking forward to the feature!

amas0:
Okay, I believe I have addressed the points you raised @WardBrian. I added in some unit tests for good measure.

The re-entrant bit seems to work, as I've confirmed this monstrosity works fine:

```python
import cmdstanpy

dis = cmdstanpy.disable_logging()

with cmdstanpy.enable_logging() as en:
    with cmdstanpy.disable_logging():
        with en:
            with dis:
                model = cmdstanpy.CmdStanModel(stan_file="./bernoulli.stan")
                model.sample(show_progress=False, show_console=False, chains=1)
```

For a future 2.0, I think turning logging off by default with a context manager helper to turn it on would be a nice interface. Fiddling about with global logging state stuff is never particularly nice.

docsrc/users-guide/outputs.rst
```diff
@@ -60,24 +60,47 @@ You may notice CmdStanPy can produce a lot of output when it is running:
...
     fit = model.sample(data=data_file, show_progress=False)
 
-Or one can remove the logging handler that CmdStanPy installs by default and install their own for more
-fine-grained control. For example, the following code sends all logs (including the ``DEBUG`` logs, which are hidden by default),
-to a file.
+We can re-enable this by calling `enable_logging()`:
+
+.. ipython:: python
+
+    cmdstanpy.enable_logging();
```
WardBrian:
```suggestion
    cmdstanpy.enable_logging()
```

docsrc/users-guide/outputs.rst
```diff
@@ -60,24 +60,47 @@ You may notice CmdStanPy can produce a lot of output when it is running:
...
+
+For example, it can be disabled entirely:
 
 .. ipython:: python
 
-    import logging
-    cmdstanpy_logger = logging.getLogger("cmdstanpy")
-    cmdstanpy_logger.disabled = True
+    import cmdstanpy
+    cmdstanpy.disable_logging();
```
WardBrian:
```suggestion
    cmdstanpy.disable_logging()
```

docsrc/users-guide/outputs.rst
WardBrian, line 1: Would you mind also adding new sections and `.. autofunction::` declarations to `docsrc/api.rst`?

cmdstanpy/utils/logging.py
```diff
@@ -1,25 +1,62 @@
...
+
+    def __enter__(self) -> "ToggleLogging":
+        self.logger.disabled = self.disable
+        return self
+
+    def __exit__(
+        self,
+        exc_type: Optional[Type[BaseException]],
+        exc_value: Optional[BaseException],
+        traceback: Optional[types.TracebackType],
```
WardBrian, line 50: [optional] I realize these are just here to make typing happy, since we don't need them can we get away with `*args, **kwargs`? That way we don't need to import quite as much just for typing purposes

WardBrian (commented):
Thanks @amas0!

I have a few comments, mostly on the documentation, but otherwise this all looks good to me

amas0:
New commits should address all those points

WardBrian (approved):
Thanks @amas0!

Issue #785, opened 2025-04-11
Path forward on IO operations


WardBrian:
This issue serves to document plans that myself, @bob-carpenter, and @mitzimorris have all discussed in person.

Issues:
- the existing IO code re-uses very little between the different services. 
- the existing IO code is slow.
- the existing IO code depends on things like the order of comments in the files.

Steps:
- [x] Split IO code into two passes. One pass should ignore all comments and use a fast reader like numpy's fromtxt to read in the draws. This can be used for all services. The second pass should deal with the comments. This will be service-specific. This change can be done any time
- [ ] Replace the second pass of the above with the newer JSON outputs rather than CSV comments. This requires a minimum supported cmdstan version: #685
- [ ] The first step could later be replaced with some sort of binary output. This would require upstream Stan changes

eroesch:
Thanks for creating the issue!

When I spoke to Mitzi, the other day, I had something like that in mind, but I was not thinking of using numpy, rather navigate through the file with seek() [1]. I cannot imagine this would be slower than numpy.fromtxt().

```python
with open("bernoulli-20250416172900_1.csv", "r") as f:

    # Move to byte 1678 in my test file
    # which is the start of the chain data
    f.seek(1678)  # probably NOT fixed across files though
    beg_pos = f.tell()

    # Read in the length of the file
    f.seek(0, 2)
    end_pos = f.tell()

    # Move back to exclude the tail information about timing
    offset = 129  # probably fixed across files
    remaining = end_pos - 1678 - offset

    f.seek(1678)
    data = f.read(remaining)
    print(data)
```

This is very fast. The only annoying thing about that solution is that the beg_pos where the chain data starts is possibly not the same, depending on the number of characters in the preamble that's commented. This is not the case of the tail of the csv, because the formatting is, I believe, forcing a dedicated structure likely to always be the same size in bytes.

Arguably, we could undershoot roughly where we think we're close to the data (like in my example), then either filter out the few remaining lines to the data. That's very likely to be faster than the current solution, which is to read each line, check if they start with "#" and filter out, line after line.

That would work, I think, but it would be cleaner to separate the preamble from the data directly in Stan, adding a new flag like save_cmdstan_config to save only the data of the chains.

The other solution I was thinking about was using grep, which is quite optimised already. :) .. but I haven't tested it.

We probably need a way to evaluate the time that these steps in cmdstanpy actually take. What do you use?

Have a great Easter break!

References:
[...]

Issue #685, opened 2023-08-24
Remove support for CmdStan pre-2.34 in a future version


WardBrian:
#### Summary:

It would be nice to remove a lot of legacy code in the package which has been made obsolete by recent CmdStan updates. For example, supporting running multiple chains _without_ using the `num_chains` argument to CmdStan

#### Current Version:
1.1

WardBrian:
I updated my wish in the title to be only supporting 2.34 and newer so we can assume the existence of the `save_cmdstan_config` (https://github.com/stan-dev/cmdstanpy/issues/714) and `save_metric` (https://github.com/stan-dev/cmdstanpy/issues/713) arguments

After another year, it would make sense to make this 2.37, which will allow passing comma-separated lists of filenames and make a lot of our logic in runset completely obsolete

amas0:
Regarding dropping support for older versions: for what it's worth, I would lean towards being a bit more aggressive with only maintaining support for recent versions (or perhaps tying some of `cmdstanpy`'s versioning to `cmdstan` versioning?) I think it makes sense for a wrapper interface to proceed mostly in step with the tool it's wrapping. 

The added degree of freedom of trying to map out the behavior of this code along with how things have changed over the last few years in cmdstan adds a bit of complexity if you don't already have a good handle on it. 

From a user perspective; warning a user to install a new version of cmdstan or downgrade cmdstanpy if there a mismatch seems not so bad.

PR #799, opened 2025-07-22
Implementation of faster MCMC CSV parsing and Stan CSV utilities

amas0:
#### Submission Checklist

- [x] Run unit tests

#### Summary

This is in an initial PR that addresses parts of #785. This introduces some general utilities for parsing Stan CSV files and uses some of those changes within `CmdStanMCMC`. 

Some key ideas:

* In an attempt to generalize some of the CSV parsing across different inference methods, this change uses the fact that the CSV outputs are made up of an sequences of commented/uncommented lines that correspond to different sections. The parser logic used here is state-based where a new parsing rule is used each time a transition occurs between a commented and un-commented section.
For example in the standard MCMC output, the sections are `config`, `warmup`, `adaptation`, `samples`, and `timing`. Then one only needs to write some parsing code for each section.
* An optimized parser based on `polars` is implemented that takes in `list[bytes]` corresponding to the lines in the CSV outputs that correspond to draws and produces a `np.array` as a drop in replacement for existing parsing.
* With these utilities, I introduce a `StanCsvMCMC` dataclass which represents the parsed output of a Stan CSV file. It includes a dict representing the sampler config (which just uses the existing `scan_config` function), the warmup draws as a numpy array (if present), the step size and mass matrix (if present), the sampling draws as a numpy array, and the timing information as a dictionary. The idea being that it will be easier to write and read code with this structured representation than have to know how the CSV files are structured.

I use these changes to update the `CmdStanMCMC._assemble_draws` method, which, with the optimized parsing, runs noticeably faster for large models.

Putting this PR as a draft for now, as I want to get some feedback and am planning on writing unit tests to cover the new code, but all existing unit tests are passing on this.

My plan for next steps is to implement a corresponding `StanCsv*` for the various inference methods (I've already done most of this because the logic is very similar) and then update how each of the Stan fit objects pull info from the CSV files. This will make things more consistent across the inference methods and hopefully clean up some parts of the library.

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting (this will be you or your assignee, such as a university or company): Myself

### DIFF
--- a/cmdstanpy/stanfit/mcmc.py
+++ b/cmdstanpy/stanfit/mcmc.py
@@ -441,30 +441,30 @@ class CmdStanMCMC:
         )
         self._step_size = np.empty(self.chains, dtype=float)
 
+        mass_matrix_per_chain = []
         for chain in range(self.chains):
-            parsed_csv = stancsv.StanCsvMCMC.from_csv(
-                self.runset.csv_files[chain],
-                is_fixed_param=self._is_fixed_param,
+            with open(self.runset.csv_files[chain], "rb") as f:
+                comments, draws = stancsv.parse_stan_csv_comments_and_draws(f)
+
+            self._draws[:, chain, :] = stancsv.csv_bytes_list_to_numpy(draws)
+
+            if not self._is_fixed_param:
+                (
+                    self._step_size[chain],
+                    mass_matrix,
+                ) = stancsv.parse_hmc_adaptation_lines(comments)
+                mass_matrix_per_chain.append(mass_matrix)
+
+        if mass_matrix_per_chain[0] is not None:
+            mm_shape = mass_matrix_per_chain[0].shape
+            if self.metric_type == "diag_e":
+                mm_shape = mm_shape[1:]
+            self._metric = np.empty(
+                (self.chains, *mm_shape),
+                dtype=np.float32,
             )
-            self._step_size[chain] = parsed_csv.step_size
-            if self._save_warmup and parsed_csv.warmup_draws is not None:
-                self._draws[:, chain, :] = np.concatenate(
-                    [parsed_csv.warmup_draws, parsed_csv.sampling_draws]
-                )
-            else:
-                self._draws[:, chain, :] = parsed_csv.sampling_draws
-
-            if parsed_csv.mass_matrix is not None:
-                if chain == 0:
-                    mm_shape = parsed_csv.mass_matrix.shape
-                    if self.metric_type == "diag_e":
-                        mm_shape = mm_shape[1:]
-                    self._metric = np.empty(
-                        (self.chains, *mm_shape),
-                        dtype=np.float32,
-                    )
-
-                self._metric[chain] = parsed_csv.mass_matrix
+            for chain in range(self.chains):
+                self._metric[chain] = mass_matrix_per_chain[chain]
 
         assert self._draws is not None
 
--- a/cmdstanpy/stanfit/mcmc.py
+++ b/cmdstanpy/stanfit/mcmc.py
@@ -39,6 +39,7 @@ from cmdstanpy.utils import (
     do_command,
     flatten_chains,
     get_logger,
+    stancsv,
 )
 
 from .metadata import InferenceMetadata
@@ -429,83 +430,42 @@ class CmdStanMCMC:
         """
         if self._draws.shape != (0,):
             return
+
         num_draws = self.num_draws_sampling
-        sampling_iter_start = 0
         if self._save_warmup:
             num_draws += self.num_draws_warmup
-            sampling_iter_start = self.num_draws_warmup
         self._draws = np.empty(
             (num_draws, self.chains, len(self.column_names)),
             dtype=float,
             order='F',
         )
         self._step_size = np.empty(self.chains, dtype=float)
-        for chain in range(self.chains):
-            with open(self.runset.csv_files[chain], 'r') as fd:
-                line = fd.readline().strip()
-                # read initial comments, CSV header row
-                while len(line) > 0 and line.startswith('#'):
-                    line = fd.readline().strip()
-                if not self._is_fixed_param:
-                    # handle warmup draws, if any
-                    if self._save_warmup:
-                        for i in range(self.num_draws_warmup):
-                            line = fd.readline().strip()
-                            xs = line.split(',')
-                            self._draws[i, chain, :] = [float(x) for x in xs]
-                    line = fd.readline().strip()
-                    if line != '# Adaptation terminated':  # shouldn't happen?
-                        while line != '# Adaptation terminated':
-                            line = fd.readline().strip()
-                    # step_size, metric (diag_e and dense_e only)
-                    line = fd.readline().strip()
-                    _, step_size = line.split('=')
-                    self._step_size[chain] = float(step_size.strip())
-                    if self._metadata.cmdstan_config['metric'] != 'unit_e':
-                        line = fd.readline().strip()  # metric type
-                        line = fd.readline().lstrip(' #\t').rstrip()
-                        num_unconstrained_params = len(line.split(','))
-                        if chain == 0:  # can't allocate w/o num params
-                            if self.metric_type == 'diag_e':
-                                self._metric = np.empty(
-                                    (self.chains, num_unconstrained_params),
-                                    dtype=float,
-                                )
-                            else:
-                                self._metric = np.empty(
-                                    (
-                                        self.chains,
-                                        num_unconstrained_params,
-                                        num_unconstrained_params,
-                                    ),
-                                    dtype=float,
-                                )
-                        if line:
-                            if self.metric_type == 'diag_e':
-                                xs = line.split(',')
-                                self._metric[chain, :] = [float(x) for x in xs]
-                            else:
-                                xs = line.strip().split(',')
-                                self._metric[chain, 0, :] = [
-                                    float(x) for x in xs
-                                ]
-                                for i in range(1, num_unconstrained_params):
-                                    line = fd.readline().lstrip(' #\t').rstrip()
-                                    xs = line.split(',')
-                                    self._metric[chain, i, :] = [
-                                        float(x) for x in xs
-                                    ]
-                    else:  # unit_e changed in 2.34 to have an extra line
-                        pos = fd.tell()
-                        line = fd.readline().strip()
-                        if not line.startswith('#'):
-                            fd.seek(pos)
 
-                # process draws
-                for i in range(sampling_iter_start, num_draws):
-                    line = fd.readline().strip()
-                    xs = line.split(',')
-                    self._draws[i, chain, :] = [float(x) for x in xs]
+        for chain in range(self.chains):
+            parsed_csv = stancsv.StanCsvMCMC.from_csv(
+                self.runset.csv_files[chain],
+                is_fixed_param=self._is_fixed_param,
+            )
+            self._step_size[chain] = parsed_csv.step_size
+            if self._save_warmup and parsed_csv.warmup_draws is not None:
+                self._draws[:, chain, :] = np.concat(
+                    [parsed_csv.warmup_draws, parsed_csv.sampling_draws]
+                )
+            else:
+                self._draws[:, chain, :] = parsed_csv.sampling_draws
+
+            if parsed_csv.mass_matrix is not None:
+                if chain == 0:
+                    mm_shape = parsed_csv.mass_matrix.shape
+                    if self.metric_type == "diag_e":
+                        mm_shape = mm_shape[1:]
+                    self._metric = np.empty(
+                        (self.chains, *mm_shape),
+                        dtype=np.float32,
+                    )
+
+                self._metric[chain] = parsed_csv.mass_matrix
+
         assert self._draws is not None
 
     def summary(
--- a/cmdstanpy/stanfit/mcmc.py
+++ b/cmdstanpy/stanfit/mcmc.py
@@ -436,10 +436,10 @@ class CmdStanMCMC:
             num_draws += self.num_draws_warmup
         self._draws = np.empty(
             (num_draws, self.chains, len(self.column_names)),
-            dtype=float,
+            dtype=np.float32,
             order='F',
         )
-        self._step_size = np.empty(self.chains, dtype=float)
+        self._step_size = np.empty(self.chains, dtype=np.float32)
 
         mass_matrix_per_chain = []
         for chain in range(self.chains):
@@ -455,7 +455,7 @@ class CmdStanMCMC:
                 ) = stancsv.parse_hmc_adaptation_lines(comments)
                 mass_matrix_per_chain.append(mass_matrix)
 
-        if mass_matrix_per_chain[0] is not None:
+        if not self._is_fixed_param and mass_matrix_per_chain[0] is not None:
             mm_shape = mass_matrix_per_chain[0].shape
             if self.metric_type == "diag_e":
                 mm_shape = mm_shape[1:]
--- a/cmdstanpy/stanfit/mcmc.py
+++ b/cmdstanpy/stanfit/mcmc.py
@@ -444,8 +444,9 @@ class CmdStanMCMC:
 
         mass_matrix_per_chain: List[Optional[npt.NDArray[np.float64]]] = []
         for chain in range(self.chains):
-            with open(self.runset.csv_files[chain], "rb") as f:
-                comments, draws = stancsv.parse_stan_csv_comments_and_draws(f)
+            comments, draws = stancsv.parse_stan_csv_comments_and_draws(
+                self.runset.csv_files[chain]
+            )
 
             self._draws[:, chain, :] = stancsv.csv_bytes_list_to_numpy(draws)
 
--- a/cmdstanpy/stanfit/mcmc.py
+++ b/cmdstanpy/stanfit/mcmc.py
@@ -444,18 +444,25 @@ class CmdStanMCMC:
 
         mass_matrix_per_chain: List[Optional[npt.NDArray[np.float64]]] = []
         for chain in range(self.chains):
-            comments, draws = stancsv.parse_stan_csv_comments_and_draws(
-                self.runset.csv_files[chain]
-            )
+            try:
+                comments, draws = stancsv.parse_stan_csv_comments_and_draws(
+                    self.runset.csv_files[chain]
+                )
 
-            self._draws[:, chain, :] = stancsv.csv_bytes_list_to_numpy(draws)
+                self._draws[:, chain, :] = stancsv.csv_bytes_list_to_numpy(
+                    draws
+                )
 
-            if not self._is_fixed_param:
-                (
-                    self._step_size[chain],
-                    mass_matrix,
-                ) = stancsv.parse_hmc_adaptation_lines(comments)
-                mass_matrix_per_chain.append(mass_matrix)
+                if not self._is_fixed_param:
+                    (
+                        self._step_size[chain],
+                        mass_matrix,
+                    ) = stancsv.parse_hmc_adaptation_lines(comments)
+                    mass_matrix_per_chain.append(mass_matrix)
+            except Exception as exc:
+                raise ValueError(
+                    f"Parsing output from {self.runset.csv_files[chain]} failed"
+                ) from exc
 
         if all(mm is not None for mm in mass_matrix_per_chain):
             if self.metric_type == "diag_e":
--- a/cmdstanpy/stanfit/mcmc.py
+++ b/cmdstanpy/stanfit/mcmc.py
@@ -18,6 +18,7 @@ from typing import (
 )
 
 import numpy as np
+import numpy.typing as npt
 import pandas as pd
 
 try:
@@ -436,12 +437,12 @@ class CmdStanMCMC:
             num_draws += self.num_draws_warmup
         self._draws = np.empty(
             (num_draws, self.chains, len(self.column_names)),
-            dtype=np.float32,
+            dtype=np.float64,
             order='F',
         )
-        self._step_size = np.empty(self.chains, dtype=np.float32)
+        self._step_size = np.empty(self.chains, dtype=np.float64)
 
-        mass_matrix_per_chain = []
+        mass_matrix_per_chain: List[Optional[npt.NDArray[np.float64]]] = []
         for chain in range(self.chains):
             with open(self.runset.csv_files[chain], "rb") as f:
                 comments, draws = stancsv.parse_stan_csv_comments_and_draws(f)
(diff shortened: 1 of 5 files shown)


cmdstanpy/utils/stancsv.py
```diff
@@ -1,17 +1,216 @@
...
+    entering a section, the entry_action should be specified."""
+
+    action: Callable[[bytes], None]
+    entry_action: Optional[Callable[[bytes], None]] = None
+
+
+@dataclass
+class StanCsvMCMC:
+    """Class containing the parsed output of a Stan CSV file sourced
+    from the `sample` inference method."""
```
WardBrian: I'm a bit hesitant to introduce a new class per output type. I like the general look of this code and design, but I wonder if we couldn't flip it inside-out a bit to promote more re-use.

I'll try to elaborate on this in more comments below, but in short I wonder if the rules etc should be entirely in the `stanfit/` classes, and the code here much more generic

cmdstanpy/utils/stancsv.py
```diff
@@ -1,17 +1,216 @@
...
+        else:
+            rules = (
+                ParsingRule(action=config_lines.append),
+                ParsingRule(
+                    entry_action=add_header, action=warmup_lines.append
+                ),
+                ParsingRule(action=adaptation_lines.append),
+                ParsingRule(action=sampling_lines.append),
+                ParsingRule(action=timing_lines.append),
+            )
```
WardBrian: The ParsingRules idea is neat, but all of them ultimate end up being list appending. What if instead the function returned `tuple(list[list[bytes]], np.ndarray)` (or a dataclass containing those, I don't mind).

This could then be used by all the interfaces, and it leaves it up to, e.g., stanfit.mcmc to decide that the third element of that first list should contain the metric or whatever. Right now we have logic for the metric here, and then a separate `if metric is None`-style check in the consumer of this code, but it feels like they could be combined easier.

This removes the distinction between warmup and sampling draws, but that's artificial anyway, as the condition up in mcmc.py (`self._save_warmup and parsed_csv.warmup_draws is not None`) should always either be both true or both false, meaning there's no need to really differentiate them.

cmdstanpy/utils/stancsv.py
```diff
@@ -1,17 +1,216 @@
...
+    whether the bytes contains an initial header line."""
+    out = (
+        pl.read_csv(
+            io.BytesIO(b"".join(csv_bytes_list)), has_header=includes_header
+        )
+        .to_numpy()
+        .astype(np.float32)
+    )
+    # Telling the type checker we know the type is correct
+    return cast(npt.NDArray[np.float32], out)
```
WardBrian: My instinct is still to make `polars` optional to match what we do with xarray and ujson

cmdstanpy/utils/stancsv.py
```diff
@@ -1,17 +1,216 @@
...
+
+
+def parse_hmc_adaptation_lines(
```
WardBrian, line 102: Just a comment to say I really like the style of these two functions here, much preferrable to the `scan_*` style we've had before

WardBrian (commented):
Thanks @amas0. The code looks really good. 

My comments below are thoughts on a slightly different style that I think will make it more reusable across the different inference algorithms and try to consolidate the logic for each in one place. Let me know what you think!

cmdstanpy/utils/stancsv.py
amas0: I went ahead and made `polars` optional with fallback logic to `np.loadtxt` if it isn't present.

cmdstanpy/utils/stancsv.py
amas0: Yeah I kind of had a feeling this approach was a little too clever. When I started going down this path, I assumed that there would be more variation in the parsing across the different inference methods, but it's mostly the same.

Splitting the parsed outputs into non-draws and draws would be simple and then any per-method parsing would happen against comment lines. I can rework this a bit.

cmdstanpy/utils/stancsv.py
amas0: How about per output functions that handle the different comment lines present? Those can be defined in `utils/stancsv.py` and then within the `stanfit/` classes they just use those to handle extract the relevant In combination with the below comment. So within the `stanfit/mcmc.py` it could look like:

```python
with open(stan_csv, "rb") as f:
  comment_lines, draws: List[bytes], np.ndarray = stancsv.parse_general_stan_csv_from_lines(f)

config, step_size, mass_matrix, timings = stancsv.parse_mcmc_comment_lines(comment_lines)
```

I generally like the idea of the `stanfit/` classes not needing to concern themselves too much with IO logic, but some stuff like parsing of the adaptation comment blocks is unique to a particular inference method. At the very least logic for parsing the draws out, the config lines, and the timing lines all make sense to me to exist in the `utils/stancsv.py` space (as a side note, I think the config parsing that exists now probably should be cleaned up, but that'll be another day's task).

If we want to have the method-specific parsing logic, i.e. parsing the adaptation lines, to live within the `stanfit/` classes and the shared sections to use the shared utilities, I think that's reasonable.

Let me know what you think.

WardBrian: That looks pretty good to me. Alt. would be `mass_matrix = stancsv.parse_mcmc_metric(comments[3])`, but the `3` feels arbitrary there.

amas0: At that point I imagine its not even worth separating out the comments into sections and just have a `comment_lines: list[bytes]` that gets passed around to various parsing methods. Could imagine something like:

```py
config = stancsv.parse_config(comments)
# I assume these should go together because they come from the same block
step_size, mass_matrix = stancsv.parse_mcmc_adaptation(comments)
timings = stancsv.parse_timings(comments)
```

Not sure, something like this probably. I'll play around with some changes and come back with updates when I'm happy. Thanks for the feedback.

amas0:
Okay, I think this PR should be good to go for more review/feedback. In response to the previous comments, I've made the following changes from the last version:

* Made `polars` an optional dependency with fallback-parsing done by `numpy`
* Added unit tests against the new `utils/stancsv` functions (and caught some bugs)
* Removed the `ParsingRule`s based parser in favor of a simpler structure that splits a CSV into comments/draws lines
* Comment section parsing functions like timing/adaptation parsing expect to be provided the full comment lines
* Remove `StanCsvMCMC` in favor of using the utility functions within the `CmdStanMCMC._assemble_draws()` method

This version is simpler and cleaner than the previous. I still think there's probably some cleanup that can happen in the `CmdStanMCMC._assemble_draws()` method, but at least I think it's improved compared to the existing version.

I ran one last bit of performance testing comparing the new `numpy`/`polars` based parsing against the existing method:

image

This is with the defaults of 4 chains, 1000 draws per chain. I didn't stretch the test out to the extremes of number of parameters, but for moderate to large model sizes, the speedup should be noticeable.

cmdstanpy/utils/stancsv.py
```diff
@@ -1,17 +1,150 @@
...
+        if line.startswith(b"#"):  # is comment line
+            comment_lines.append(line)
+        else:
+            draws_lines.append(line)
+    return comment_lines, draws_lines
+
+
+def csv_bytes_list_to_numpy(
+    csv_bytes_list: List[bytes], includes_header: bool = True
+) -> npt.NDArray[np.float32]:
```
WardBrian: We should always load these as float64. With the default CSV output it's not that noticeable, but people can request all 16 digits if they wanted to.

cmdstanpy/utils/stancsv.py
```diff
@@ -1,17 +1,150 @@
...
+    """Efficiently converts a list of bytes representing whose concatenation
+    represents a CSV file into a numpy array. Includes header specifies
+    whether the bytes contains an initial header line."""
+    try:
+        import polars as pl
+
+        try:
+            out = (
+                pl.read_csv(
+                    io.BytesIO(b"".join(csv_bytes_list)),
```
WardBrian, line 72: should I be surprised that this isn't `b'\n'`?

cmdstanpy/utils/stancsv.py
```diff
@@ -1,17 +1,150 @@
...
+            warnings.filterwarnings("ignore")
+            out = np.loadtxt(
+                csv_bytes_list,
+                skiprows=int(includes_header),
+                delimiter=",",
+                dtype=np.float32,
+                ndmin=1,
+            )
+        if out.shape == (0,):
+            raise ValueError("No data found to parse") from exc
```
WardBrian: the `from exc` seems useless here. If it was caused by a pylint error you can safely `# ignore` it I think

cmdstanpy/utils/stancsv.py
```diff
@@ -1,17 +1,150 @@
...
+        if len(out.shape) == 1:
+            out = out.reshape(1, -1)
+
+    # Telling the type checker we know the type is correct
+    return cast(npt.NDArray[np.float32], out)
```
WardBrian: Casts have a (small) runtime cost and generally defeat the purpose of typechecking. It's equivalent to `# type: ignore`, which I would prefer if we cant make the type work out some other way (e.g. by declaring it when `out` is defined?)

cmdstanpy/stanfit/mcmc.py
```diff
@@ -429,83 +430,42 @@ def _assemble_draws(self) -> None:
...
-                        if not line.startswith('#'):
-                            fd.seek(pos)
-
-                # process draws
-                for i in range(sampling_iter_start, num_draws):
-                    line = fd.readline().strip()
-                    xs = line.split(',')
-                    self._draws[i, chain, :] = [float(x) for x in xs]
+            with open(self.runset.csv_files[chain], "rb") as f:
+                comments, draws = stancsv.parse_stan_csv_comments_and_draws(f)
```
WardBrian: Should the `open` be inside the stancsv function?

cmdstanpy/stanfit/mcmc.py
```diff
@@ -429,83 +430,42 @@ def _assemble_draws(self) -> None:
...
+
         num_draws = self.num_draws_sampling
-        sampling_iter_start = 0
         if self._save_warmup:
             num_draws += self.num_draws_warmup
-            sampling_iter_start = self.num_draws_warmup
         self._draws = np.empty(
             (num_draws, self.chains, len(self.column_names)),
-            dtype=float,
+            dtype=np.float32,
```
WardBrian:
```suggestion
            dtype=np.float64,
```

cmdstanpy/stanfit/mcmc.py
```diff
@@ -429,83 +430,42 @@ def _assemble_draws(self) -> None:
...
+        if not self._is_fixed_param and mass_matrix_per_chain[0] is not None:
+            mm_shape = mass_matrix_per_chain[0].shape
+            if self.metric_type == "diag_e":
+                mm_shape = mm_shape[1:]
+            self._metric = np.empty(
+                (self.chains, *mm_shape),
+                dtype=np.float32,
+            )
+            for chain in range(self.chains):
+                self._metric[chain] = mass_matrix_per_chain[chain]
```
WardBrian: I think this is equivalent to

```python
if mass_matrix_per_chain[0] is not None:
    self._metric = np.array(mass_matrix_per_chain)
```

cmdstanpy/stanfit/mcmc.py
```diff
@@ -429,83 +430,42 @@ def _assemble_draws(self) -> None:
...
+                comments, draws = stancsv.parse_stan_csv_comments_and_draws(f)
+
+            self._draws[:, chain, :] = stancsv.csv_bytes_list_to_numpy(draws)
+
+            if not self._is_fixed_param:
+                (
+                    self._step_size[chain],
+                    mass_matrix,
+                ) = stancsv.parse_hmc_adaptation_lines(comments)
+                mass_matrix_per_chain.append(mass_matrix)
```
WardBrian: Thoughts on wrapping these in a try/catch, and then re-raising with the filename if something goes wrong? Otherwise users will just see e.g. `No data found to parse` with no way of knowing which file was bad

test/test_stancsv.py
```diff
@@ -0,0 +1,307 @@
...
+    with pytest.raises(ValueError):
+        stancsv.csv_bytes_list_to_numpy(lines)
+
+
+def test_csv_bytes_to_numpy_empty_no_polars():
+    lines = [b""]
+    with mock.patch.dict("sys.modules", {"polars": None}):
+        with pytest.raises(ValueError):
+            stancsv.csv_bytes_list_to_numpy(lines)
+
```
WardBrian: We have a helper function for this, see e.g.
https://github.com/stan-dev/cmdstanpy/blob/8b37adb96b33a7feac088897060687235f0288f6/test/test_sample.py#L1843-L1861

cmdstanpy/utils/stancsv.py
```diff
@@ -1,17 +1,150 @@
...
+            _, ss_str = line.split(b" = ")
+            step_size = float(ss_str)
+        elif line.startswith(b"Diagonal") or line.startswith(b"Elements"):
+            in_matrix_block = True
+        elif line.startswith(b"No free"):
+            break
+    if step_size is None:
+        raise ValueError("Unable to parse adapated step size")
+    if matrix_lines:
+        mass_matrix = csv_bytes_list_to_numpy(
```
WardBrian, line 136: [comment] already makes me happy to see this is re-using code more than what came before!

cmdstanpy/utils/stancsv.py
```diff
@@ -1,17 +1,150 @@
...
+            continue
+        match = re.findall(r"([\d\.]+) seconds \((.+)\)", str(line))
+        if match:
+            seconds = float(match[0][0])
+            phase = match[0][1]
+            out[phase] = seconds
+    return out
+
+
 def check_sampler_csv(
```
WardBrian, line 144: Just noting that, eventually, I'd like to re-write these `check_` functions to consume the same kind of input as the above `parse_` functions (if we don't replace them entirely with a JSON read first, c.f. #685).

Not important for this PR, really just a note to myself.

WardBrian (changes requested):
This is looking really good to me now, thanks!

Don't let the number of comments here be discouraging, most are very small changes. Structurally I think this is really clean now :)

cmdstanpy/utils/stancsv.py
amas0: Good catch -- updated. I had been working only off the default CSV and f32 seemed fine.

cmdstanpy/utils/stancsv.py
amas0: When using a file object as an iterator in Python, e.g. 

```py
with open('file.txt', 'r') as f:
  for line in f:
    foo(line)
```

the line terminators are kept at the end of each `line` (this is also true of `.readlines()`). So the partitioned lines all include `b"\n"` at the end already, so we just join on an empty byte to recreate the proper multi-line structure expected by the `read_csv`.

cmdstanpy/utils/stancsv.py
amas0: Removed and added a pylint ignore comment

cmdstanpy/utils/stancsv.py
amas0: Should be fixed up now -- `cast` removed. Adding `npt.NDArray[np.float64]` to the declaration of `out` from the polars bit seems to have resolved it.

cmdstanpy/stanfit/mcmc.py
amas0: This is something I went back and forth on. I could add in the relatively common interface where `parse_stan_csv_comments_and_draws` can accept `Union[str, Path, Iterator[bytes]]` (or something like that) and then if it gets the path-like object, it'll first open the file. Thoughts on that?

That way it would still work in other cases like parsing the mass matrix where we aren't coming directly from a file.

test/test_stancsv.py
amas0: Neat -- moved over to using that.

cmdstanpy/stanfit/mcmc.py
WardBrian: That would be fine by me. Though, note that

> That way it would still work in other cases like parsing the mass matrix where we aren't coming directly from a file.

The metric is never calling this function, it's using csv_bytes_list_to_numpy, which I agree should _not_ accept a file

amas0: Woops, misread my own code when I made this comment.

Updated the `parse_stan_csv_comments_and_draws` to be able to accept a file path as well and removed the `open` within the stanfit function

cmdstanpy/stanfit/mcmc.py
amas0: I was able to clean this bit up. My original parsing code of adaptation lines always returned a 2D array (even in the case of the `diag_e` metric) because that's what the csv parsing code does. I realized this could just be handled in the `parse_hmc_adaptation_lines` code, so now if the `diag_e` metric is used, the parsing will return a 1D array, which allow this to now work:

```py
if all(mm is not None for mm in mass_matrix_per_chain):
    self._metric = np.array(mass_matrix_per_chain)
```

The `all(...)` here will also handle the case when `mass_matrix_per_chain` is an empty list, which happens in the fixed param case.

cmdstanpy/stanfit/mcmc.py
amas0: Done! 

I now raise a `ValueError` with the file name included if any exceptions are thrown in the csv parsing step. It's raised `from exc` with the original exception so the underlying error should propagate through.

cmdstanpy/stanfit/mcmc.py
amas0: Everything should be `np.float64` now.

cmdstanpy/utils/stancsv.py
amas0: Yeah I think there's a good amount of space to clean up some of these older style functions in light of the new stuff here.

amas0:
Okay @WardBrian, I think that's everything addressed? Thanks for the feedback, definitely improved things in a few areas. 

Let me know if there's anything else that stands out from the additions. All test look good (and I added some new ones).

cmdstanpy/utils/stancsv.py
```diff
@@ -1,17 +1,165 @@
...
+    ) -> Tuple[List[bytes], List[bytes]]:
+        comment_lines, draws_lines = [], []
+        for line in lines:
+            if line.startswith(b"#"):  # is comment line
+                comment_lines.append(line)
+            else:
+                draws_lines.append(line)
+        return comment_lines, draws_lines
+
+    if isinstance(stan_csv, (str, Path)):
```
WardBrian: [minor] I think it's preferred to use os.PathLike rather than Path for this sort of check

cmdstanpy/utils/stancsv.py
```diff
@@ -1,17 +1,165 @@
...
+        import polars as pl
+
+        try:
+            out: npt.NDArray[np.float64] = (
+                pl.read_csv(
+                    io.BytesIO(b"".join(csv_bytes_list)),
+                    has_header=includes_header,
+                )
+                .to_numpy()
+                .astype(np.float64)
```
WardBrian, line 78: Rather than doing this afterwards, should we be providing a `schema` (or `schema_override`, can't really tell the difference) argument?

cmdstanpy/utils/stancsv.py
```diff
@@ -1,17 +1,165 @@
...
+    if matrix_lines:
+        mass_matrix = csv_bytes_list_to_numpy(
+            matrix_lines, includes_header=False
+        )
+        if diag_e_metric and mass_matrix.shape[0] == 1:
+            mass_matrix = mass_matrix[0]
+    return step_size, mass_matrix
+
+
+def parse_timing_lines(
```
WardBrian: It looks like the only place this new function is called is in the tests?

WardBrian (approved):
Thanks @amas0!

I have two small questions before merging, but this looks really great

cmdstanpy/utils/stancsv.py
amas0: Updated

cmdstanpy/utils/stancsv.py
amas0: I originally didn't the `schema_override` because it requires knowing the number of columns before parsing, but it turns out that you get even more performance out of this is if you do this and disable schema inference entirely. The below is for converting 1000 x 10007 to a numpy array:

```py
%%timeit
n_cols = draws[0].count(b',') + 1
pl.read_csv(io.BytesIO(b"".join(draws)),
            has_header=True,
            schema_overrides=[pl.Float64]*n_cols,
            infer_schema=False
            ).to_numpy()
```
175 ms ± 990 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)


```py
%%timeit
pl.read_csv(io.BytesIO(b"".join(draws)), has_header=True).to_numpy().astype(np.float64)
```

225 ms ± 4.48 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

Updated.

WardBrian: My guess is a lot of that speed is less about not trying to infer the schema and more from the fact that polars probably (correctly, I might add) determines that some columns are integers rather than floats, and then the cast actually has to do the conversion. By telling it ahead of time we will be interpreting them as floats always, it avoids the post-hoc recasting

cmdstanpy/utils/stancsv.py
amas0: Yeah, this is somewhat vestigial. The parsed timing data was included back when I had the class-based output from the StanCSV. I figured it might be something that could be useful to include as an attribute in the stanfit classes, but it's not something that's used anywhere now. 

Want me to just cut this out and if it's needed in the future, it can be re-implemented?

WardBrian: We currently do read in the output timings elsewhere, so if that code can be replaced with this I'm happy to keep it, otherwise I would remove before merging, yes.

cmdstanpy/utils/stancsv.py
amas0: Actually if you drop the `infer_schema=False` kwarg, then you lose almost all the benefit:

```py
%%timeit
n_cols = draws[0].count(b',') + 1
pl.read_csv(io.BytesIO(b"".join(draws)),
            has_header=True,
            schema_overrides=[pl.Float64]*n_cols,
            ).to_numpy()
```
220 ms ± 4.65 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

I imagine the casting does have a cost, but at least in my test case where its 10007 columns and only `treedepth__, n_leapfrog__,divergent__` are inferred as ints, it doesn't make up the bulk. It would probably be a different story if a large number of columns happened to be `int`s but I can't imagine that'd be too common?

WardBrian: It depends, people can have a lot of generated quantities that are ints, but certainly not the normal case.

Very interesting!

cmdstanpy/utils/stancsv.py
amas0: The existing timing parsing is pretty deep in some of the `scan_` functions (and not so straightforwardly replaced without digging into those details), so I think it makes sense to just remove this function for now, which I've done.

WardBrian:
Thanks again @amas0!

Looking forward to these utilities spreading to the other methods as well

cmdstanpy/utils/stancsv.py
ahartikainen: is the difference in reading file data or converting to numpy (homogenous vs heterogenous data -> more copying with heterogenous)?

amas0: The conversion to numpy should be very lightweight for polars, their `to_numpy()` [docs](https://docs.pola.rs/api/python/stable/reference/dataframe/api/polars.DataFrame.to_numpy.html) suggest copying data only when necessary. I would guess that with everything inferred as Float64, it would meet the conditions for zero copy. In the heterogeneous case, I wouldn't be surprised if it were only necessary to copy the non-float columns?

In either case, I would guess the main difference is in reading the bytes into the table. My understanding is polars will peek at the first few rows of the csv, attempt to infer the dtype for each column, and then load it in to the table. 

Just kind of speculating though.

PR #804, opened 2025-08-14
Deprecate metric argument and property, rename to inv_metric

### LINKED ISSUE #803: Name confusion between inverse metric, metric
**EDIT:  This is worse than I thought.**

If I take a fitted model and do this:

```
fit = model.sample(...)
metric = fit.metric
```

then the right way to restart is as follows.

```python
    metric={"inv_metric": metric}`
```

This is confusing because of the mismatch.  Something must be misnamed.  If I use `1 / metric` instead of `metric` to get the inverse of what `fit.metric` returns, then the sampler grinds to a halt because of the anti-conditioning.

What I would really like is to be able to write is the following.

```python
    metric=metric
```

It would be helpful to call out the various cases of specifying an (inverse) metric with bullets the same way `inits` is documented---as is, it's hard to visually identify the cases.

WardBrian:
#### Submission Checklist

- [x] Run unit tests

#### Summary

Closes #803 

This renames the `metric` property of `CmdStanMCMC` to `inv_metric`, with the old property still available but raising a warning.

Secondly, it updates how the initial inverse metric is provided. The pre-existing `metric` argument is split into two, `metric` which still exists to specify only the metric type (e.g. `"diag_e"`), and a new `inv_metric` which specifies the value. This handling is written from-scratch using some of the helpers that already existed for e.g. init files. Besides the existing types, it can also now accept numpy arrays, meaning it is valid to write

`model.sample(..., inv_metric=old_fit.inv_metric)`

I am not completely confident that the error handling is as robust as previous, it's probably possible to come up with a bad argument that wouldn't be caught until actually running cmdstan, rather than trapped as a nicer error in the Python level.

#### Copyright and Licensing

Simons Foundation

### DIFF
--- a/cmdstanpy/model.py
+++ b/cmdstanpy/model.py
@@ -1009,7 +1009,13 @@ class CmdStanModel:
                             'Chain_id must be a non-negative integer value,'
                             ' found {}.'.format(chain_id)
                         )
-        if metric not in [None, 'diag', 'dense', 'unit_e', 'diag_e', 'dense_e']:
+        if metric is not None and metric not in (
+            'diag',
+            'dense',
+            'unit_e',
+            'diag_e',
+            'dense_e',
+        ):
             get_logger().warning(
                 "Providing anything other than metric type for"
                 " 'metric' is deprecated and will be removed"
--- a/cmdstanpy/model.py
+++ b/cmdstanpy/model.py
@@ -26,6 +26,7 @@ from typing import (
     Union,
 )
 
+import numpy as np
 import pandas as pd
 from tqdm.auto import tqdm
 
@@ -57,7 +58,12 @@ from cmdstanpy.utils import (
     get_logger,
     returncode_msg,
 )
-from cmdstanpy.utils.filesystem import temp_inits, temp_single_json
+from cmdstanpy.utils.filesystem import (
+    temp_inits,
+    temp_metrics,
+    temp_single_json,
+)
+from cmdstanpy.utils.stancsv import try_deduce_metric_type
 
 from . import progress as progbar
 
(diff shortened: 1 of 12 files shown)


WardBrian:
@mitzimorris I'd appreciate a review if you are able!

cmdstanpy/model.py
```diff
@@ -785,29 +798,23 @@ def sample(
...
-            vector consisting of the diagonal elements of the covariance
-            matrix ('diag' or 'diag_e') or the full covariance matrix
-            ('dense' or 'dense_e').
-
-            If the value of the metric argument is a string other than
-            'diag', 'diag_e', 'dense', or 'dense_e', it must be
-            a valid filepath to a JSON or Rdump file which contains an entry
-            'inv_metric' whose value is either the diagonal vector or
-            the full covariance matrix.
+        :param metric: Specify the type of the mass matrix. Options are
```
mitzimorris: type of the inverse mass matrix?

cmdstanpy/stanfit/mcmc.py
```diff
@@ -223,19 +223,27 @@ def metric_type(self) -> Optional[str]:
             else None
         )
 
+    # TODO(2.0): remove
     @property
     def metric(self) -> Optional[np.ndarray]:
+        """Deprecated. Use ``.inv_metric`` instead."""
+        get_logger().warning(
+            'The "metric" property is deprecated, use "inv_metric" instead. '
+            'This will be the same quantity, but with a more accurate name.'
```
mitzimorris, line 232: very well put!

mitzimorris (changes requested):
If I understand correctly:  the metric is specified by the inverse mass matrix which is either unit, dense, or diagonal,  therefore the only property name we change is "metric" to "inv_metric" which holds, directly or indirectly, a matrix or list of matrices?

Therefore we should say "inverse mass matrix" in the doc comments and messages when talking about the matrix and its contents.

Otherwise this looks good and I like the encapsulated functions for dealing with metric files.

cmdstanpy/model.py
WardBrian: Corrected!

bob-carpenter:
The mass matrix and metric are the same matrix.  For a multivariate normal, the best value is the covariance, as that preconditions the multivariate normal back to unit covariance.  See page 31 of Betancourt's [conceptual intro to HMC](https://arxiv.org/pdf/1701.02434).

For years, I'd had a bunch of this turned around in my head and in our doc and on the board as in the Stan meeting.  Sorry!

mitzimorris (approved):
LGTM!

PR #806, opened 2025-08-16
Additional Stan CSV IO updates and refactoring

amas0:
#### Submission Checklist

- [x] Run unit tests

#### Summary

This is the second PR addressing #785 coming after #799. The main proposed changes in this PR are:

* Update Laplace, Pathfinder, Variational, MLE, and GQ draw parsing to use optimized parser introduced in #799.
* Add new CSV parsing logic for comment blocks that conforms to a more declarative style (and is hopefully a bit more legible)
* Complete removal of existing side-effect oriented `scan_` CSV parsing functions
* Update the CSV validation logic that exists as part of the `CmdStanMCMC` initialization

There's quite a bit going on in these changes. I think there are still a number of areas in the codebase that could be cleaned up, but I want to avoid these PRs getting too large.

#### Copyright and Licensing

### DIFF
--- a/cmdstanpy/stanfit/__init__.py
+++ b/cmdstanpy/stanfit/__init__.py
@@ -104,7 +104,7 @@ def from_csv(
             )
 
     try:
-        comments, _ = stancsv.parse_stan_csv_comments_and_draws(csvfiles[0])
+        comments, *_ = stancsv.parse_comments_header_and_draws(csvfiles[0])
         config_dict = stancsv.parse_config(comments)
     except (IOError, OSError, PermissionError) as e:
         raise ValueError('Cannot read CSV file: {}'.format(csvfiles[0])) from e
--- a/cmdstanpy/stanfit/__init__.py
+++ b/cmdstanpy/stanfit/__init__.py
@@ -117,42 +117,42 @@ def from_csv(
                 method, config_dict['method']
             )
         )
+    model: str = config_dict['model']  # type: ignore
     try:
         if config_dict['method'] == 'sample':
-            assert isinstance(config_dict['num_samples'], int)
-            assert isinstance(config_dict['num_warmup'], int)
-            assert isinstance(config_dict['thin'], int)
-            assert isinstance(config_dict['model'], str)
             save_warmup = config_dict['save_warmup'] == 1
             chains = len(csvfiles)
+            num_samples: int = config_dict['num_samples']  # type: ignore
+            num_warmup: int = config_dict['num_warmup']  # type: ignore
+            thin: int = config_dict['thin']  # type: ignore
             sampler_args = SamplerArgs(
-                iter_sampling=config_dict['num_samples'],
-                iter_warmup=config_dict['num_warmup'],
-                thin=config_dict['thin'],
+                iter_sampling=num_samples,
+                iter_warmup=num_warmup,
+                thin=thin,
                 save_warmup=save_warmup,
             )
             # bugfix 425, check for fixed_params output
             try:
                 check_sampler_csv(
                     csvfiles[0],
-                    iter_sampling=config_dict['num_samples'],
-                    iter_warmup=config_dict['num_warmup'],
-                    thin=config_dict['thin'],
+                    iter_sampling=num_samples,
+                    iter_warmup=num_warmup,
+                    thin=thin,
                     save_warmup=save_warmup,
                 )
             except ValueError:
                 try:
                     check_sampler_csv(
                         csvfiles[0],
-                        iter_sampling=config_dict['num_samples'],
-                        iter_warmup=config_dict['num_warmup'],
-                        thin=config_dict['thin'],
+                        iter_sampling=num_samples,
+                        iter_warmup=num_warmup,
+                        thin=thin,
                         save_warmup=save_warmup,
                     )
                     sampler_args = SamplerArgs(
-                        iter_sampling=config_dict['num_samples'],
-                        iter_warmup=config_dict['num_warmup'],
-                        thin=config_dict['thin'],
+                        iter_sampling=num_samples,
+                        iter_warmup=num_warmup,
+                        thin=thin,
                         save_warmup=save_warmup,
                         fixed_param=True,
                     )
@@ -162,8 +162,8 @@ def from_csv(
                     ) from e
 
             cmdstan_args = CmdStanArgs(
-                model_name=config_dict['model'],
-                model_exe=config_dict['model'],
+                model_name=model,
+                model_exe=model,
                 chain_ids=[x + 1 for x in range(chains)],
                 method_args=sampler_args,
             )
@@ -180,19 +180,18 @@ def from_csv(
                     "Cannot find optimization algorithm"
                     " in file {}.".format(csvfiles[0])
                 )
-            assert isinstance(config_dict['algorithm'], str)
-            assert isinstance(config_dict['model'], str)
+            algorithm: str = config_dict['algorithm']  # type: ignore
             save_iterations = config_dict['save_iterations'] == 1
             jacobian = config_dict.get('jacobian', 0) == 1
 
             optimize_args = OptimizeArgs(
-                algorithm=config_dict['algorithm'],
+                algorithm=algorithm,
                 save_iterations=save_iterations,
                 jacobian=jacobian,
             )
             cmdstan_args = CmdStanArgs(
-                model_name=config_dict['model'],
-                model_exe=config_dict['model'],
+                model_name=model,
+                model_exe=model,
                 chain_ids=None,
                 method_args=optimize_args,
             )
@@ -207,28 +206,19 @@ def from_csv(
                     "Cannot find variational algorithm"
                     " in file {}.".format(csvfiles[0])
                 )
-            assert isinstance(config_dict['model'], str)
-            assert isinstance(config_dict['algorithm'], str)
-            assert isinstance(config_dict['iter'], int)
-            assert isinstance(config_dict['grad_samples'], int)
-            assert isinstance(config_dict['elbo_samples'], int)
-            assert isinstance(config_dict['eta'], (int, float))
-            assert isinstance(config_dict['tol_rel_obj'], float)
-            assert isinstance(config_dict['eval_elbo'], int)
-            assert isinstance(config_dict['output_samples'], int)
             variational_args = VariationalArgs(
-                algorithm=config_dict['algorithm'],
-                iter=config_dict['iter'],
-                grad_samples=config_dict['grad_samples'],
-                elbo_samples=config_dict['elbo_samples'],
-                eta=config_dict['eta'],
-                tol_rel_obj=config_dict['tol_rel_obj'],
-                eval_elbo=config_dict['eval_elbo'],
-                output_samples=config_dict['output_samples'],
+                algorithm=config_dict['algorithm'],  # type: ignore
+                iter=config_dict['iter'],  # type: ignore
+                grad_samples=config_dict['grad_samples'],  # type: ignore
+                elbo_samples=config_dict['elbo_samples'],  # type: ignore
+                eta=config_dict['eta'],  # type: ignore
+                tol_rel_obj=config_dict['tol_rel_obj'],  # type: ignore
+                eval_elbo=config_dict['eval_elbo'],  # type: ignore
+                output_samples=config_dict['output_samples'],  # type: ignore
             )
             cmdstan_args = CmdStanArgs(
-                model_name=config_dict['model'],
-                model_exe=config_dict['model'],
+                model_name=model,
+                model_exe=model,
                 chain_ids=None,
                 method_args=variational_args,
             )
@@ -238,18 +228,15 @@ def from_csv(
                 runset._set_retcode(i, 0)
             return CmdStanVB(runset)
         elif config_dict['method'] == 'laplace':
-            assert isinstance(config_dict['mode'], str)
-            assert isinstance(config_dict['draws'], int)
-            assert isinstance(config_dict['model'], str)
             jacobian = config_dict['jacobian'] == 1
             laplace_args = LaplaceArgs(
-                mode=config_dict['mode'],
-                draws=config_dict['draws'],
+                mode=config_dict['mode'],  # type: ignore
+                draws=config_dict['draws'],  # type: ignore
                 jacobian=jacobian,
             )
             cmdstan_args = CmdStanArgs(
-                model_name=config_dict['model'],
-                model_exe=config_dict['model'],
+                model_name=model,
+                model_exe=model,
                 chain_ids=None,
                 method_args=laplace_args,
             )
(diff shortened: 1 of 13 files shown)


cmdstanpy/stanfit/metadata.py
```diff
@@ -38,6 +55,12 @@ def cmdstan_config(self) -> Dict[str, Any]:
         """
         return copy.deepcopy(self._cmdstan_config)
 
+    @property
+    def column_names(self) -> Tuple[str, ...]:
+        col_names = self['column_names']
+        assert isinstance(col_names, tuple)
```
WardBrian: Just `# type: ignore` for this IMO

cmdstanpy/utils/stancsv.py
```diff
@@ -140,234 +138,384 @@ def parse_hmc_adaptation_lines(
...
+    comment_lines: List[bytes], draws_lines: List[bytes]
+) -> Dict[str, Union[str, int, float, Tuple[str, ...]]]:
+    """Extracts config and header info from comment/draws lines parsed
+    from a Stan CSV file."""
+    config = parse_config(comment_lines)
+    raw_header = extract_header_line(draws_lines)
+    return {
+        **config,
+        **{"raw_header": raw_header, "column_names": parse_header(raw_header)},
+    }
```
WardBrian: This function and the error logic in `extract_header_line` make me wonder if `parse_stan_csv_comments_and_draws` should return `(comments, header, draws)` instead?

This would have the nice benefit of changing the `csv_bytes_list_to_numpy` to not need the `includes_header` flag.

What do you think?

cmdstanpy/stanfit/__init__.py
```diff
@@ -199,6 +207,15 @@ def from_csv(
...
                 )
+            assert isinstance(config_dict['model'], str)
+            assert isinstance(config_dict['algorithm'], str)
+            assert isinstance(config_dict['iter'], int)
+            assert isinstance(config_dict['grad_samples'], int)
+            assert isinstance(config_dict['elbo_samples'], int)
+            assert isinstance(config_dict['eta'], (int, float))
+            assert isinstance(config_dict['tol_rel_obj'], float)
+            assert isinstance(config_dict['eval_elbo'], int)
+            assert isinstance(config_dict['output_samples'], int)
```
WardBrian: I don't love all these asserts. If they're just for typing purposes, I think we can just ignore them, and if they're actually to catch errors we should be doing something different IMO

cmdstanpy/stanfit/gq.py
```diff
@@ -99,48 +102,39 @@ def __getstate__(self) -> dict:
...
+            'id',
+            'fitted_params',
+            'diagnostic_file',
+            'metric_file',
+            'profile_file',
+            'init',
+            'seed',
+            'start_datetime',
+        }
+        assert self.chains >= 1  # Should be impossible
```
WardBrian: Probably fine to just remove, but if you want to keep it the right place to fail is probably in the constructor

cmdstanpy/utils/stancsv.py
```diff
@@ -66,19 +69,16 @@ def csv_bytes_list_to_numpy(
...
-                .astype(np.float64)
+        out: npt.NDArray[np.float64] = (
+            pl.read_csv(
+                io.BytesIO(b"".join(csv_bytes_list)),
+                has_header=includes_header,
+                schema_overrides=[pl.Float64] * num_cols,
+                infer_schema=False,
             )
-        except pl.exceptions.NoDataError:
-            return np.empty((0, num_cols))
```
WardBrian: Can this not be raised any more?

WardBrian (changes requested):
This is pretty heroic but I also think it looks good at first glance. 

I have a few starter questions to prompt some discussion and possibly a few small changes before diving all the way in:

cmdstanpy/stanfit/__init__.py
amas0: Yeah, these aren't pretty. 

I dropped them all in favor of just some ignores. Going through the exercise of looking at these did turn up a few instances where bool-typed arguments were receiving ints (as 0/1), which is very minor given bools are subclasses of ints, but still.

amas0:
Appreciate the initial glance. This turned out to be more work than I expected (it really gets into the weeds with all the csv validation logic that ultimately happens within the `stancsv.check_sampler_csv` function).

Now that I've been through much (maybe all?) of the code that interacts with the Stan csvs, it seems to me that the overall organization and ergonomics of working with this I/O would really benefit from some sort of "Stan output schema" that could be validated. Whether that's a schema that all the inference methods adhere to or method-specific schemas, something where, at initial load, full validation occurs and a structured, typed-representation is available for the rest of the library to use would be very valuable. (This seems like the kind of thing that would naturally go into the stanio package?)

In the code currently, validation is pretty inconsistent. The CmdStanMCMC class performs a significant amount of validation, but other inference methods are a pretty light-touch. Obviously, this corresponds with importance, but it'd be nice to close that gap. Also, working with the parsed outputs ends up being awkward in practice. We have a lot of key info being passed around in dicts that are mostly populated from parsing the config block in the Stan CSV files (but also adding additional fields like the headers (and parsed headers)). The result of this is having to do things like check for field existence and fight with the type checker (and mostly just ignore the warnings).

I think this PR is an improvement in clarity, but it still struggles with the same fundamental challenges that working with these CSVs have. In #785, you mention newer JSON outputs and possibly future binary outputs for draws; I think changes like that would go a long way to cleaning this up -- something like that with a defined schema would be nice.

cmdstanpy/utils/stancsv.py
amas0: The original way I intended it to be raised was essentially unreachable. 

Looking at it again, the case where it _could_ be hit is if something like `[b"\n"]` were passed in, so no data/headers but a non-empty list. I re-added this handling to return an empty array in those cases.

cmdstanpy/utils/stancsv.py
amas0: I like this idea -- I'll take a look

WardBrian:
Yeah, it would be fairly natural to define a schema for the output_config.json and use that going forward. With the current CSV files, I actually lean toward relatively little validation, which you've observed with the other algorithms added more recently. Partially this is because they only have one output file. With MCMC it is fairly important to check that the different files all came from the same run, but this is just not a category of error that exists with e.g. optimization

bob-carpenter:
> the overall organization and ergonomics of working with this I/O would really benefit from some sort of "Stan output schema" 

That would be great, because lack of validation logic is one of the bigger obstacles to refactoring.  It'd also be nice to put it in standard I/O formats rather than CSV (not really a standard) plus comments (even worse).

cmdstanpy/utils/stancsv.py
amas0: Implemented this -- saved some lines overall and simplified `csv_bytes_list_to_numpy` which now doesn't have to deal with handling the possibility of a header line to parse. Makes that bit a little clear. 

Some edge case behavior like an header with no draws desiring to produce an array with shape `(0, n_cols)` now needs to be handled elsewhere (in the mcmc case, it's done in `_assemble_draws`. I think this make more sense rather than trying to do it via inspection of a header line when constructing an array of the draws.

amas0:
Implemented some changes in response to the initial comments.

@WardBrian when you say:

> it would be fairly natural to define a schema for the output_config.json

Is that implying that there is an alternative cmdstan output where the config info (and possibly other non-draw output) is stored in json format? I recall looking for something like this recently, but didn't see anything. 

Either way defining some kind of schema for json (or some other format) output would be great. Even if it isn't used as a standard output for cmdstan for a while, with a defined schema we could develop against that now and clean up some key bits of the library. And, presuming the Stan CSV format doesn't really change radically, we could just implement a converter that takes a given Stan CSV to structured json and handle backwards compatibility for as long as those output formats are in common use.

WardBrian:
> Is that implying that there is an alternative cmdstan output where the config info (and possibly other non-draw output) is stored in json format?

Yes, `output save_cmdstan_config=true` was added in CmdStan 2.34, which is a JSON version of the pre-header comment in the output CSV. Other non-sample items like the metric are available as JSON via other arguments, and [the timing hopefully will be available as JSON soon](https://github.com/stan-dev/stan/issues/3340)

Because it is its own file, you do end up with the unfortunate "what happens if I mix a file from run A with a file from run B" problem, but I think that's just life. We could consider introducing the requirement that a output directory be blank before a run, which would avoid almost all 'natural' ways that problem could arise. 

> And, presuming the Stan CSV format doesn't really change radically, we could just implement a converter that takes a given Stan CSV to structured json and handle backwards compatibility for as long as those output formats are in common use.

Believe it or not, I tried this same thing nearly 4 years ago: https://github.com/WardBrian/experimental-cmdstan-parsing

There were some irregularities/inconsitencies with the headers that made converting them into a JSON not super natural. The built-in support in cmdstan is much better. But I still wish it had worked out, because then cmdstanpy would have been free of this nonsense for years by now!

That little experiment was using pydantic, which would probably be a natural choice if we were using the new file as well.

cmdstanpy/stanfit/laplace.py
```diff
@@ -89,16 +86,10 @@ def _assemble_draws(self) -> None:
...
-                fd,
-                dtype=float,
-                ndmin=2,
-                delimiter=',',
-                comments="#",
-            )
+        *_, draws = stancsv.parse_comments_header_and_draws(
+            self._runset.csv_files[0]
+        )
+        self._draws = stancsv.csv_bytes_list_to_numpy(draws)
```
WardBrian, line 92: Not immediately sure if these functions would ever raise for GQs but if so we should wrap in a `try`/re-raise like we do in MCMC, to at least identify which chain was the problem.

We might want to do this in the other algorithms, though because they're single files it probably doesn't add as much compared to just letting the underlying error propagate -- thoughts?

cmdstanpy/utils/stancsv.py
```diff
@@ -129,245 +134,384 @@ def parse_hmc_adaptation_lines(
...
+                out[key] = 0
+            else:
+                for cast in (int, float):
+                    try:
+                        out[key] = cast(val)
+                        break
+                    except ValueError:
+                        pass
+                else:
+                    out[key] = val
```
WardBrian, line 187: Could this just be `for cast in (bool, int, float)` instead of special casing `true` and `false` like this?

WardBrian: Oh, probably not because they're not capitalized?

cmdstanpy/utils/stancsv.py
```diff
@@ -129,245 +134,384 @@ def parse_hmc_adaptation_lines(
...
+
+    filtered = filter_csv_bytes_by_columns(sampling_draws, indexes_to_keep)
+    arr = csv_bytes_list_to_numpy(filtered).astype(int)
+
+    num_max_treedepth = np.sum(arr[:, 0] == max_treedepth)
+    num_divergences = np.sum(arr[:, 1])
+    return num_max_treedepth, num_divergences
+
+
+def is_sneaky_fixed_param(header: str) -> bool:
```
WardBrian, line 255: Fun name. Can you add a `# TODO: Remove after CmdStan 2.37 is the minimum version`?

cmdstanpy/utils/stancsv.py
```diff
@@ -8,60 +8,69 @@
...
 
 import numpy as np
 import numpy.typing as npt
-import pandas as pd
 
 from cmdstanpy import _CMDSTAN_SAMPLING, _CMDSTAN_THIN, _CMDSTAN_WARMUP
 
 
-def parse_stan_csv_comments_and_draws(
+def parse_comments_header_and_draws(
```
WardBrian, line 19: I don't see an immediate way to resolve this (I think it's too subtle a pattern to encode as a generator) but I just wanted to note that I _think_ all of the usages of this function that don't need the draws (e.g. the calls in the InferenceMetadata factory) also don't need any of the comments after the header, so they might be doing an unnecessary pass over the rest of the CSV.

Short of splitting into two functions I don't see a nice pattern to avoid this, but it's probably not too expensive to be worth worrying about anyway.

cmdstanpy/utils/stancsv.py
```diff
@@ -129,245 +134,384 @@ def parse_hmc_adaptation_lines(
...
+                adaptation_block_idx - header_line_idx - 1  # type: ignore
+            )
+        num_sampling = timing_block_idx - sampling_block_idx
+        return num_warmup, num_sampling
+
+    if isinstance(stan_csv, (str, os.PathLike)):
+        with open(stan_csv, "rb") as f:
+            return determine_draw_counts(f)
+    else:
+        return determine_draw_counts(stan_csv)
```
WardBrian, line 324: almost worth extracting this pattern out into a `call_on_file(func, stan_csv)`

cmdstanpy/stanfit/gq.py
```diff
@@ -633,11 +626,10 @@ def _assemble_generated_quantities(self) -> None:
             order='F',
         )
         for chain in range(self.chains):
-            with open(self.runset.csv_files[chain], 'r') as fd:
-                lines = (line for line in fd if not line.startswith('#'))
-                gq_sample[:, chain, :] = np.loadtxt(
-                    lines, dtype=np.ndarray, ndmin=2, skiprows=1, delimiter=','
-                )
+            *_, draws = stancsv.parse_comments_header_and_draws(
```
WardBrian: Not immediately sure if these functions would ever raise for GQs but if so we should wrap in a `try`/re-raise like we do in MCMC, to at least identify which chain was the problem.

We might want to do this in the other algorithms, though because they're single files it probably doesn't add as much compared to just letting the underlying error propagate -- thoughts?

WardBrian (commented):
I really like how this is looking now! A few comments/questions:

cmdstanpy/utils/stancsv.py
amas0: If val were the strings `'true'` or `'false'`, then `bool('true') == bool('false') == True` since the truth value of a string is `True` if it is non-empty.

cmdstanpy/utils/stancsv.py
amas0: Yeah, I've though about this. 

I think the current structure of the csv files somewhat forces a tradeoff between passes over the files and clarity. In the current code, when going through the process of fitting a model with `sample` and loading the draws, each csv is read 3 times? Twice in the validation step after fitting and once when reading the draws. Additionally within that, once loaded into memory the comments are scanned through a few times. Draws, I think 3 times? Once to validate that all draw rows have the same number of columns, once to count divergences + max treedepth warnings, and once when actually loading into numpy.

All that is to say, we do have a lot of loading things into memory that we don't need, but in practice it doesn't seem to be a big deal. In all the various testing I did, scanning through the files is very efficient. In particular scanning through these files as lists of bytes is very lightweight compared to relatively expensive process of loading draws in as a numpy array.

I think it's probably more prudent to design against cleaner future output formats than try to really cleverly redesign this parsing.

cmdstanpy/utils/stancsv.py
amas0: Added!

cmdstanpy/stanfit/gq.py
amas0: I think this is a good idea -- even in the single csv file case, it would be useful to at least identify the file location that cause the issue. If nothing else, it would probably speed up debugging in some unknown instances where errors are thrown.

I've added `try...except`s around the various places we process the stan csvs to raise the problem file path from the original exception.

cmdstanpy/utils/stancsv.py
amas0: I personally fall on the side that repeating this short pattern twice isn't _quite_ enough to justify another function, but I don't hold that belief too strongly. Happy to change if you want.

amas0:
> Yes, output save_cmdstan_config=true was added in CmdStan 2.34, which is a JSON version of the pre-header comment in the output CSV.

Ahh, I definitely missed this. Thanks for pointing it out.

> Believe it or not, I tried this same thing nearly 4 years ago: https://github.com/WardBrian/experimental-cmdstan-parsing
> 
> There were some irregularities/inconsitencies with the headers that made converting them into a JSON not super natural. The built-in support in cmdstan is much better. But I still wish it had worked out, because then cmdstanpy would have been free of this nonsense for years by now!
> 
> That little experiment was using pydantic, which would probably be a natural choice if we were using the new file as well.

I'm going to add checking some of this out to my to-do list. Since things are more stable on the cmdstan side, it might be worth a re-visit.

WardBrian (approved):
Looks good to go. Thanks again!

PR #808, opened 2025-08-22
Remove Python 3.8 support

### LINKED ISSUE #802: Removal of support for end-of-life Python versions
Currently, cmdstanpy supports a minimum Python version of 3.8. Python 3.8 went [end-of-life in October 2024](https://devguide.python.org/versions/). While it's possible to continue supporting these older versions, various dependencies like pandas, pytest, polars, etc. have all removed support for 3.8, so we need to use older package versions to maintain compatibility.

I'd like to propose adopting a similar practice where new releases drop support for Python versions as they go end-of-life. So, 3.8 support could be dropped now and 3.9 support would be dropped in October 2025. 

This would also have the benefit of allowing the library to use newer features, which would clean up certain areas (some of the typing features especially have gotten cleaner since 3.8).

amas0:
#### Submission Checklist

- [x] Run unit tests

#### Summary

This PR bumps the required Python version of the library to 3.9. Resolves #802 .

Also uses the 3.9 features of typing with generics for `dict, list, tuple` and dictionary unions with `|`. 

I also removed 3.8 testing from the Github workflow for CI/CD and added 3.13 (I think the way I did it is correct? I've not used it before).

#### Copyright and Licensing

### DIFF
--- a/cmdstanpy/cmdstan_args.py
+++ b/cmdstanpy/cmdstan_args.py
@@ -105,7 +105,9 @@ class SamplerArgs:
         * length of per-chain lists equals specified # of chains
         """
         if not isinstance(chains, (int, np.integer)) or chains < 1:
-            raise ValueError('Sampler expects number of chains to be greater than 0.')
+            raise ValueError(
+                'Sampler expects number of chains to be greater than 0.'
+            )
         if not (
             self.adapt_delta is None
             and self.adapt_init_phase is None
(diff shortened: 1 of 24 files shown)


cmdstanpy/cmdstan_args.py
WardBrian, line 1: It looks like this file got processed with string normalization turned on before you updated the pyproject.toml -- mind reverting those changes?

amas0: Woops, good catch. 

As a side note, we're pretty inconsistent throughout the library with string quoting. It'd be a big edit, but is there any desire to just go ahead and normalize everything to either single or double quotes and be consistent going forward?

WardBrian (approved):
Thank you!

WardBrian:
> As a side note, we're pretty inconsistent throughout the library with string quoting. It'd be a big edit, but is there any desire to just go ahead and normalize everything to either single or double quotes and be consistent going forward?

I wouldn't be opposed to doing this in it's own PR, but the setting disabling string normalization predates me so we may also want to ask @mitzimorris if she minds

PR #820, opened 2025-10-21
Update model to always be compiled, remove deprecated compilation features

### LINKED ISSUE #695: Breaking change: Disallow CmdStanModel without executable
#### Summary:
As discussed in #689.

The current behavior when `exe_file=None, compile=False` yields a broken - essentially all methods other than `compile()` cannot be called until after `compile()` is called.

I think it should be factored such that:

`cmdstanpy.compile(stan_file, force=False, stanc_args={}, cmdstan_args={}) -> str`
is a function outside of any class which invokes `make` as requested.

And the constructor looks like
`CmdStanMLE(stan_file=None, exe_file=None, force_compile=False, stanc_args={}, cmdstan_args={})`

This raises an error if:
- stan_file and exe_file are both None, or
- force_compile is True but stanc_file was None

Otherwise, it will always be the case that when the constructor is finished, we have an executable ready for use (either because the user supplied one, or because we compiled it)

This would be a 2.0 change

### LINKED ISSUE #773: `force_compile` in `CmdStanModel` Does Not Work
#### Summary:
Title describes the problem :) 

#### Description:
I believe the problem comes from this [line](https://github.com/stan-dev/cmdstanpy/blob/0ed8811c7b0ce67cf62496a2b0e426399492b914/cmdstanpy/model.py#L256). Shouldn't it read `if compile or self._exe_file is None` instead of the current `if compile and self._exe_file is None`? As written, `force_compile` will only actually force compilation if no exe file is provided, which kind of defeats the point IMO. Changing the "and" to an "or" results in a compilation if `force_compile = True` OR if `force_compile = False` and no executable was provided.

#### Current Version:
The referenced code issue is in the main branch. Probably not helpful, but here's the info for what I'm working with:

INSTALLED VERSIONS
---------------------
python: 3.12.5 | packaged by conda-forge | (main, Aug  8 2024, 18:36:51) [GCC 12.4.0]
python-bits: 64
OS: Linux
OS-release: 6.5.0-35-generic
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: None
LANG: en_US.UTF-8
LOCALE: ('en_US', 'UTF-8')
cmdstan_folder: /home/bwittmann/micromamba/envs/dms_stan/bin/cmdstan
cmdstan: (2, 33)
cmdstanpy: 1.2.4
pandas: 2.2.2
xarray: 2023.6.0
tqdm: 4.66.4
numpy: 1.26.4

' '

WardBrian:
#### Submission Checklist

- [x] Run unit tests

#### Summary

Closes #695. As described there, the exefile always exists after the constructor concludes.

#### Copyright and Licensing

Simons Foundation

### DIFF
--- a/cmdstanpy/model.py
+++ b/cmdstanpy/model.py
@@ -13,16 +13,7 @@ from collections import OrderedDict
 from concurrent.futures import ThreadPoolExecutor
 from io import StringIO
 from multiprocessing import cpu_count
-from typing import (
-    Any,
-    Callable,
-    Iterable,
-    Literal,
-    Mapping,
-    Optional,
-    TypeVar,
-    Union,
-)
+from typing import Any, Callable, Iterable, Mapping, Optional, TypeVar, Union
 
 import numpy as np
 import pandas as pd
@@ -114,15 +105,12 @@ class CmdStanModel:
 
     def __init__(
         self,
-        model_name: Optional[str] = None,
         stan_file: OptionalPath = None,
         exe_file: OptionalPath = None,
         force_compile: bool = False,
         stanc_options: Optional[dict[str, Any]] = None,
         cpp_options: Optional[dict[str, Any]] = None,
         user_header: OptionalPath = None,
-        *,
-        compile: Union[bool, Literal['force'], None] = None,
     ) -> None:
         """
         Initialize object given constructor args.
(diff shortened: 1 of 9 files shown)


cmdstanpy/model.py
```diff
@@ -194,41 +160,29 @@ def __init__(
...
-            else:
-                if self._name != os.path.splitext(exename)[0]:
-                    raise ValueError(
-                        'Name mismatch between Stan file and compiled'
-                        ' executable, expecting basename: {}'
-                        ' found: {}.'.format(self._name, exename)
-                    )
+                model_info = self.src_info()
+                if 'parameters' in model_info:
+                    self._fixed_param |= len(model_info['parameters']) == 0
```
amas0, line 172: Are we planning on removing support for the older cmdstan versions in 2.0 (see #685)? If so, this check should probably be modified to just `cmdstan_version_before(2, 36)`?

cmdstanpy/model.py
```diff
@@ -276,33 +226,25 @@ def stan_file(self) -> OptionalPath:
         return self._stan_file
 
     @property
-    def exe_file(self) -> OptionalPath:
+    def exe_file(self) -> Union[os.PathLike, str]:
```
amas0: I believe the only places that `self._exe_file` is set now are lines 135 and 171. Since both `os.path.realpath` and `compilation.compile_stan_file` return `str`, this can probably be set to simply `def exe_file(self) -> str`.

amas0 (approved):
Looks good! Minor thoughts, but it's great to see some of this deprecated code finally getting removed.

cmdstanpy/model.py
WardBrian: I think we should! 

Though I would wait to make your suggested change here until a separate PR where we actually do so more broadly

And the other question would be what would the new minimum be, and can we just delete this code entirely

cmdstanpy/model.py
WardBrian: Good catch, you are correct

cmdstanpy/model.py
amas0: Makes sense to me.

PR #825, opened 2025-10-29
2.0 more deprecation removals

WardBrian:
#### Submission Checklist

- [x] Run unit tests

#### Summary
A few more deprecation removals:

- model.format
- vb.stan_variable now always returns an np array, and returns the variational sample by default
- mle.stan_variable now always returns an np array

#### Copyright and Licensing

Simons Foundation

### DIFF
--- a/cmdstanpy/stanfit/mle.py
+++ b/cmdstanpy/stanfit/mle.py
@@ -77,9 +77,7 @@ class CmdStanMLE:
         """
         # pylint: disable=unused-argument
 
-        return {
-            name: np.array(val) for name, val in self.stan_variables().items()
-        }
+        return self.stan_variables()
 
     def __repr__(self) -> str:
         repr = 'CmdStanMLE: model={}{}'.format(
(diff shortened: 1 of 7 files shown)


cmdstanpy/stanfit/mle.py
```diff
@@ -273,7 +265,7 @@ def stan_variable(
 
     def stan_variables(
         self, inc_iterations: bool = False
-    ) -> dict[str, Union[np.ndarray, float]]:
+    ) -> dict[str, np.ndarray]:
         """
```
amas0, line 267: This typing chain means we can rewrite `CmdStanMLE.create_inits(...)` to be:

```python
        return {name: val for name, val in self.stan_variables().items()}
```

where we previously had a `np.array(val)` to account for the possible `float` output.

amas0 (approved):
Looks good to me -- one minor change we can make now that the typing of `stan_variables` has been simplified.

cmdstanpy/stanfit/mle.py
WardBrian: Good catch, we can go one step further and just `return self.stan_variables()` then

amas0: Perfect

PR #829, opened 2025-11-03
Drop support for Python 3.9

amas0:
#### Submission Checklist

- [x] Run unit tests

#### Summary

Drops support for Python 3.9, which transitioned to end-of-life 1 Nov 2025.

@WardBrian, I have a style question for this.

3.10 introduces `|` syntax for `Union` so we could make the following changes in the library `Union[A, B] -> A | B` and `Optional[A] -> A | None`. 

I prefer the newer style, I find it cleaner and you need fewer `typing` imports. If you're amenable to it, I'd like to go through the library and update our use of `Union` and `Optional` to the new syntax. Let me know what you think.

#### Copyright and Licensing


WardBrian:
> I prefer the newer style, I find it cleaner and you need fewer `typing` imports. If you're amenable to it, I'd like to go through the library and update our use of `Union` and `Optional` to the new syntax.

I'm also in favor.

I think 3.10 as a minimum means we could also use match statements now.

My only concern is that I think 3.9 was the default version for a _lot_ of OSes etc once they finally moved off 2.7/3.6. Hopefully people are using environments, but who knows. On the other hand, we did just have a release that still did support 3.9, so probably by the time we release 2.0 it will all shake out alright

ahartikainen:
I would recommend dropping the support. The current version will work long enough if there are 3.9 users still left.

amas0:
Yeah, I think my position is "if your system is using formally end-of-life software, than our responsibility does not go beyond pointing out that there are older versions of this package that _did_ support those older versions".

I'll go ahead and propagate the new syntax throughout before marking this ready.

WardBrian:
Looks like this needs a rebase

3.10+ also lets us solve https://github.com/stan-dev/cmdstanpy/pull/788#discussion_r2073538761 more nicely

WardBrian (approved):
Thanks!

Issue #714, opened 2023-10-30
[Stan 2.34] Utilize argument save_cmdstan_config


WardBrian:
#### Summary:
Added in https://github.com/stan-dev/cmdstan/pull/1204, `save_cmdstan_config` outputs the arguments provided to cmdstan as a JSON file. This should remove most of the need to parse the CSV comments, besides perhaps for optional validation

amas0:
Do we want to simply enable this across the board or provide an ability to disable it?

Since this is mostly for internal use, seems like we could just make a one-line change to enable it universally in `CmdStanArgs`.

WardBrian:
I’d be in favor of doing it by default

WardBrian:
https://github.com/stan-dev/cmdstanpy/pull/851

Issue #713, opened 2023-10-28
[Stan 2.34] expose sampler argument "save_metric"


mitzimorris:
#### Summary:

Add functionality to save metric as JSON to CmdStanPy.

#### Description:

CmdStan PR https://github.com/stan-dev/cmdstan/pull/1203 added new adaptive sampler argument `save_metric`.   Make this available from CmdStanPy:

* Expose adaptive sampler argument `save_metric`
* Add methods to `CmdStanMCMC` to access the saved metric and stepsize.
* Document use and use cases.

#### Current Version:

This will be available in 3.34

WardBrian:
Should we expose it, or just always enable it by default?

The files are never that large relative to the output csvs

mitzimorris:
always enable by default.   what are we doing with `save_config`?   we should be consistent.

WardBrian:
I think `save_cmdstan_config` will always be enabled if it is supported. Eventually I think it makes sense to move essentially all the config logic to reading that file and then ignore the comments in the output csvs entirely.

amas0:
We still want to add this, correct? It would be fairly straightforward to do so in line with the other output files we're tracking.

With this enabled, we would then replace the current CSV parsing for stepsize and metric with the content of these files?

WardBrian:
> With this enabled, we would then replace the current CSV parsing for stepsize and metric with the content of these files?

Yes, exactly. We’d ideally like to get to the point where the only thing we read from the csv file is the draws, ignoring all the other comments

amas0:
I have been taking a look at this today and have a question. 

When we do something like perform a `CmdStanModel.generate_quantities(..., previous_fit=[files])`, where `files` is a list of stan csv file outputs from a previous MCMC run, we create a new `CmdStanMCMC` object from these files. As we are transitioning to _not_ parsing all of the information from the stan csv files, we can't reliably guarantee that the corresponding `metric.json` files are available as well when making the underlying `from_csv` call. 

How should we handle the metric information in that case? Should we try to lazily load it so you can create the `CmdStanMCMC` object from the csv, but if you try to access the metric info it will throw an exception? Should we just make those properties fault tolerant and perhaps just throw a warning if they can't be found? Or just have them be `None` when we don't have access to the JSON files?

WardBrian:
I think lazily loading it would make sense, it also might make sense to transition to thinking of the a previous fit as a folder rather than just a list of csv files, since we will start caring about more than just csv files.

[...]

Issue #805, opened 2025-08-15
Possible invalid/incorrect Stan csv used in unit testing


amas0:
While in the process of working to update some of the Stan CSV I/O functions, I ran into `test/data/fixed_param_sample.csv`, which seems to only be used in `test/test_sample.py::test_from_csv_fixed_param`. The naming here seems to imply that this CSV should have come from fixed param sampling?

But within the config block of this CSV file, we have:

```
#     algorithm = hmc (Default)
```

So this doesn't seem to actually be a CSV generated via fixed_param sampling? I'm not sure if I'm missing anything here. If so, let me know. If this is a mistake, I can go ahead and update it.

WardBrian:
I believe that file was representative of one that generated by a model with 0 parameters. Cmdstan would “helpfully” detect this and switch to using the fixed_param sampler, but the CSV header would still report whatever the requested settings were. 

The only way you could notice this was that there were fewer diagnostic parameters (only three __ columns, rather than 7, no adaptation comment, and the model would print something to stdout about it as well. 

This misfeature was finally removed a few versions ago, but I guess we’ve kept that test file around. We should probably have a more modern equivalent, if not as a replacement at least as an additional test

amas0:
Ah okay, interesting. Is there a policy on how old cmdstan versions are supported by cmdstanpy? If it's been a number of versions since generating a file like this has been possible, it seems reasonable to drop it in favor of a modern alternative.

I imagine for this test an appropriate csv would be one actually generated with the fixed_param sampler?

WardBrian:
We do not have an official policy (yet, c.f. #685). You can still find some code that claims to be in support of versions from july 2020:
https://github.com/stan-dev/cmdstanpy/blob/71d97ca4e153398f580cf8c8c89cc9739ca2eb7c/cmdstanpy/stanfit/mcmc.py#L537

I'm in favor of capping this and instituting some kind of "current + 3" scheme in the 2.0 release.

Unfortunately the change for fixed_param no longer being necessary was only one version ago, so we'd probably support it for some time anyway

amas0:
Ah okay that makes sense -- I'll go ahead and make sure this file structure is properly supported. 

Closing since this is intended behavior.

PR #834, opened 2025-11-06
Remove remaining deprecations, very old cmdstan version checks

### LINKED ISSUE #685: Remove support for CmdStan pre-2.34 in a future version

### LINKED ISSUE #766: `iter_warmpup=0` and `fixed_param=True` will always fail for stan 2.34.1
#### Summary:

If `fixed_param=True`, `adapt_engaged` will always be left as default (True) because of this early return https://github.com/stan-dev/cmdstanpy/blob/develop/cmdstanpy/cmdstan_args.py#L361.

This is okay for the newer stan since this [if-statement](https://github.com/stan-dev/cmdstan/pull/1277/files#diff-de5c2eb1977dd486b035c6a7ea3539d0b5c7bdac2cff55705b1344734a3a4f50R571). But for previous stan (tried 2.34.1), this will always fail since `adapt_engaged` must be False even when `fixed_param` is supplied.

#### Description:

This only affects previous versions of Stan. I am not sure about the policy of supporting versions but it is a breaking change that will make previously working code fail if cmdstan itself is not updated.

WardBrian:
#### Submission Checklist

- [x] Run unit tests

#### Summary

Removes the remaining deprecated code paths and closes #685. I took the somewhat-conservative latest-plus-two-versions path here, so I left checks for 2.35/2.36, but deleted any older than that. 

This also clears the path for us to tackle #713 and #714

#### Copyright and Licensing


amas0 (approved):
This looks good. 

I think this means we can also remove the test and testing csv mentioned in #805 ; which, if I recall, could only be produced by a cmdstan version we are no longer supporting.

WardBrian:
The change to make #805 go away was made in 2.35 I believe, so it's just within the current-minus-two range.

PR #835, opened 2025-11-06
Standardize output filenames

### LINKED ISSUE #833: Mismatch between `-stdout.txt` output files and also some other output naming conventions
As part of running a Stan model, we generate a range of output files; by default `-stdout.txt` and `.csv` files that contain the process output and inference output, respectively. Optionally, we can output `diagnostic` and `profile` files to track latent dynamics and profiling information, if present. 

The current logic will produce a set of output files (under the default 4 chains, one process per chain) like this:

```
-rw-r--r-- 1 amas amas   2367 Nov  5 15:28 bernoulli-20251105152812_0-stdout.txt
-rw-r--r-- 1 amas amas  88052 Nov  5 15:28 bernoulli-20251105152812_1.csv
-rw-r--r-- 1 amas amas   2357 Nov  5 15:28 bernoulli-20251105152812_1-stdout.txt
-rw-r--r-- 1 amas amas  88050 Nov  5 15:28 bernoulli-20251105152812_2.csv
-rw-r--r-- 1 amas amas   2358 Nov  5 15:28 bernoulli-20251105152812_2-stdout.txt
-rw-r--r-- 1 amas amas  88090 Nov  5 15:28 bernoulli-20251105152812_3.csv
-rw-r--r-- 1 amas amas   2358 Nov  5 15:28 bernoulli-20251105152812_3-stdout.txt
-rw-r--r-- 1 amas amas  86919 Nov  5 15:28 bernoulli-20251105152812_4.csv
-rw-r--r-- 1 amas amas 150457 Nov  5 15:28 bernoulli-20251105152812-diagnostic_1.csv
-rw-r--r-- 1 amas amas 150444 Nov  5 15:28 bernoulli-20251105152812-diagnostic_2.csv
-rw-r--r-- 1 amas amas 150808 Nov  5 15:28 bernoulli-20251105152812-diagnostic_3.csv
-rw-r--r-- 1 amas amas 149214 Nov  5 15:28 bernoulli-20251105152812-diagnostic_4.csv
-rw-r--r-- 1 amas amas    190 Nov  5 15:28 bernoulli-20251105152812-profile_1.csv
-rw-r--r-- 1 amas amas    189 Nov  5 15:28 bernoulli-20251105152812-profile_2.csv
-rw-r--r-- 1 amas amas    190 Nov  5 15:28 bernoulli-20251105152812-profile_3.csv
-rw-r--r-- 1 amas amas    190 Nov  5 15:28 bernoulli-20251105152812-profile_4.csv
```

For every file except the `stdout` files, we use the `chain_id` to index the file itself, so we get this mismatch where the `stdout` files are indexed 0-3 and the rest are indexed 1-4. This can result in some confusion where someone would thinking that the file suffixed by `_1-stdout.txt` corresponds to the output `_1.csv`, when it's actually `0-stdout.txt`.

I think we should change this naming to align, will be an easy fix.

Second point I want to raise is that we have this inconsistency in how we name these files with indexes and extra bits. For `stdout`, we do `{idx}-stdout.txt`, so the index comes first, but for profile and diagnostic, we do `(diagnostic|profile)_{idx}.csv`, so the index comes second. I think we should probably align these so that our output files have consistent naming.

Welcome any thoughts from others -- if we have a consensus, I'll go ahead and implement.

amas0:
#### Submission Checklist

- [x] Run unit tests

#### Summary

This change proposes a new standard naming convention for output files to address some inconsistencies identified in #833.

Specifically it fixes the mismatched stdout mismatched index and enforces all output files to stick to a naming convention that starts with `{model_name}-{datetime}_{id}`. Extra names, e.g. `stdout`, `profile`, will be append with an `_{extra}` before the file suffix.

This structure was chosen as it is compatible with the intended new argument `save_cmdstan_config` referenced in #714. 

Under these proposed changes, a default 4-chain sample with latent dynamics and profiling saved will create the following output files:

```
bernoulli-20251106132150_1.csv
bernoulli-20251106132150_1_diagnostic.csv
bernoulli-20251106132150_1_profile.csv
bernoulli-20251106132150_1_stdout.txt
bernoulli-20251106132150_2.csv
bernoulli-20251106132150_2_diagnostic.csv
bernoulli-20251106132150_2_profile.csv
bernoulli-20251106132150_2_stdout.txt
bernoulli-20251106132150_3.csv
bernoulli-20251106132150_3_diagnostic.csv
bernoulli-20251106132150_3_profile.csv
bernoulli-20251106132150_3_stdout.txt
bernoulli-20251106132150_4.csv
bernoulli-20251106132150_4_diagnostic.csv
bernoulli-20251106132150_4_profile.csv
bernoulli-20251106132150_4_stdout.txt
```

Not included in this PR, but anticipated is the inclusion of the cmdstan config saved, which produces the following output if included:

```
bernoulli-20251106132150_1_config.json
bernoulli-20251106132150_1.csv
bernoulli-20251106132150_1_diagnostic.csv
bernoulli-20251106132150_1_profile.csv
bernoulli-20251106132150_1_stdout.txt
bernoulli-20251106132150_2_config.json
bernoulli-20251106132150_2.csv
bernoulli-20251106132150_2_diagnostic.csv
bernoulli-20251106132150_2_profile.csv
bernoulli-20251106132150_2_stdout.txt
bernoulli-20251106132150_3_config.json
bernoulli-20251106132150_3.csv
bernoulli-20251106132150_3_diagnostic.csv
bernoulli-20251106132150_3_profile.csv
bernoulli-20251106132150_3_stdout.txt
bernoulli-20251106132150_4_config.json
bernoulli-20251106132150_4.csv
bernoulli-20251106132150_4_diagnostic.csv
bernoulli-20251106132150_4_profile.csv
bernoulli-20251106132150_4_stdout.txt
```

#### Copyright and Licensing

### DIFF
--- a/cmdstanpy/stanfit/runset.py
+++ b/cmdstanpy/stanfit/runset.py
@@ -222,7 +222,7 @@ class RunSet:
             case (id, ""):
                 file = f"{self._base_outfile}_{id}{suffix}"
             case (id, extra) if extra != "":
-                file = f"{self._base_outfile}_{id}_{extra}{suffix}"
+                file = f"{self._base_outfile}_{extra}_{id}{suffix}"
             case _:
                 raise ValueError("Cannot construct valid file name")
         return os.path.join(self._outdir, file)
--- a/cmdstanpy/stanfit/runset.py
+++ b/cmdstanpy/stanfit/runset.py
@@ -214,17 +214,12 @@ class RunSet:
         self, suffix: str, *, extra: str = "", id: int | None = None
     ) -> str:
         """Generate a standard file name according to CmdStan output pattern"""
-        match (id, extra):
-            case (None, ""):
-                file = f"{self._base_outfile}{suffix}"
-            case (None, extra) if extra != "":
-                file = f"{self._base_outfile}_{extra}{suffix}"
-            case (id, ""):
-                file = f"{self._base_outfile}_{id}{suffix}"
-            case (id, extra) if extra != "":
-                file = f"{self._base_outfile}_{extra}_{id}{suffix}"
-            case _:
-                raise ValueError("Cannot construct valid file name")
+        file = self._base_outfile
+        if extra:
+            file += f"_{extra}"
+        if id is not None:
+            file += f"_{id}"
+        file += suffix
         return os.path.join(self._outdir, file)
 
     def _retcode(self, idx: int) -> int:
--- a/cmdstanpy/stanfit/runset.py
+++ b/cmdstanpy/stanfit/runset.py
@@ -198,10 +198,7 @@ class RunSet:
 
     def _check_retcodes(self) -> bool:
         """Returns ``True`` when all chains have retcode 0."""
-        for code in self._retcodes:
-            if code != 0:
-                return False
-        return True
+        return all(retcode == 0 for retcode in self._retcodes)
 
     @property
     def diagnostic_files(self) -> list[str]:
--- a/cmdstanpy/stanfit/runset.py
+++ b/cmdstanpy/stanfit/runset.py
@@ -38,10 +38,7 @@ class RunSet:
         self._args = args
         self._chains = chains
         self._one_process_per_chain = one_process_per_chain
-        if one_process_per_chain:
-            self._num_procs = chains
-        else:
-            self._num_procs = 1
+        self._num_procs = chains if one_process_per_chain else 1
         self._retcodes = [-1 for _ in range(self._num_procs)]
         self._timeout_flags = [False for _ in range(self._num_procs)]
         if chain_ids is None:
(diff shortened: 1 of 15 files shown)


amas0:
@WardBrian Do you have any insight into this test failure? I don't have a windows machine on hand to look more closely.

WardBrian:
I think it might just be sporadic, lets see

Related to the PR, I think this might be wrong when using STAN_THREADs to run multiple chains in one executable. In those cases, the `_ID` will always be forced to be at the end. For example, if you run

```
$ ./bernoulli sample num_chains=4 data file=bernoulli.data.json output diagnostic_file=foo.csv
$ ls
bernoulli            bernoulli.data.R  foo_1.csv  foo_3.csv  output_1.csv  output_3.csv
bernoulli.data.json  bernoulli.stan    foo_2.csv  foo_4.csv  output_2.csv  output_4.csv
```

Runset has to deal with this complexity of abstracting over whether we picked the filenames or cmdstan did, at least until the minimum version is 2.37 and we can use a comma separated list and do whatever we want. 

I suspect we may be lacking sufficient tests that set `force_one_process_per_chain=False` to actually expose these issues

Note that when it is time to set save_cmdstan_config, only one of those files will be generated regardless of `num_chains`, so that will add yet another special case.

amas0:
Hmmmm, okay I see the distinctions in a quick test. Enabling `STAN_THREADS` and sampling under this current code will generate (with save_cmdstan_config enabled):

```python
bernoulli-20251106145201_1.csv
bernoulli-20251106145201_2.csv
bernoulli-20251106145201_3.csv
bernoulli-20251106145201_4.csv
bernoulli-20251106145201_config.json
bernoulli-20251106145201_diagnostic_1.csv
bernoulli-20251106145201_diagnostic_2.csv
bernoulli-20251106145201_diagnostic_3.csv
bernoulli-20251106145201_diagnostic_4.csv
bernoulli-20251106145201_profile.csv
bernoulli-20251106145201_stdout.txt
```

So we get on per-chain files (just diagnostic) an appended `_{id}`.

This is my understanding: 

In the case where we run one-process per file, we can get consistency across the output files, because we set the the ID as part of the base file names we give to cmdstan, but if we sample in a single process, cmdstan will generate the per-chain files with an appended `_{id}`, which seems to only affect the diagnostic file (and the stancsv, but extra bits to the filename makes it the same) since everything else is per-process.

Does that sound right?

I'm not sure if we can enforce absolute consistency between the single vs multi process cases unless we renamed files after they were written by cmdstan? In the one process for all chains case, we could rename diagnostic files to have the id before `_diagnostic` to make the id always be in the beginning. If we attempted to move the id to the end, then we would have to rename the `config.json` file in the multiple process case since that seems to inherit naming from the stancsv name.

We could also just accept that when `force_one_process_per_chain` is true, then the ids occur first, and in the alternative, they occur after (which would only happen when `save_latent_dynamics` is enabled anyways).

WardBrian:
I think it’s more consistent if we always end with _id, and then when we only have one chain we just know we won’t have any _# for those other files 

I’m still hopeful one day we can make the multithreaded behavior the default and nuke all this…

amas0:
Okay cool -- I've swapped the order. This will consistent throughout all the output files we have right now, regardless of single/multi process. For clarity, this is the output now:

`one_process_per_chain = False`

```
bernoulli-20251106180710_1.csv
bernoulli-20251106180710_2.csv
bernoulli-20251106180710_3.csv
bernoulli-20251106180710_4.csv
bernoulli-20251106180710_diagnostic_1.csv
bernoulli-20251106180710_diagnostic_2.csv
bernoulli-20251106180710_diagnostic_3.csv
bernoulli-20251106180710_diagnostic_4.csv
bernoulli-20251106180710_profile.csv
bernoulli-20251106180710_stdout.txt
```

and `one_process_per_chain = True`:

```
bernoulli-20251106180852_1.csv
bernoulli-20251106180852_2.csv
bernoulli-20251106180852_3.csv
bernoulli-20251106180852_4.csv
bernoulli-20251106180852_diagnostic_1.csv
bernoulli-20251106180852_diagnostic_2.csv
bernoulli-20251106180852_diagnostic_3.csv
bernoulli-20251106180852_diagnostic_4.csv
bernoulli-20251106180852_profile_1.csv
bernoulli-20251106180852_profile_2.csv
bernoulli-20251106180852_profile_3.csv
bernoulli-20251106180852_profile_4.csv
bernoulli-20251106180852_stdout_1.txt
bernoulli-20251106180852_stdout_2.txt
bernoulli-20251106180852_stdout_3.txt
bernoulli-20251106180852_stdout_4.txt
```

I've got to add some tests for these cases, update some docstrings, and probably update some documentation somewhere. Will let you know when this is good to go.

What are the major barriers to making the multithreaded behavior the default? Would certainly clean some of this up..

cmdstanpy/stanfit/runset.py
```diff
@@ -216,16 +210,22 @@ def profile_files(self) -> list[str]:
...
-            suffix = f"_{id}{suffix}"
-        file = os.path.join(
-            self._output_dir, f"{self._base_outfile}{extra}{suffix}"
-        )
-        return file
+        """Generate a standard file name according to CmdStan output pattern"""
+        match (id, extra):
+            case (None, ""):
+                file = f"{self._base_outfile}{suffix}"
+            case (None, extra) if extra != "":
```
WardBrian: Are these `if extra != ""`s really necessary? I would've thought the case above would match if they were "" and then this would be redundant

As big a fan as I am of match statements in other languages, I'm not sure this is cleaner than
```
file = self._base_outfile
if extra:
    file += f"_{extra}"
if id is not None:
   file += f"_{id}"
file += suffix
```

amas0: You're right -- I'll simplify, thanks for the look

WardBrian:
> What are the major barriers to making the multithreaded behavior the default? Would certainly clean some of this up

See the discussion at https://github.com/stan-dev/cmdstan/issues/1179. tl;dr is that our current Windows toolchains suffer from pretty dramatic slowdowns when threading is enabled at compile time. This isn't an inherent Windows issue (e.g., using WSL or clang-cl does not have this issue), so there's some hope that updated or swapped-out toolchains could allow us to do this

amas0:
@WardBrian  Going through updating docstrings and such. Quite a few places that say:

```
Output files written to the temporary directory contain an additional
8-character random string, e.g. 'bernoulli-201912081451-1-5nm6as7u.csv'.
```

I'm assuming this is quite out of date? Blame places these lines at ~5 years old and this isn't something that happens on my machine. Just wanted to double check that this isn't a platform-specific thing or an edge case before I remove it.

WardBrian:
Oh yeah, I don't think we've done that in a very long time

amas0:
I think this should be good to go @WardBrian

WardBrian (approved):
Looks good, thank you!

PR #840, opened 2025-11-14
Add multithreading and stanc_optimization arguments to CmdStanModel

### LINKED ISSUE #836: Update user guide section on compiling
Following #820, we need to update the user guide workflow page, which has a section "Compile the Stan model". 

This section references now-removed compilation logic and code.

### LINKED ISSUE #837: Should we expose common cpp/stanc options as keyword arguments?
There are a few features that, in order to use, we require passing flags in a dictionary to the model when compiling. This includes options like `STAN_THREADS` and `STAN_OPENCL`, but I'm sure there's more (I don't know the full set).

This places a burden on users to figure out that these options exist and then remember how to include them. I feel like it would be more natural to expose at least the common ones as a keyword argument in the CmdStanModel class, e.g.:

```py
import cmdstanpy

model = cmdstanpy.CmdStanModel(stan_file="bernoulli.stan", enable_stan_threads=True)
```

Could insert `{"STAN_THREADS": "True"}` into the `cpp_options` dict under the hood. This seems like it would be a nice quality of life feature and would be fairly straightforward to implement.

amas0:
#### Submission Checklist

- [x] Run unit tests

#### Summary

This adds two new boolean keyword arguments to `CmdStanModel`: `multithreading` and `stanc_optimizations`. These are included as a short-hand way to enable the common compilation options `STAN_THREADS` and `O1` optimizations in stanc.

Also updates workflow documentation on model compilation to mention these new options and reflect recent removal of deprecated `compile` options.

Resolves #837 and #836.

#### Copyright and Licensing

### DIFF
--- a/cmdstanpy/model.py
+++ b/cmdstanpy/model.py
@@ -119,10 +119,12 @@ class CmdStanModel:
         """
         self._name = ''
         self._stan_file = None
-        self._stanc_options = self._resolve_stanc_options(
+        self._stanc_options = compilation.resolve_stanc_options(
             stanc_options, stanc_optimizations
         )
-        cpp_options = self._resolve_cpp_options(cpp_options, multithreading)
+        cpp_options = compilation.resolve_cpp_options(
+            cpp_options, multithreading
+        )
 
         self._fixed_param = False
 
@@ -259,26 +261,6 @@ class CmdStanModel:
                 'Cannot read file Stan file: %s', self._stan_file
             )
         return code
-
-    @staticmethod
-    def _resolve_cpp_options(
-        cpp_options: dict[str, Any] | None, multithreading: bool
-    ) -> dict[str, Any]:
-        out = cpp_options or {}
-        out = out.copy()
-        if multithreading and "STAN_THREADS" not in out:
-            out["STAN_THREADS"] = "TRUE"
-        return out
-
-    @staticmethod
-    def _resolve_stanc_options(
-        stanc_options: dict[str, Any] | None, stanc_optimizations: bool
-    ) -> dict[str, Any]:
-        out = stanc_options or {}
-        out = out.copy()
-        if stanc_optimizations and "O" not in out:
-            out["O"] = 1
-        return out
 
     def optimize(
         self,
--- a/cmdstanpy/model.py
+++ b/cmdstanpy/model.py
@@ -104,8 +104,10 @@ class CmdStanModel:
         :param exe_file: Path to compiled executable file.
         :param force_compile: Whether or not to force recompilation if
             executable file already exists.
-        :param stanc_options: Options for stanc compiler.
-        :param cpp_options: Options for C++ compiler.
+        :param stanc_options: Options for stanc compiler. Note, this
+            will override the `stanc_optimizations` if in conflict.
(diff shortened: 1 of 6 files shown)


cmdstanpy/model.py
```diff
@@ -101,14 +104,25 @@ def __init__(
...
+        :param cpp_options: Options for C++ compiler. Note, this will
+            override the `multithreading` option if in conflict.
         :param user_header: A path to a header file to include during C++
             compilation.
+        :param multithreading: Enables multithreading in a Stan model.
+            Equivalent to `cpp_options = {"STAN_THREADS": "TRUE"}`.
+            Defaults to False.
+        :param stanc_optimizations: Enables O1 optimizations in the
+            stanc compiler. Equivalent to `stanc_options = {"O": 1}`.
+            Defaults to False.
```
WardBrian, line 112: I think we should feel free to put these above the existing flags in the documentation

cmdstanpy/model.py
```diff
@@ -246,6 +260,24 @@ def code(self) -> str | None:
...
         return code
 
+    @staticmethod
+    def _resolve_cpp_options(
+        cpp_options: dict[str, Any] | None, multithreading: bool
+    ) -> dict[str, Any]:
+        out = {}
+        if multithreading:
+            out["STAN_THREADS"] = "TRUE"
+        return out | cpp_options if cpp_options else out
```
WardBrian: it's minor, but I'd prefer writing this as 
```python

out = cpp_options or {}
if multithreading and `STAN_THREADS` not in out:
  out["STAN_THREADS"] = "TRUE"
return out
```

Same for stanc. My reasoning is the union operation and the inline if are hard to read in this context. We may also want to throw a copy.copy in there to prevent modifying the user's dictionary they pass in.

cmdstanpy/model.py
```diff
@@ -246,6 +260,24 @@ def code(self) -> str | None:
...
+
+    @staticmethod
+    def _resolve_stanc_options(
```
WardBrian: nit: both of these functions should live in compilation.py IMO

WardBrian (commented):
Thanks for tackling! A couple minor comments:

cmdstanpy/model.py
amas0: To clarify, just bumping them up in order for the docstring and not changing the actual argument order in the signature?

WardBrian: Correct. I agree these should be keyword-only, so they have to stay after the `*` in the signature, but we can promote them more in the docstring

WardBrian:
Should we add these to compilation.compile_stan_model do you think? That is ultimately what gets called by the constructor, but it is also available to users.

amas0:
I considered it when implementing this. 

It'd be an easy change, but I wasn't sure I liked the idea of adding the same keyword args to both `CmdStanModel` and `compile_stan_model` every time we wanted to add a shorthand flag. I also perceived `compile_stan_model` to be a bit lower-level.

I don't feel that strongly about it. Happy to make the change if you think it'd be useful.

WardBrian:
Seems fair — it’s easy enough to add them later if we desire

WardBrian (approved):

PR #844, opened 2025-12-02
Enable `save_metric=1` and sources MCMC metric info from new JSON file

### LINKED ISSUE #713: [Stan 2.34] expose sampler argument "save_metric"

amas0:
#### Submission Checklist

- [x] Run unit tests

#### Summary

This PR enables `save_metric=1` for cmdstan sampling with adaptation. This outputs a new metric JSON file that contains the step size, inv_metric, and metric type for the sampling run. This PR also now lazily sources this info from these files when the corresponding properties on the `CmdStanMCMC` objects are accessed. By changing the source of this info, we can now also remove all the code where we were parsing adaptation info from the Stan CSV file, which this PR does.

**Importantly, this PR proposes adding `pydantic` as a dependency to `cmdstanpy`.** With sourcing more info from JSON files (in this metric change and soon from the config files), we want to be more careful about I/O validation. Pydantic is now a fairly standard way to parse and validate data in the Python ecosystem and I think is a good fit for our needs.

Closes #713.

@WardBrian In the issue comments, we discussed an update to the `save_csvfiles` methods that would include the other output files. This PR doesn't include that change. I could work to incorporate it here, but I think it makes sense to look at that as a standalone PR.

#### Copyright and Licensing


WardBrian:
I'll need some time to look over this but at first glance it looks pretty good. I don't think Pydantic would make a ton of sense for this alone, but as you note we will probably want something more powerful when loading the config files.

I do worry about being _too_ strict about IO, since it could mean that cmdstanpy becomes unusable with development versions or right when a new version is released before cmdstanpy has itself updated. Right now this will most-likely 'just work' as long as you don't request information related to the part of the config that is new/different in that cmdstan version, but if we were doing full validation it would fail even if you only actually need some unrelated piece

amas0:
Yeah if it were just this, I wouldn't be interested in adding the dependency, but with potentially quite a few output files that need to be parsed, I think it would be useful. We could always implement the equivalent structures without pydantic, would just be a bit extra work.

Good point about strictness -- perhaps we could have validation issues throw warnings instead of errors? And only be strict where it would indicate something has gone wrong

cmdstanpy/stanfit/mcmc.py
```diff
@@ -92,6 +93,8 @@ def __init__(
         # info from CSV header and initial and final comment blocks
         config = self._validate_csv_files()
         self._metadata: InferenceMetadata = InferenceMetadata(config)
+        self._chain_metric_info: list[MetricInfo] = []
+        self._metric_info_parsed: bool = False
```
WardBrian: Do we really need this flag? It seems like it's always equivalent to `self._metric_type == None`

cmdstanpy/stanfit/metadata.py
```diff
@@ -1,10 +1,16 @@
 """Container for metadata parsed from the output of a CmdStan run"""
 
+from __future__ import annotations
```
WardBrian, line 3: Is this necessary for pydantic? I know there was a lot of discussion about annotation handling and pydantic, but I didn't keep up with it

cmdstanpy/stanfit/metadata.py
```diff
@@ -79,3 +85,53 @@ def stan_vars(self) -> dict[str, stanio.Variable]:
         These are the user-defined variables in the Stan program.
         """
         return self._stan_vars
+
+
+class MetricInfo(BaseModel):
+    """Structured representation of HMC-NUTS metric information,
+    as output by CmdStan"""
+
+    chain_id: int = Field(gt=0)
```
WardBrian: I believe chain id can be equal to 0, just not greater than

cmdstanpy/stanfit/metadata.py
```diff
@@ -79,3 +85,53 @@ def stan_vars(self) -> dict[str, stanio.Variable]:
...
+    stepsize: float
+    metric_type: Literal["diag_e", "dense_e", "unit_e"]
+    inv_metric: np.ndarray
+
+    model_config = {"arbitrary_types_allowed": True}
```
WardBrian: This sounds like a scary config option -- what does it do for us here?

cmdstanpy/stanfit/metadata.py
WardBrian: Also -- is there a reason we need the chain ID in here? Since it isn't in the file is necessitates our own from_json rather than just using `model_validate_json`

test/data/logistic_output_1_metric.json
WardBrian, line 1: All the test files here use diag_e, probably worth trying the other values as well?

WardBrian (commented):
First set of questions!

cmdstanpy/stanfit/mcmc.py
amas0: It's not strictly necessary. I felt it was a little cleaner that the check explicitly states that it's whether the information has been parsed rather than one of the attributes that is set is not none or false-y.

I can remove it in favor of checking `_metric_type` is set if you think that would be more consistent.

cmdstanpy/stanfit/metadata.py
amas0: It's not necessary for Pydantic. All it does is allow us to use the return type of the class name without quotes when we're doing the class methods or model validators (which are class methods). So in lines 114 and 131 the return type is just the class name `MetricInfo` instead of `'MetricInfo'`. I think this becomes the default behavior in like 3.14 or something.

cmdstanpy/stanfit/metadata.py
amas0: It's necessary to allow the Pydantic model to have types that it doesn't know how to validate by default. So by including the `ndarray`, we have to allow arbitrary types and then do the validation ourselves.

cmdstanpy/stanfit/metadata.py
amas0: Good catch! 

I think the chain ID is entirely dead code (I probably thought it would come up in implementation, which is why I added it in the first place). 

I'm going to go ahead and remove it.

test/data/logistic_output_1_metric.json
amas0: So, these particular files were created to ensure that existing tests did not break with the change of parsing metric info from these files. I do perform tests with alternative `metric_type` in files. However, those are done via temporary files being created as part of the `test_metadata` file rather than adding a bunch of different files directly.

WardBrian: Ah, I see that now. Thanks!

cmdstanpy/stanfit/metadata.py
WardBrian: Seems reasonable -- I'd just throw a comment on it explaining that this is the reasoning. We could also consider storing them as `list[float]|list[list[float]]` here and then array-ifying them when accessed.

cmdstanpy/stanfit/metadata.py
WardBrian: Ah, got it. In the few other places we've needed to do this, we have used the stringified version (there's one in InferenceMetadata, for example). I don't have any strong preference here, besides for consistency.

amas0: Since it seems as good a time as any, I've made a change to remove all the stringified instances in favor of the future default. 

It's rare, but I found that this can play nicer with certain IDE configurations.

cmdstanpy/stanfit/mcmc.py
WardBrian: I do think that would be slightly cleaner

amas0: Okay, should be updated.

cmdstanpy/stanfit/mcmc.py
```diff
@@ -382,6 +399,27 @@ def _validate_csv_files(self) -> dict[str, Any]:
...
+                    MetricInfo.model_validate_json(f.read())
+                )
+
+        metric_types = {cmi.metric_type for cmi in self._chain_metric_info}
+        if len(metric_types) != 1:
+            raise ValueError("Inconsistent metric types found across chains")
+        self._metric_type = self._chain_metric_info[0].metric_type
+        self._metric = np.asarray(
+            [cmi.inv_metric for cmi in self._chain_metric_info]
+        )
```
WardBrian, line 418: Because we end up needing an `np.asarray` for this anyway, I do wonder if the pydantic model for this should simply encode that the metric is a list of floats or list of lists of floats. We could then ditch the arbitrary_types_allowed config, at the cost of needing slightly more logic for shape validation.

test/test_metadata.py
```diff
@@ -1,9 +1,18 @@
-"""Metadata tests"""
+"Metadata tests"
```
WardBrian: I think it's better to keep all the docstrings triple-quoted

test/test_metadata.py
```diff
@@ -67,3 +76,235 @@ def test_good() -> None:
...
+        """Test that dense_e with 1D array raises ValueError"""
+        with pytest.raises(ValidationError) as exc_info:
+            MetricInfo(
+                stepsize=0.5,
+                metric_type="dense_e",
+                inv_metric=[1.0, 2.0],  # type: ignore
+            )
+        assert "Dense inv_metric must be 2D" in str(exc_info.value)
+
+    def test_dense_e_non_square_raises_error(self) -> None:
```
WardBrian, line 209: I might be missing it, but is there a test for an invalid metric_type?

WardBrian (commented):
I think this is pretty close to good, just a couple small questions and test comments

test/test_metadata.py
amas0: Just added some!

cmdstanpy/stanfit/mcmc.py
amas0: Okay, I dropped the numpy array as an attribute in favor of native types. Wasn't so bad of a change.

WardBrian (approved):
I think this is good to go! Sorry for the lag time between reviews

WardBrian:
(assuming those test failures are just noise?)

amas0:
No worries! 

There was one test failure that was numerical. That was just noise. 

I'm scratching my head on these Windows failures, though. I tried to replicate in a VM locally, and they were all passing. Have you seen anything like this before?

WardBrian:
I wouldn't be shocked if there was some weird difference, but it's hard to tell with just the test output as it is. Maybe print the metric files, and put `[]` inside the `all` call so it's a list comprehension rather than a generator expression? That will make it more obvious which index is the problem

amas0:
:shrug: 

I guess we're good

WardBrian:
huh, alright!

Issue #757, opened 2024-05-24
Feature Request: Pass Stan model as string to CmdStanModel


fonnesbeck:
#### Summary:

Add an optional argument for passing Stan model as string to `CmdStanModel`

#### Description:

It would be a huge convenience to be able to pass a Stan model to `CmdStanModel` as a string, rather than first having to write it to a file. Adding boilerplate code to write a string to a file from Python only to load it back for the model to use seems unnecessary from a user's standpoint.

#### Current Version:
1.2.2

WardBrian:
This was last discussed here: 
https://github.com/stan-dev/cmdstanpy/issues/513

The discussion at the time mainly focused on how Stan code is code, so it makes sense to live in its own files that can be separately version controlled etc.

I think the way we would implement this would also force you to re-compile each time Python got restarted, which is also downside to allowing this

vankesteren:
Just to add to this point, feel free to ignore or explicitly choose `wontfix` (I appreciate clear scope decisions in a project like this). 

I understand that you want to instil good coding practice into your user base. But having this as a feature would be really nice nonetheless for quick experimentation. I miss it when I want to explain or try out some small model. For new users, the barrier will be lowered if you allow stan code to be passed. It also has advantages for developers: a minimal reprex will be much easier to create and read, for example in github issues. Compare these:

### Option 1: current situation
```py
from cmdstanpy import CmdStanModel
from tempfile import TemporaryFile

stan_string = """
data {
  int<lower=0> N;
  array[N] int<lower=0,upper=1> y;
}
parameters {
  real<lower=0,upper=1> theta;
}
model {
  theta ~ beta(1,1);
  y ~ bernoulli(theta);
}
"""

with TemporaryFile(mode="w", suffix=".stan", delete=False) as file:
    file.writelines(stan_string)

stan_model = CmdStanModel(stan_file=file.name)
stan_data = {"N": 10, "y": [0, 1, 0, 0, 0, 0, 0, 0, 0, 1]}
stan_fit = stan_model.sample(stan_data)
```

### Option 2: allowing stan code as a string in the constructor
```py
from cmdstanpy import CmdStanModel

stan_string = """
data {
  int<lower=0> N;
  array[N] int<lower=0,upper=1> y;
}
parameters {
  real<lower=0,upper=1> theta;
}
model {
  theta ~ beta(1,1);
  y ~ bernoulli(theta);
}
"""
stan_model = CmdStanModel(stan_code=stan_string)
[...]

PR #849, opened 2026-01-28
Add from_string classmethod constructor for CmdStanModel

vankesteren:
#### Submission Checklist

- [x] Run unit tests

#### Summary
This implements a basic version of instantiating a CmdStanModel from a code string. For discussion, see #757; this implements option 3 from that discussion. I've implemented the classmethod itself as well as a (basic) unit test. I've ensured code passes both flake8 and pylint.

This is a first draft, feel free to reject this pull request, to request changes, or to completely ignore :+1: I don't want to overload you as maintainers!

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting (this will be you or your assignee, such as a university or company): Erik-Jan van Kesteren

### DIFF
--- a/cmdstanpy/model.py
+++ b/cmdstanpy/model.py
@@ -208,24 +208,24 @@ class CmdStanModel:
         """
         Initialize CmdStanModel using stan code from a string.
 
-        Internally, this method first writes the Stan code to a 
+        Internally, this method first writes the Stan code to a
         temporary file and then calls the regular CmdStanModel()
         constructor.
 
         Note that this method is only meant to be used for small
         examples and / or for development purposes. For all other
-        purposes, we recommend developing the model in a separate 
+        purposes, we recommend developing the model in a separate
         .stan file and calling CmdStanModel() directly.
 
         :param stan_code: Stan program code as a string.
-        :param **kwargs: keyword arguments passed to the CmdStanModel() 
+        :param **kwargs: keyword arguments passed to the CmdStanModel()
             constructor.
         """
-        with tempfile.TemporaryFile(mode="w", suffix=".stan", delete=False) as f:
+        with tempfile.TemporaryFile("w", suffix=".stan", delete=False) as f:
             f.writelines(stan_code)
-        
-        return cls(stan_file = f.name, **kwargs)
-        
+
+        return cls(stan_file=f.name, **kwargs)
+
     @property
     def name(self) -> str:
         """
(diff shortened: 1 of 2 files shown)


mitzimorris:
I really don't want to see this added to the code base, given that we now have the most excellent stanplayground which does this in a browser, which seems to me, where you want to be doing quick experimentation.
https://github.com/flatironinstitute/stan-playground

bob-carpenter:
@mitzimorris:  I think the motivation is that people like to construct Stan programs programmatically out of pieces and it's a pain to have to find a location on disk that's stable to read and write from.  

Personally, I don't like when people encode their models as a string in Python.  It causes at least three problems.  First, if you use `'`, you can't use vector or matrix transposition without escaping the `'` and if you use `"`, you can't use `print()` with strings.  The second problem is that line numbers in error messages no longer make sense.  Third, you don't get syntax highlighting (unless you have a fancier editor than I have!).  Having said that, lots of people seem to just write their Stan code into the Python file so there aren't multiple files to manage.

Stan Playground has its own problems with being a web app that's not under a user's control, so there's no expectation of privacy, no service level agreement for response time or uptime, etc.  Of course, you could always serve it up locally, but that seems painful and doesn't interface with the rest of Python tools without saving output and importing.

mitzimorris:
> I think the motivation is that people like to construct Stan programs programmatically out of pieces and it's a pain to have to find a location on disk that's stable to read and write from.

@bob-carpenter - I was responding to this comment on the referenced issue:

> But having this as a feature would be really nice nonetheless for quick experimentation. I miss it when I want to explain or try out some small model. For new users, the barrier will be lowered if you allow stan code to be passed. 

but the use case Bob mentions is valid - I defer to the active cmdstanpy devs on this one.

amas0:
I think the all the criticisms/concerns raised about encoding a Stan program as a string are valid; however, I do think there is a user experience challenge that people run into when using cmdstanpy. 

Jupyter notebooks are ubiquitous in the Python data ecosystem. The notebook interface encourages not have separation between the different components of your work, whether that's code, narrative, visualizations, etc. Requiring users to separate out their Stan code to a different file breaks that paradigm. It's just ends up being awkward for users.

Of course, this paradigm isn't always good, but that doesn't mean it's not useful in some instances. I would always encourage people use better development practices, but I think compromising to meet users where they are at is also important.

vankesteren:
Thanks all for your engagement with this! Good that you reached a consensus on this; if this request comes back then we'll have a really clear answer to point to 😉 

I'll close the issue now.

Issue #848, opened 2026-01-27
Implemention discussion: using Stan config JSON throughout library and other changes


amas0:
As part of ongoing work to revamp the IO of the library (#785), we would like to transition from sourcing Stan config info out of comment lines in the Stan CSV to the structured configuration json files (#714). Since #838, the config files are now generated as part of running any Stan method and saved to the output dir along with other files. This includes storing the location of the config in the RunSet object.

In line with work in #844 which introduced pydantic as a dependency for parsing and validating data sourced from json files, we'd like to do the same here. A underlying challenge here is that the config information is used across a number of different methods and not always in a consistent way. There is also a desire to maintain a flexible interface that can accommodate new or in-development outputs/methods. An ideal solution would standardize the IO configuration across methods while not constraining future development. 

It's not obvious to me the right path, so I'd like to use this thread to get some thoughts down and discuss ideas before moving forward with a particular implementation.

### Where do we actually use the configuration in the library?

To my understanding, the primary use of parsing the Stan configuration is the `from_csv()` function to reconstruct Fit object from the Stan CSV files. When constructing a Fit object directly from a method like `model.sample()`, the configuration stored is sourced from the arguments passed by the user. (I believe the only time there is an explicit comparison between the output configuration and the user args is in `CmdStanMCMC._validate_csv_files()`). 

So, under the current approach, we really only need to parse and validate configuration insofar as it is used to create Fit objects from CmdStan output files (this would need a change from a `from_csv` function to a `from_output_files` function). The good news is that the required fields to do this is relatively small.

As a side note, in theory I think there may be some value in validating user-passed arguments against the arguments parsed out of config files. Could be the kind of thing that would catch bugs that are otherwise hiding.

### What should a configuration object look like? How should validation be done?

When it comes to using the configuration, one way I see that this can be done is a single pydantic model for all the different methods, such as:

```python
  from typing import Literal                                                                                                                                                                        
  from pydantic import BaseModel, ConfigDict                                                                                                                      
                                                                                                                                                                                                    
  class ConfigInfo(BaseModel):                                                                                                                                                                      
      """Structured representation of a config JSON file"""                                                                                                                                        
      model_config = ConfigDict(extra="allow")                                                                                                                                                      
                                                                                                                                                                                                    
      model_name: str                                                                                                                                                                               
      stan_major_version: int                                                                                                                                           
      stan_minor_version: int                                                                                                                                                                       
      stan_patch_version: int                                                                                                                                                                       
[...]

Issue #846, opened 2025-12-04
CmdStanMCMC.summary: allow filtering of variables


WardBrian:
This can use the -i argument from https://github.com/stan-dev/cmdstan/pull/1143

Some care will be needed if we want to support mapping from containers to all their indices, e.g. if the user can specify `vars=['a']` for a `vector[3] a;` and that's equivalent to `-i a.1 -i a.2 -i a.3`

PR #851, opened 2026-04-12
Use cmdstan config and decouple stanfit objects from RunSet

### LINKED ISSUE #785: Path forward on IO operations

amas0:
PR intending to close #785 and implementing ideas discussed in #848.

---

Okay. I'll start by saying that the line diff is highly inflated due to needing to replace/generate large numbers of test files to be compatible with the new output files that are being consumed. The actual diff in the core code is:

```
amas ~/projects/cmdstanpy $ git diff --stat develop HEAD cmdstanpy | tail -n1
 16 files changed, 1597 insertions(+), 1786 deletions(-)
```

So still a major PR, but not as insane as the top-line numbers look. 

---

This is a major refactor of how stanfit objects are structured and constructed. The major points are:

1. Config JSON output that was added in #838 is now the primary source of run configuration, replacing CSV comment parsing. (Parsing the variational `eta` value is the only survivor). Each method now has a typed Pydantic object to validate/parse the config JSON and is stored in `fit.config`.
2. Stanfit objects and their construction now decoupled from `RunSet`s. The stanfit objects are all now dataclasses constructed directly from the output files of the cmdstan process. This relegates RunSet's role to constructing the cmdstan argument/output file name patterns.
   - `CmdStanMLE` takes a `converged` argument on construction sourced from the retcodes of the process (assumed true when from files).
3. Stanfit objects now share a generic `StanFit` base class which collects a good chunk of code that was getting repeated. This breaks down further into `SingleFileFit` (MLE/VB/Laplace/Pathfinder) and `MultiChainFit` (Sample/GQ).
4. Handful of noteworthy public API changes downstream from all of these changes
   - `from_csv` is now `from_output_files`
   - `save_csvfile` is now `save_output_files`
   - `fit.runset` is gone
5. As we now build fit objects `from_output_files`, we _require_ config JSONs next to the CSV files
6. Tests updated to include proper output files for cases where they would now be expected (config json/metric files).

Also, since we are on version 2.39 now it seemed a good time to bump the minimum version to 2.37 (this simplified some things).

There are some things that I think can still be improved, but this PR is already close to getting out of hand and I'd like to defer those to try to focus on getting this one "done". Just as note, these are:

* One consequence of now relying on a greater number of output files is the discovery of those output files has gotten clumsy. The logic in `stanfit/__init__.py` bothers me. When we pulled everything out of a CSV, we didn't really have to worry about it. 
  That being said, I think the solution here might be to promote the _config_ to be the primary file for resolving the others. It includes explicit references to other output files and a proper Stanfit could be built starting from it. Right now we have to crawl around the output dir to find the appropriate output files. The config file, since it is generated _by cmdstan_, should always have accurate references to the file locations (provided they hadn't been moved).
* The base StanFit objects that I introduced aren't explicitly ABCs, but they could be. Not sure if there's a reason one way or another? I chose not to make it an ABC for now.

Would really appreciate some thoughts on these changes @WardBrian, especially if any of these changes would cause issues for how this library is used with dev versions of cmdstan.

Disclaimer: AI was used throughout this PR -- critically for helping with regenerating the appropriate test files and applying initial prototype structures across the different methods. All code in this PR was reviewed by me before marking it as ready.


cmdstanpy/stanfit/pathfinder.py
```diff
@@ -2,30 +2,63 @@
...
     """
 
-    def __init__(self, runset: RunSet):
-        """Initialize object."""
-        if not runset.method == Method.PATHFINDER:
+    metadata: InferenceMetadata
+    model_name: str
+    csv_file: str
+    config: PathfinderConfig
+    config_file: str | None = None
```
WardBrian: Is this really optional? Doesn't look like it in from_files

cmdstanpy/stanfit/pathfinder.py
```diff
@@ -2,30 +2,63 @@
...
+                f"Did not find Pathfinder config, instead found: {conf_name}"
             )
-        self.runset = runset
-        self._draws: np.ndarray = np.array(())
-        self._metadata = InferenceMetadata.from_csv(self.runset.csv_files[0])
+
+        metadata = InferenceMetadata.from_csv(csv_file)
+        return cls(
+            metadata=metadata,
+            csv_file=str(csv_file),
```
WardBrian: The super-correct way to convert a `str|os.PathLike` to a str is [`os.fspath()`](https://docs.python.org/3/library/os.html#os.fspath)

cmdstanpy/stanfit/pathfinder.py
```diff
@@ -2,30 +2,63 @@
...
+            model_name=stan_config.model_name,
+            config=stan_config.method_config,
```
WardBrian: It's obviously nice that we don't have any downstream isinstance checks, but it is a bit of a bummer that we end up throwing away the extra information that is present in the StanConfig (e.g. model name, etc). Is there some trick to keep it all around, while maintaining that this is a Pathfinder-specific config?

cmdstanpy/stanfit/__init__.py
```diff
@@ -256,21 +255,21 @@ def from_csv(
...
-                model_exe=model,
-                chain_ids=None,
-                method_args=pathfinder_args,
+            if len(csvfiles) != 1:
+                raise ValueError(
+                    'Expecting a single Pathfinder Stan CSV file, '
+                    f'found {len(csvfiles)}'
+                )
+            csv_file = csvfiles[0]
+            config_file = os.path.splitext(csv_file)[0] + '_config.json'
```
WardBrian, line 272: I assume this is temporary but it's worth putting a comment saying so, since this is kind of a nasty assumption

WardBrian (commented):
A few comments/questions, but overall I like the look of the code!

Responding to your comments in the PR:
- Runset being left out of the result entirely makes sense. We will have to see how this feels when we tackle something with more moving parts like the MCMC class, but for now I agree.
- Repr changes make sense.
- Stdout file is good. We may want to also do the same thing with diagnostic files and profiling files, which gets us closer to just including the whole Runset again... maybe we have a small dataclass that is like a cut-down runset and stores the relevant files, can be re-used across all of them?
- I agree, it makes sense to replace `csv` with `output` across the board, both in the save functions and the `from_csv` function. This would make sense going forward anyway, c.f. https://github.com/stan-dev/design-docs/pull/58
- Yep, totally agree that InferenceMetadata might want to take a hike when all is said and done. The header could be stored in each class, or in the cut-down-Runset idea ("`FileInfo`"?)

cmdstanpy/stanfit/pathfinder.py
amas0: It's optional in the sense that if you pass a `config` in directly, you don't need to reference the file itself. Outside of simplifying some testing, I'm not sure if it will come up much? The intended entrypoint in practice is always going to be the `from_files`.

cmdstanpy/stanfit/pathfinder.py
amas0: Good point. We could just store references to both `stan_config` and `config`? The top level `StanConfig` would still contain the inference config within it, but it would just be the same object. Storing both would keep the top level info and still let us have `config` be validated that the method is correct. Keeps the type checker happy and doesn't drop any info. Thoughts?

cmdstanpy/stanfit/__init__.py
amas0: Yeah definitely temporary. I was planning on doing away with this function entirely once the full changes are in.

cmdstanpy/stanfit/pathfinder.py
WardBrian: That seems fine. I was wondering if pydantic had a more elegant way, but it's not like doing that actually costs any extra memory or anything

amas0: We could do something fancy like make `StanConfig` a generic, which pydantic has good support for:

```python
from typing import Generic, TypeVar

AnyMethodConfig = Annotated[                                                                                                                                               
    SampleConfig | OptimizeConfig | PathfinderConfig
    | LaplaceConfig | VariationalConfig | GeneratedQuantitiesConfig,                                                                                                       
    Discriminator("method"),                                                                                                                                               
]
                                                                                                                                                                             
MethodT = TypeVar("MethodT", bound=BaseModel, default=AnyMethodConfig)                                                                                                     
   
class StanConfig(BaseModel, Generic[MethodT]):                                                                                                                             
    model_config = ConfigDict(extra="allow")                                                                                                                               
   
    model_name: str                                                                                                                                                        
    stan_major_version: str
    stan_minor_version: str
    stan_patch_version: str

    method_config: MethodT
```

Then for example we could have:

```python
@dataclass
class CmdStanPathfinder:
   # other fields ...
   config: StanConfig[PathfinderConfig]
```

And parsing like `StanConfig[PathfinderConfig].model_validate(jsons)` would validate it against the Pathfinder instance of the config. This would result in only one reference to the config but the type checker would know that it has the pathfinder configuration fields within `method_config`.

WardBrian: That seems like a good representation of what's actually happening under the hood

amas0:
@WardBrian Working through some of the other methods. 

For `CmdStanMLE`, it looks like our typical approach to checking for convergence is looking at the return codes via the runset. Is there a more natural way to do this for reading from the output files directly?

I see we could maybe parse that info out of the `stdout` file, but I don't love that.

WardBrian:
Hm, that is indeed tricky. I think we should add the piece of information that currently ends up in stdout (`Optimization terminated normally:`/`with error:`) to the output csv, but that won't help with existing, released versions.


Looking at the code, it looks like the only place we actually check the return codes is in model.py, before we call the CmdStanMLE constructor, so maybe it is fine?

amas0:
Yeah. In my local implementation, I take `is_converged` as an argument to the constructor. We have it in `model.py` so it should be fine in practice, but it means we can't get full reconstitution from just output files for MLE (unless we parse stdout).

It does seem like the kind of thing that should be available in the CSV? Like a `converged__` column or something? Not a major issue for now, just doesn't make things as clean as I'd like.

WardBrian:
>  but it means we can't get full reconstitution from just output files for MLE (unless we parse stdout).

This is true for the existing `from_csv` function, isn't it?

amas0:
Yeah good point, this is the snippet from `from_csv`:

```python
            for i in range(len(runset._retcodes)):
                runset._set_retcode(i, 0)
            return CmdStanMLE(runset)
```

Looks like in recreating the runset, we just manually set all return codes to 0, which is just assuming convergence I believe.

WardBrian:
Yeah, so we won't be doing any worse, at least!

amas0:
@WardBrian Is there an equivalent to saving the metric json for the variational method? Looking at CmdStanVB now and we parse the variational eta value from the stancsv comments, but afaik we don't have an alternative way to source it for now.

bob-carpenter:
There's no preconditioning in our two variational inference implementations, ADVI and Pathfinder, so there's no equivalent of a metric/mass matrix/preconditioner.

WardBrian:
@bob-carpenter I believe @amas0 was asking just if we wrote the metadata from the algorithm as a json, similar to the metric in sampling, not if there was actually a preconditioner

Unfortunately to my knowledge the answer to this is no. It's a shame to have to keep all the comment-reading machinery around for something as rarely-needed as the final eta value from ADVI...

bob-carpenter:
Sorry for the confusion.

Keeping things around for `eta` is also silly in that estimating `eta` is one of the major weaknesses of our ADVI implementation (another being restriction to diagonal or dense approximations, a third being the high variance of low-sample ELBO evaluations). 

A better approach for `eta` might be to treat it as another column in the tabular output like step size.  Like step size, it will change as the algorithm progresses then hopefully stabilize.

amas0:
Yeah, in particular, I was referring to this section that appears in the output CSV if you run the variational method:

```csv
lp__,log_p__,log_g__,theta
# Stepsize adaptation complete.
# eta = 0.1
0,0,0,0.5838444
```
I just meant that we output the algorithm stepsize info to the CSV in the same way that we write the sampler adaptation output (as comment lines after the column headers). But in the sample case, we can output equivalent info to a metric JSON file, but that doesn't seem to be the case for the variational method here. 

So, I guess the question is what do we want to do with this in the near term? I think it makes sense to have it be an additional column as Bob suggested? I suppose the question is whether it is important enough to maintain the comment parsing in the near term or should we drop keeping track of it for now until an alternative source is available?

WardBrian:
I'm not sure it makes sense as a column, since the advi output is only the one row of the varational parameters, in which eta has already been adapted, and then forward sampling iterations 

Github code search doesn't seem to find any public code which actually calls `.eta`, but I'm not sure this is really a good measure. 

Maybe we preserve a dumber version of comment parsing just for that, i.e. a function that just does 

```python
for line in file:
  if line.startswith("# eta ="):
    return float(line.split('=')[1])
```

Rather than all the present machinery? We can also try to add a structured/json output like for the metric, but that would take longer to thread through the cmdstan code and we'd still need something in the meantime

amas0:
Honestly, the code surface for just being able to parse eta out is relatively small under the current structure. It's just one call that partitions the CSV into comment lines and draw lines and then a second call to pull out the eta value from the comment lines. Even keeping those, we'll be able to drop a good chunk Stan CSV parsing logic that goes through the comments.

amas0:
As I continue to chip away at this, I'd like to get some thoughts on how we should manage the way we current do testing in light of these changes. 

We have a large test directory that contains a number of pre-generated cmdstan output files. These are used in fixtures throughout testing to reconstruct stanfit objects. With this new approach of not parsing everything out of the CSVs, we'll need to have config files in some way.

As part of this, should we got through all of those instances and regenerate the output files to make it so our testing fixtures correspond properly to the new structure? That would probably be the minimal change for testing purposes. Alternatively, are there other approaches we may want to consider for testing that wouldn't require this? (Like mocking file outputs or generating output files on the fly as part of testing)

WardBrian:
I think re-generating them is reasonable. My opinion is that storing the files is basically the same as a really-good mock, since it actually interfaces with real outputs. We like having a few end-to-end tests that actually run cmdstan (especially to spot changes in the output formats over time), but it's too slow to do that for every test IMO

amas0:
Okay, I think I'm happy enough with where this stands to have it ready for review. At the very least to have someone else other than me look at this.

WardBrian:
@amas0 it is on my list for this week!

cmdstanpy/stanfit/__init__.py
```diff
@@ -35,31 +28,116 @@
...
+    latent dynamics or profiling data. Those are written as
+    ``<base>_diagnostic[_<id>].csv`` and ``<base>_profile[_<id>].csv``,
+    so they would otherwise be mistaken for additional chains.
+    """
+    return not _CHAIN_ID_RE.sub('', csv_file.stem).endswith(
+        ('_diagnostic', '_profile')
+    )
+
+
+def _sidecar(csv_file: Path, kind: str) -> Path:
```
WardBrian: [nit] don't love the name, maybe "_accompanying_json" or something?

cmdstanpy/stanfit/__init__.py
```diff
@@ -35,31 +28,116 @@
...
-    respectively.
+    checks that the files were produced by that method.
+
+    The files are assumed to follow CmdStanPy's own output naming: each
+    chain's CSV ends in ``_<chain_id>``, and the config (and, for sample,
+    metric) JSONs CmdStan writes sit next to the CSVs they belong to, with
+    a run that shared one process across chains having a single config
+    named for its first chain. Outputs written by running CmdStan directly
+    with a single output name and ``num_chains`` follow a different naming
+    scheme and are not currently supported.
```
WardBrian: Is this comment correct? Because that will definitely be a blocker, we don't want to drop support for the built-in parallelism

cmdstanpy/stanfit/__init__.py
```diff
@@ -35,31 +28,116 @@
...
-    | CmdStanPathfinder
-    | CmdStanLaplace
-    | None
-):
+) -> CmdStanMCMC | CmdStanMLE | CmdStanVB | CmdStanPathfinder | CmdStanLaplace:
     """
-    Instantiate a CmdStan object from a the Stan CSV files from a CmdStan run.
-    CSV files are specified from either a list of Stan CSV files or a single
+    Instantiate a CmdStan object from the output files of a CmdStan run.
+    Output files are specified as either a list of Stan CSV files or a single
```
WardBrian: I feel like we should allow the _config.json to be in the list, and use that if so rather than going looking for it?

cmdstanpy/stanfit/__init__.py
```diff
@@ -71,213 +149,73 @@ def from_csv(
...
-            # bugfix 425, check for fixed_params output
-            try:
-                check_sampler_csv(
-                    csvfiles[0],
-                    iter_sampling=num_samples,
-                    iter_warmup=num_warmup,
-                    thin=thin,
-                    save_warmup=save_warmup,
+        if method_name == 'sample':
+            # Collect the sidecar JSONs that exist next to each CSV. A
```
WardBrian: I think the word sidecar is the new em-dash. I almost never see it outside llm comments/thought traces (unless you count by bartenders)

cmdstanpy/stanfit/__init__.py
```diff
@@ -71,213 +149,73 @@ def from_csv(
...
-                num_draws=config_dict['num_draws'],  # type: ignore
-                num_paths=config_dict['num_paths'],  # type: ignore
-            )
-            cmdstan_args = CmdStanArgs(
-                model_name=model,
-                model_exe=model,
-                chain_ids=None,
-                method_args=pathfinder_args,
+        builder = _SINGLE_FILE_METHODS.get(method_name)
+        if builder is None:
```
WardBrian: I don't know if we really have enough of these to justify this over just an if/else chain or a match statement

cmdstanpy/__init__.py
```diff
@@ -32,7 +32,7 @@ def _cleanup_tmpdir() -> None:
     CmdStanMLE,
     CmdStanPathfinder,
     CmdStanVB,
-    from_csv,
+    from_output_files,
```
WardBrian, line 36: It's unfortunate we don't have a release in the 1.x cycle that formally deprecated from_csv, we might want to keep it around as an alias as a result. Thoughts?

cmdstanpy/model.py
WardBrian, line 1: General comment: I really like how the changes look in this file, to me it really motivates the rest of the changes to be able to have this all be nicer!

WardBrian (commented):
Starting small, really only reviewed a few pieces for now and decided to stop with my questions

cmdstanpy/__init__.py
amas0: My approach with this PR was to specifically target a 2.0 release, which is why I opted for a clean compatibility break here. That being said an alias that replicates the previous experience of using from_csv probably wouldn't be an issue?

cmdstanpy/stanfit/__init__.py
amas0: Yeah...I had a similar reaction. I satisfied myself when I saw it was apparently [a thing](https://en.wikipedia.org/wiki/Sidecar_file), but I can clean that bit up.

cmdstanpy/stanfit/__init__.py
amas0: To clarify, this refers specifically to if you run a model with built-in parallelism manually in cmdstan _without cmdstanpy at all_ and then try to load it in to cmdstanpy. When you run it via cmdstanpy, we set the output filenames in the cmdstan call to conform to an expected naming convention.

As long as you are trying to load in a cmdstan output that was run via cmdstanpy, you will be fine.

amas0:
@WardBrian you've rightfully picked at the code around file discovery so far. I wanted to raise this point I made in PR description (in case you missed it) to get your thoughts:

> One consequence of now relying on a greater number of output files is the discovery of those output files has gotten clumsy. The logic in `stanfit/__init__.py` bothers me. When we pulled everything out of a CSV, we didn't really have to worry about it.
That being said, I think the solution here might be to promote the config to be the primary file for resolving the others. It includes explicit references to other output files and a proper Stanfit could be built starting from it. Right now we have to crawl around the output dir to find the appropriate output files. The config file, since it is generated by cmdstan, should always have accurate references to the file locations (provided they hadn't been moved).

I had a working version of a change where we essentially start from the config.json, which has enough information in it to  reconstruct the output file names without needing to crawl around the directory (I believe). The current structure we have goes through a few heuristics to identify which CSVs are draws (vs diagnostics/profile) and to resolve which chain id each draw csv corresponds to. Starting from a config(s) would remove that step -- it would also enable easier support for the case where you ran cmdstan outside of cmdstanpy and are trying to load fit in (related to your comment above). It would also handle the case where there are multiple fits in the same directory (which is a current problem area if you provide a directory).

The big drawback is the potential brittleness of the path references in the config. If a fit is generated and the output files are moved, then those references are no longer correct. One mitigation there would be to have the stanfit's `save_output_files(...)` method also update the path references in the config when moved. So managing fit "bundles" from within cmdstanpy would be safe, but not necessarily outside of that.

We do now require a config to be present to construct a stanfit, so I don't think it's too crazy to consider this route, but it would certainly be a shift from past behavior.

WardBrian:
If I recall correctly, in the multi threaded case the output field in the config.json is still not a literal valid path, but instead the “template” without the chain id. Is that not the case?

amas0:
Ah yeah, I see - that's how it is for cmdstan directly. I had confused myself with how we run it through cmdstanpy.

In these changes I take advantage of the 2.37 output file csv to have full paths for each chain csv in the config.

WardBrian:
Ah, because we now pass the names we’d like as a comma separated list? That’s nice

amas0:
Yeah, it makes it so that if we run a model via cmdstanpy then we get a pretty clean output set. I think it was also necessary to maintain the naming convention discussed in #835 now that we have the config in the mix?

The underlying tension here I think is between whether the output files were generated by cmdstan directly or indirectly via cmdstanpy. In the latter case, we get a good amount of control to have a clean output to work with. In the former case, it's trickier.

WardBrian:
How about:

- if the config files come from cmdstanpy or follow our desired format, you can pass a subset (maybe just the config) or a directory and we figure out the rest
- if they didn’t, you need to list every single one of them in the call to the function

Reasonable?

amas0:
I'm good with that.

test/data/runset-good/bern-1_config.json
```diff
@@ -0,0 +1,66 @@
...
+  "id" : 1,
+  "data" : {
+    "file" : "test\/data\/bernoulli.data.json"
+  },
+  "init" : "2",
+  "random" : {
+    "seed" : 12345
+  },
+  "output" : {
+    "file" : "\/tmp\/tmpuisjazeu\/bernoullik799pfr2\/bernoulli-20260720125836_1.csv",
```
WardBrian: Note that it looks like our own test files invalidate the assumption that the path in the config is authoritative (and i don't see how it could be otherwise, unless we made these fictitiously relative?)

amas0: Well these would represent the instance where the output files were generated and then moved/renamed, right? At the time that this config file was actually generated, it was authoritative.

Then when it was moved it to the test dir and renamed to be consistent with the old files, it became wrong.

Whether or not these paths are authoritative is only relevant for the output file discovery process, i.e. `from_output_files`. When we are constructing a particular stanfit such as `CmdStanMCMC.from_files(...)`, we always pass all the files explicitly anyways.

test/test_laplace.py
```diff
@@ -55,16 +55,30 @@ def test_laplace_from_csv() -> None:
         assert isinstance(fit3.mode, cmdstanpy.CmdStanMLE)
 
 
+def test_laplace_pickle_ability() -> None:
```
WardBrian: Should this be with the others in test_compliance?

amas0: Yep, good catch

cmdstanpy/stanfit/naming.py
WardBrian, line 1: I think this file could belong in /utils, or just put these functions in utils/stancsv?

cmdstanpy/stanfit/mle.py
```diff
@@ -1,60 +1,63 @@
...
-            all_draws = stancsv.csv_bytes_list_to_numpy(draws_lines)
 
-        except Exception as exc:
-            raise ValueError(
-                f"An error occurred when parsing Stan csv {csv_file}"
-            ) from exc
-        self._mle: np.ndarray = all_draws[-1]
-        if self._save_iterations:
-            self._all_iters: np.ndarray = all_draws
+    def _warn_if_no_iterations(self) -> bool:
```
WardBrian: [nit] because this returns the bool, I think it should be called `check` or `require` or similar, not `warn`

cmdstanpy/stanfit/mle.py
```diff
@@ -1,60 +1,63 @@
...
+    @classmethod
+    def from_files(
+        cls,
+        csv_file: str | os.PathLike,
+        config_file: str | os.PathLike,
+        stdout_file: str | os.PathLike | None = None,
+        converged: bool = True,
+    ) -> CmdStanMLE:
+        return cls(
+            converged=converged,
```
WardBrian, line 44: Do you plan to read this out of the csv file on newer CmdStans? We can't require it yet, but we could use it if we have it?

cmdstanpy/stanfit/pathfinder.py
```diff
@@ -2,227 +2,65 @@
...
+            **cls._from_files_kwargs(
+                csv_file, config_file, stdout_file, PathfinderRunConfig
             )
-        self.runset = runset
-        self._draws: np.ndarray = np.array(())
-        self._metadata = InferenceMetadata.from_csv(self.runset.csv_files[0])
-
-    def create_inits(
-        self, seed: int | None = None, chains: int = 4
-    ) -> list[dict[str, np.ndarray]] | dict[str, np.ndarray]:
```
WardBrian, line 32: Was this an intentional deletion? We still want it, and it's still in CmdStanVB

WardBrian: (Does this mean it was untested? maybe!)

WardBrian (commented):
Ok, I'm up to 96/103 'viewed' on this, which means I mostly have only the big files like GQ, MCMC, and `base` left. A few questions in the mean time

cmdstanpy/stanfit/mle.py
amas0: Yes we should do this -- was that a 2.39 update?

WardBrian: [Yep](https://github.com/stan-dev/stan/pull/3379)!

cmdstanpy/stanfit/pathfinder.py
amas0: This moved into `base.SingleFileFit` -- pathfinder and laplace share the same implementation inherited from the base and MLE/VB overwrite it with their own implementation. The VB and base implementations are very close though -- VB pulls from `variational_sample` whereas the others pull from `draws`, so those can probably be combined and Pathfinder/Laplace/VB could share an implementation.

WardBrian: Ah, I see. Yeah, it seems like a good idea to introduce a `_draws_for_inits()` method that the superclass can call?

cmdstanpy/stanfit/naming.py
amas0: I'm in the process of trying to clean up the whole file discovery logic bit, which includes this file. I agree this doesn't need to be its own file. It got that way because Laplace was doing its own file discovery when doing `from_files` because it looked for the mode csv file as a sibling file.

I am chipping away at an update that will hopefully simplify this. Stay tuned.

cmdstanpy/stanfit/pathfinder.py
amas0: Cool, I like that -- easy fix.

cmdstanpy/stanfit/naming.py
amas0: This file was removed

amas0:
Okay I implemented a refactor of the `from_output_files` flow that I think clarifies the approach (although it is still admittedly complex).

To start, I finally the loading from file logic into its own `stanfit/from_outputs.py` file and cleaned out `__init__.py`. Within this new file we have all the logic of how we go from a set of output files -> StanFit object. The two pathways are (from your suggestion):

* Fit file discovery based on expected naming convention/config output names. This can be done by passing a directory, a config json path, or a stan csv path. Each of these try to get to a config and then uses the file names in the config to collect the appropriate other files (which are assumed to be next to the config).
* Explicit file passing. This expects a list of file paths to be passed and will attempt to classify the files into known kinds and build an appropriate fit from there. This does no file discovery, so if you, for example take a multi-chain sample run and only pass in the output files for a single chain, you will get a single chain fit back.

This diagram shows visually the valid paths of how one can build a fit via this function:

```mermaid
flowchart TD
    A["from_output_files"]
    B["Single path provided"]
    C["All files explicitly passed"]
    A --> B
    A --> C
    D["Directory"]
    E["Config JSON"]
    F["Stan CSV"]
    G["Find sibling config JSON"]
    B --> D
    B --> E
    B --> F
    F --> G
    H["Build FitFiles manifest"]
    E --> H
    G --> H
    I["Find all valid Configs"]
    D --> I
    J["One unique valid config (multiple configs from the same sample fit produce the same config)"]
    K["Two unique valid configs (Laplace edge case with mode fit)"]
    I --> J
    I --> K
    J --> H
    K --> H
    L["Classify each input file by kind"]
    C --> L
    M["Verify minimally valid set of files included (no discovery)"]
    L --> M
    M --> H
    N["Construct StanFit from FitFiles"]
    H --> N
```

If one wishes to be more explicit, of course, one can always call a particular StanFit's `from_files` method directly.

cmdstanpy/stanfit/pathfinder.py
amas0: This has been added - VB, Laplace, and Pathfinder now use the base class create_inits. MLE and Sample have their own implementations.

cmdstanpy/__init__.py
amas0: I re-added `from_csv`, which now wraps `from_output_files` (with some extra logic to maintain the historical compatibility) and added a deprecation warning. It just says: `from_csv is deprecated and will be removed in a future 
release; use from_output_files instead`

Not sure if we want to specify a particular deprecation timeline? Or if we want to deprecate it at all?

cmdstanpy/stanfit/mle.py
amas0: Added support for this. Since we're still supporting < 2.39, I made it so when `model.optimize()` has `required_converged=True`, then the process check is still the primary convergence check. On 2.39+ we'll verify they agree.

amas0:
Okay I think I addressed the comments you had.

There's a section of the docs that needs to be rewritten if we decide to go through with these changes, so I'll get to that in a bit.

cmdstanpy/stanfit/mcmc.py
```diff
@@ -328,18 +321,21 @@ def draws(
             return flatten_chains(self._draws[start_idx:, :, :])
         return self._draws[start_idx:, :, :]
 
+    @classmethod
     def _comparable_config(
-        self, config: StanConfig[SampleConfig]
+        cls, config: StanConfig[SampleConfig]
     ) -> dict[str, Any]:
         """Extends the base cross-chain checks with the sampler settings
         which affect how the draws are laid out."""
```
WardBrian: This comment seems out of date

cmdstanpy/stanfit/gq.py
WardBrian, line 1: It feels like this could be simplified even more using the new base class. For example, should `_draws_start` be in the base classes instead? (I think it's more or less equivalent to the new `_draws_for_inits`?)

cmdstanpy/stanfit/from_outputs.py
```diff
@@ -0,0 +1,817 @@
...
+            f'{len(configs)} configs, and {len(metric_files)} metrics.'
+        )
+    return FitFiles(
+        method=method,
+        csv_files=(csv_files[0],),
+        config_files=(configs[0].path,),
+    )
+
+
+def _build_fit(files: FitFiles) -> AnyStanFit:
```
WardBrian, line 687: style-wise, I'd inline this function

test/data/output_bad_metric_4.csv
WardBrian, line 1: Not specific to any of these individual files, but I think it's time to mark all the csv and json files in the `test/` folder as `linguist-generated` in .gitattributes: https://github.com/github-linguist/linguist/blob/main/docs/overrides.md

cmdstanpy/stanfit/from_outputs.py
WardBrian, line 1: This file is a lot of code but it all looks pretty reasonable. I do wonder if it should be split up or if the parts that are e.g. super-Laplace specific be moved to the laplace.py file

WardBrian (commented):
Sorry for my delay! I'm quite happy with how things are working

cmdstanpy/stanfit/from_outputs.py
amas0: I just went ahead and moved all the logic into from_output_files

test/data/output_bad_metric_4.csv
amas0: Neat - didn't know this was a thing.

cmdstanpy/stanfit/from_outputs.py
amas0: I did another pass on this file -- some of the laplace stuff made sense to move out to laplace.py. The file probably could be broken up further, but I wasn't sure if splitting this logic across multiple files was better than just a ~700 line file? If you have a particular suggestion, I can see how that might play out.

cmdstanpy/stanfit/gq.py
amas0: I gave a shot at looking at what _draws_start in the base class would look like, but I didn't feel it was much of an improvement.

That being said, I did make a handful of changes that I think are cleanups of some of the longstanding code in here (also fixed up some bugs that AI flagged).

Let me know if you think there are any other pathways you wanna go here.

amas0:
As a side note, I made an update to the docsrc conf.py to remove a check on a nearly 10 year old sphinx version? I think we can probably assume that we'll have newer versions than that?

WardBrian:
That seems like a safe bet…

I haven’t forgotten about this PR, just had a bunch of other stuff to do with the 2.40 release. Hopefully next week I will have enough time to get this review done and the PR merged

amas0:
No worries! I figure if it took me the better part of a year to get this together, what's a few more weeks. Appreciate all the comments.

WardBrian (approved):
I think this is good to go! Thanks a ton @amas0 

It's worth going through and seeing what else we'd like to do before 2.0, if anything, and we can start planning that release (~mid October?)

amas0:
Agreed!  #846  and #863 seem like obvious choices?

WardBrian:
Yes! That being said, both of those are features that would also be fine in a 2.1, the main thing we need to get locked down is any other changes to the API or removals

PR #856, opened 2026-05-25
Update timing parsing regex to account for scientific notation.

chinandrew:
#### Submission Checklist

- [x] Run unit tests

#### Summary
The regex in `parse_timing_lines()` has a bug where if scientific notation is present, only the exponent is captured: 
```
In [1]: from cmdstanpy.utils.stancsv import parse_timing_lines

In [2]: lines = [
   ...:     b'# \n',
   ...:     b'#  Elapsed Time: 630099 seconds (Warm-up)\n',
   ...:     b'#                435856 seconds (Sampling)\n',
   ...:     b'#                1.06595e+06 seconds (Total)\n',
   ...:     b'# \n'
   ...: ]

In [3]: parse_timing_lines(lines)
Out[3]: {'Warm-up': 630099.0, 'Sampling': 435856.0, 'Total': 6.0}
```

This change updates the regex for the first capture group to include `e`, `+`, and `-`, so the output becomes
```
In [3]: parse_timing_lines(lines)
Out[3]: {'Warm-up': 630099.0, 'Sampling': 435856.0, 'Total': 1065950.0}
```
Not sure if support for `-` is really needed in practice, but I added it for completeness. In keeping with the existing escaped `.` in the regex, I also escape `+` and `-`, though I believe no escapes are actually needed in the character class, i.e.  `([\d.e+-]+)` is valid. 

Note that this can lead to the case where  the `Warm-Up + Sampling != Total` due to truncation from the scientific notation. In this example, `1.06595e+06` gets parsed to `1065950.0` but Warm-Up + Sampling is `630099.0 + 435856.0 = 1065955.0`. Arguably this is not a _parsing_ issue since it's correctly reading the comment, but perhaps some additional logic to set Total = Warm-Up + Sampling might be useful.

#### Copyright and Licensing

Myself

### DIFF
--- a/cmdstanpy/utils/stancsv.py
+++ b/cmdstanpy/utils/stancsv.py
@@ -324,7 +324,7 @@ def parse_timing_lines(
 
         if not in_timing_block:
             continue
-        match = re.findall(r"([\d\.]+) seconds \((.+)\)", str(line))
+        match = re.findall(r"([\d\.e\+\-]+) seconds \((.+)\)", str(line))
         if match:
             seconds = float(match[0][0])
             phase = match[0][1]
--- a/test/test_stancsv.py
+++ b/test/test_stancsv.py
@@ -459,6 +459,22 @@ def test_parsing_timing_lines() -> None:
     assert out['Total'] == 0.001581
 
 
+def test_parsing_timing_lines_scientific_notation() -> None:
+    lines = [
+        b'# \n',
+        b'#  Elapsed Time: 630099e0 seconds (Warm-up)\n',
+        b'#                4358560.0e-1 seconds (Sampling)\n',
+        b'#                1.06595e+06 seconds (Total)\n',
+        b'# \n',
+    ]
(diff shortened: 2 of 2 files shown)


WardBrian:
Thanks for the contribution @chinandrew. Is this a hypothetical issue or did you actually get the new test output from a run of cmdstan?

chinandrew:
The example above is from an actual cmdstanpy run:
```
b'#  Elapsed Time: 630099 seconds (Warm-up)\n',
b'#                435856 seconds (Sampling)\n',
b'#                1.06595e+06 seconds (Total)\n',
```
For the unit test, I modified `435856` to  `4358560.0e-1` to capture the `-` case, though I just ran an extremely short run and small values appear to get rounded to 0:
```
#  Elapsed Time: 0.015 seconds (Warm-up)
#                0 seconds (Sampling)
#                0.015 seconds (Total)
```
so perhaps the `-` case isn't even needed.

WardBrian (approved):
Thanks!

PR #864, opened 2026-09-18
Update Windows RTools support and add Windows ARM64

ndabas:
#### Submission Checklist

- [x] Run unit tests

#### Summary

I have updated the RTools installation support on Windows, so versions up to 4.5 are now supported. Critically, that now adds native support for Windows on ARM64. I have updated the rest of the build-related code to work on both AMD64 and ARM64 versions of Windows.

The one thing worth pointing out is that I ran into 4 tests failing only on Windows ARM; this is because those particular tests were running an exe and immediately trying to delete it. On Windows 11 ARM, [Windows Defender cannot be disabled](https://github.com/actions/runner-images/issues/14326), per [the readme](https://github.com/actions/runner-images/blob/main/images/windows/Windows11-Arm64-Readme.md); and Windows Defender tries to scan every executable when it is run, so deleting it immediately fails.

We can choose to not add Windows ARM to the main workflow, as that workflow has a subset of supported platforms anyway. In that case I can remove the changes to main.yml and the delete-retry workaround, let me know.

I also added lots of tests and a new workflow to test the toolchain installation on Windows.

#### Copyright and Licensing

Nikhil Dabas


.github/workflows/windows-toolchain.yml
```diff
@@ -0,0 +1,131 @@
+name: Windows toolchain
+
+# Exercises the RTools install + discovery path on real Windows runners.
+# Kept out of the main matrix because the RTools 4.4/4.5 installers are ~450MB.
+
+on:
+  workflow_dispatch:
+  schedule:
+    # weekly: catches CRAN retiring the pinned installer revisions
```
WardBrian, line 9: My recollection is this is part of why we've stuck with rtools40 for so long, because there is not a stable link for the newer ones. Is that what this comment means?

cmdstanpy/utils/filesystem.py
```diff
@@ -20,6 +21,27 @@
 EXTENSION = '.exe' if platform.system() == 'Windows' else ''
 
 
+def delete_file(path: str, timeout: float = 5.0) -> None:
```
WardBrian, line 24: I think this function could be in the test utilities, rather than the actual package

cmdstanpy/compilation.py
```diff
@@ -195,8 +199,8 @@ def validate_user_header(self) -> None:
                 )
             if "allow-undefined" not in self._stanc_options:
                 self._stanc_options["allow-undefined"] = True
-            # set full path
-            self._user_header = os.path.abspath(self._user_header)
+            # clang treats backslashes in the -include path as escapes
+            self._user_header = Path(self._user_header).absolute().as_posix()
```
WardBrian, line 203: It might make sense to improve the path handling for this in cmdstan itself:
https://github.com/stan-dev/cmdstan/blob/d3d5df6a22565edbe13edbd4eb40762cc8c5a4d6/make/program#L47

For other variables we use `subst` to swap backslashes:
https://github.com/stan-dev/cmdstan/blob/d3d5df6a22565edbe13edbd4eb40762cc8c5a4d6/make/program#L83

cmdstanpy/install_cmdstan.py
```diff
@@ -71,9 +73,11 @@ def is_windows() -> bool:
     return platform.system() == 'Windows'
 
 
-MAKE = os.getenv('MAKE', 'make' if not is_windows() else 'mingw32-make')
 EXTENSION = '.exe' if is_windows() else ''
 
+# Windows ARM64 needs the Stan Math support added in stan-dev/math#3051
+MIN_WINDOWS_ARM64_VERSION = (2, 35)
```
WardBrian, line 79: This is happening in parallel to #851, which will be defining a minimum cmdstan version greater than this for support anyway, so I'm not sure we need to check this specifically here

cmdstanpy/install_cxx_toolchain.py
```diff
@@ -25,17 +25,38 @@
...
+    },
+}
 
 
 def usage() -> None:
     """Print usage."""
     print(
         """Arguments:
-        -v (--version) :CmdStan version
+        -v (--version) : RTools version: 3.5, 4.0, 4.4 or 4.5
```
WardBrian, line 59: Since we will be releasing version 2.0 soon, we can definitely remove the old 3.5 support, if not also 4.0

cmdstanpy/install_cxx_toolchain.py
```diff
@@ -25,17 +25,38 @@
...
+    validate_dir,
+    wrap_url_progress_hook,
+)
 
 EXTENSION = '.exe' if platform.system() == 'Windows' else ''
-IS_64BITS = sys.maxsize > 2**32
+
+# CRAN embeds build revisions in the RTools 4.2+ installer filenames, so we
+# use the r-hub mirror, which publishes them under a stable 'latest' tag.
+# These are the builds the CmdStan guide points users at.
```
WardBrian, line 42: This seems to go against the comment in the ci workflow about the links changing

docsrc/installation.rst
```diff
@@ -218,6 +221,17 @@ machine when running ``install_cmdstan``. If the wrong choice is made, or if you
...
 to one of the above options, or to "false" to use the standard x86 download.
 
+Windows on ARM64
+................
+
+CmdStan can be built on ARM 64-bit (``aarch64``) Windows machines using the
+ARM64 build of RTools 4.4 or 4.5, which CmdStanPy installs and detects
+automatically. CmdStan 2.35 or later is required.
+
+A Python interpreter built for Intel/AMD and running under emulation reports an
```
WardBrian, line 231: I was wondering about emulation, since we don't provide windows/aarch64 builds for the `stanc3` compiler -- but I guess it works out?

cmdstanpy/utils/__init__.py
```diff
@@ -8,12 +8,19 @@
 
 from .cmdstan import (
     EXTENSION,
+    RTOOLS_LAYOUTS,
+    RToolsLayout,
     cmdstan_path,
```
WardBrian, line 13: It doesn't look like all these new imports are actually used outside of the other utilities, so probably don't need to appear here

WardBrian (commented):
Thanks @ndabas! This looks very nice overall.

My main concerns are about the overlap this has with the large refactor currently happening in #851. I think that together they will actually clean up a lot of the code, but it will likely require some changes here once that one is merged.

.github/workflows/windows-toolchain.yml
ndabas: Yes that is what this comment was about. However I noticed that in the [CmdStan docs](https://mc-stan.org/docs/cmdstan-guide/installation.html#windows), they point to stable download URLs from the R-Hub project (like https://github.com/r-hub/rtools45/releases/download/latest/rtools45.exe) so I incorporated those in the code. Let me know if that sounds fine -- and I will remove this comment and the weekly run.

cmdstanpy/utils/filesystem.py
ndabas: Agreed, I'll move it over.

cmdstanpy/compilation.py
ndabas: Thanks for the pointers, I'll see how to clean this up a bit.

cmdstanpy/install_cmdstan.py
ndabas: Okay that sounds great -- I'll remove this check then.

.github/workflows/windows-toolchain.yml
WardBrian: That does sound better to me

cmdstanpy/install_cxx_toolchain.py
ndabas: It's probably harmless to leave the old ones around, but yes, we can eliminate some test burden by at least getting rid of 3.5. I'll do that.

cmdstanpy/install_cxx_toolchain.py
ndabas: Yes indeed, when I was checking the CmdStan docs I discovered that they have these stable links. I'll remove the CI comment as mentioned.

docsrc/installation.rst
ndabas: Yes since it's just a compiler of sorts it works fine on ARM64 under emulation. I can look at that as well after this PR to get native support.

cmdstanpy/utils/__init__.py
ndabas: I'll verify that and remove these if needed.

.github/workflows/windows-toolchain.yml
ndabas: Done, I have removed that comment and the weekly cron.

cmdstanpy/utils/filesystem.py
ndabas: Done, moved to the tests.

cmdstanpy/compilation.py
ndabas: I had a look, and yes, it would be great to fix it there. I'll submit a PR for CmdStan too, but in the meanwhile, since we will support at least a few older versions of CmdStan, I've kept the code here and updated the comment to clarify.

cmdstanpy/install_cmdstan.py
ndabas: I have removed this check.

cmdstanpy/install_cxx_toolchain.py
ndabas: I have removed all 3.5 support now.

cmdstanpy/utils/__init__.py
ndabas: I have removed those un-needed imports.

test/test_compilation.py
```diff
@@ -201,6 +201,18 @@ def test_user_header() -> None:
...
+def test_user_header_is_posix_path() -> None:
+    """Backslashes in the -include path are eaten by clang."""
+    opts = CompilerOptions(
+        user_header=os.path.join(DATAFILES_PATH, 'return_one.hpp')
+    )
+    opts.validate()
+    assert '\\' not in opts.user_header
+    assert opts.user_header.endswith('/return_one.hpp')
+    assert os.path.isfile(opts.user_header)
+    assert str(opts.cpp_options['USER_HEADER']) == opts.user_header
```
WardBrian: This test focuses us to do the update, but I think it would be better to have a test that a windows path actually compiles, which would work whether us or CmdStan do the fix up

test/test_cxx_installation.py
```diff
@@ -40,17 +448,80 @@ def test_install_not_windows() -> None:
...
+
+def test_parse_cmdline_args(monkeypatch: pytest.MonkeyPatch) -> None:
+    monkeypatch.setattr(
+        sys, 'argv', ['install_cxx_toolchain', '-v', '4.5', '-d', 'somewhere']
+    )
+    args = install_cxx_toolchain.parse_cmdline_args()
+    assert args['version'] == '4.5'
+    assert args['dir'] == 'somewhere'
+    assert args['silent'] is False
+    assert args['progress'] is False
```
WardBrian: This can go away; it's basically just testing argparse

WardBrian (commented):
The changes look pretty good to me at this point, a couple comments on the tests

.github/workflows/windows-toolchain.yml
```diff
@@ -0,0 +1,129 @@
...
+          from cmdstanpy.utils import cxx_toolchain_path
+
+          cxx_toolchain_path("${{ matrix.rtools }}", os.path.join(os.environ["RUNNER_TEMP"], "toolchain"))
+          with open("hello.cpp", "w") as f:
+              f.write(textwrap.dedent("""
+                  #include <cstdint>
+                  #include <iostream>
+                  int main() { std::cout << "hello\\n"; return 0; }
+              """))
+          subprocess.run(["g++", "-O0", "hello.cpp", "-o", "hello.exe"], check=True)
```
WardBrian: I think windows runners might already have `g++` in their path, so I'm not confident this is really testing all that much. Maybe it could also assert on `gcc --version` or `which gcc`?

test/test_compilation.py
ndabas: Agreed, I have extracted a test in `tests\test_model.py` (`test_model_compile_user_header`) for this check.

test/test_cxx_installation.py
ndabas: Agreed, done.

.github/workflows/windows-toolchain.yml
ndabas: Agreed, I already had an assert on the `gcc` path in the previous step of the job, and I have added a `gcc --version` to this step (in addition to checking the architecture against what is expected.)

ndabas:
Thanks for the (re-)review, I have made these changes and everything is passing on my fork.

cmdstanpy/compilation.py
ndabas: https://github.com/stan-dev/cmdstan/pull/1360

WardBrian (approved):
Thank you @ndabas!

ndabas:
Thanks for merging, @WardBrian! Do you have a timeline for the next release? I'm going to look at adding Windows ARM64 support into https://github.com/facebook/prophet next.

WardBrian:
We are hoping to start the 2.0 release process sometime [in October](https://github.com/stan-dev/cmdstanpy/pull/851#pullrequestreview-5282002225). We will probably do a beta period for a bit first, but my hope would be a general release by the end of the month or so

ndabas:
Great, thanks!

